PR #22, opened 2025-12-05
/oss-forensics command

mbrg:
This repo introduces changes to the main orchestrator flow for `/oss-forensics` command. Instead of trying to spawn an orchestrator agent that spawns sub-agents (which isn't supported by Claude Code), the main Claude Code agents becomes that orchestrator.

This command is now ready to be used. However, the big warning is that if you're not careful you wouldn't love your GCP bills after the GH Archive sub-agent had its fun with it. Please use with caution for now, until I introduce a budget system.


danielcuthbert (approved):
What an amazing PR, having this oss skill is epic. LGTM

PR #57, opened 2026-01-05
Feature/frida integration

Splinters-io:


Splinters-io:
apologies, I'll share what I have and have another run at this and how it integrates with the vision that is coming into frame we might have to trash this one, I'll tap you up if I get stuck

PR #82, opened 2026-03-28
add /understand: because grep and pray died in 2003

danielcuthbert:
map it, trace it, hunt the variants. teach yourself why the thing you
  thought was safe absolutely is not. feeds into validation so you can
  stop guessing and start knowing.

  adversarial code comprehension — map your attack surface before you
  attack it. trace flows. hunt variants. teach yourself the framework
  before you trust it. outputs feed /validate so the whole thing hangs
  together.

  yes it has gates. no you cannot skip them.

### DIFF
--- a/.claude/commands/understand.md
+++ b/.claude/commands/understand.md
@@ -10,6 +10,8 @@ It is a work in progress, remember that.
 /understand <target> [--map] [--trace <entry>] [--hunt <pattern>] [--teach <subject>] [--out <dir>]
 ```
 
+If no mode flag is given, default to `--map`.
+
 ## Modes
 
 | Flag | What it does |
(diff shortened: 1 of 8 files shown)


.claude/commands/understand.md
grokjc: **No default mode** 

/understand ./src with no flags produces nothing. Default to --map (the natural starting point or print usage?

**Mode combination ordering**

--map --trace EP-001 works because map produces the EP IDs that trace consumes. Arbitrary combinations have no defined order. If that was: map → trace → hunt → teach, that would match the "natural attack progression" already stated in SKILL.md and ensure the right sequence.

grokjc (commented):
This looks like a great addition to the codebase and is good to go. 😄 

Moving forwards, the main challenge will be getting the schema overlap right between this new code and the validation pipeline so they can flow and integrate seamlessly.

I think we can put our heads together to solve that one, but that shouldn't stop /understand working as a valuable standalone feature!

**Validation Pipeline specifics**

The PR documents specific integration points with the validation pipeline in both CLAUDE.md and SKILL.md - things like context-map.json pre-populating attack-surface.json for Stage B, and variants.json expanding checklist.json for Stage 0. None of this is wired up yet - the validation stages don't look for these files, and the schemas don't match.                                                              
                                                                                                                                                                                                                   
More importantly, the relationship between /understand and the checklist needs some thought. Stage 0 builds checklist.json as a completeness inventory, and /understand's --map and --hunt modes need similar completeness guarantees (GATE-U4 explicitly requires searching the full codebase). It's not clear yet if /understand should generate its own checklist, consume an existing one, or share the logic with Stage 0.

I'd suggest replacing the specific integration claims with a general note that pipeline integration with /validate is planned, and we can get that plumbing done later. We'll need minor changes to the validation pipeline to ensure that the /understand output is consumed when available.

grokjc (approved):
Looks good to me!

PR #85, opened 2026-03-28
Allow validation to consume understand output

grokjc:
Initial plumbing. We still have work to do, but this is the first step.


grokjc:
Trivial change, self-approving 😄

Issue #125, opened 2026-04-04
Add /diagram command — Mermaid visual maps from /understand and /validate outputs


danielcuthbert:
## What's this?

Right now, once you've run `/understand` or `/validate`, you're left staring at a pile of JSON files. `context-map.json`, `flow-trace-*.json`, `attack-tree.json`, all the data is there, but you have to read it like a machine to make sense of it. This adds a `/diagram` command that turns those outputs into Mermaid flowcharts you can actually look at.

I still need to manually copy and paste into a mermaid viewer but still, you can't automate all teh things right?

## What it renders

| File | Diagram |
|------|---------|
| `context-map.json` / `attack-surface.json` | Flowchart LR — entry points → trust boundaries → sinks, with unchecked flows as dashed edges |
| `flow-trace-*.json` | Flowchart TD — each hop in the call chain, tainted variable at each step, branches, attacker control summary |
| `attack-tree.json` | Flowchart TD — knowledge graph with nodes coloured by status (confirmed/disproven/exploring/unexplored) |
| `attack-paths.json` | Flowchart TD per path — step chain with proximity score (0–10) and blocker annotations |

Point it at any `/understand` or `/validate` output directory and it picks up whatever JSON is there. No configuration needed.

## How you use it? 

```
/diagram .out/code-understanding-20240101/ --target myapp
```

Writes `diagrams.md` into the same directory. Renders in GitHub, VS Code, Obsidian, or anything Mermaid-aware.

You can also use the library directly:

```python
from packages.diagram import render_and_write
from pathlib import Path

render_and_write(Path(".out/code-understanding-20240101/"), target="myapp")
```

## Tested on PyGoat

Ran this against the deliberately vulnerable PyGoat Django app. The context map correctly flagged four unauthenticated endpoints (including an unauth RCE via Python source file overwrite), coloured public entry points differently from auth-gated ones, and drew the dashed unchecked-flow edges to the right sinks. The flow trace for the RCE path rendered all six hops cleanly.

Looks pretty epic too

<img width="647" height="997" alt="Image" src="https://github.com/user-attachments/assets/12178d45-7269-4a9a-b77e-a5b499a85a8a" />
<img width="1020" height="1238" alt="Image" src="https://github.com/user-attachments/assets/6c2b8006-fb26-403f-974f-40f8b1ceee7f" />

## Current state

This is a WIP. It works, but there's room to grow:

[...]

PR #126, opened 2026-04-04
added new diagram ting, addressing issue #125

danielcuthbert:
The joys of jetlag.... so this should fit well into our /understand and /validate and mapping stages. Keen to see your thunks. 

Also check #125


grokjc (commented):
This is awesome 😄 

There's only one real bug here which is the list vs dict thing
(and adding hypothesis support is probably a good move)

PR #153, opened 2026-04-08
Add coverage tracking plugin and coverage records

grokjc:
Coverage Phase 2 prerequisite. Tracks what tools examined for
  accurate coverage reporting.
  
  - plugins/coverage/: Claude Code plugin with async PostToolUse
    hook on Read. Bash+jq, fails fast when no run active. Filters
    by source extension and target directory.
  - core/coverage/: Python package for coverage records. Builds
    coverage-record.json from hook manifest (LLM tools) or Semgrep
    JSON output (paths.scanned).
  - core.run complete converts manifest to record automatically.
  - Launcher auto-loads the plugin via --plugin-dir.
  
  11 files, 445 tests passing. E2E verified.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -88,6 +88,10 @@ The `start` command automatically resolves the output directory using the active
 
 Commands run via `python3 raptor.py` (scan, agentic, codeql, fuzz, web) manage lifecycle internally — do not call the stubs separately for those.
 
+### Coverage tracking
+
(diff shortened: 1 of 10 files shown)


grokjc:
The advantage of this being a Claude plugin is that we can load it from our launcher but anyone running 'claude' doesn't have to enable it unless they want it (so no surprises for people just forking the project) and we don't have to override settings.json directly.

danielcuthbert (approved):
love this! got me thinking how we bring in codeQL too but that's one ill make an issue for

PR #156, opened 2026-04-09
feat: SAGE persistent memory integration

l33tdawg:
## Summary

- **Hybrid SDK + MCP integration** with SAGE consensus-validated persistent memory, enabling cross-session learning across all RAPTOR analysis workflows
- **SDK layer** (`core/sage/`): Async client, config, and drop-in `SageFuzzingMemory` replacement that adds semantic recall while preserving the existing API
- **MCP layer**: All 16 Claude Code agents updated with SAGE memory instructions and domain-specific knowledge persistence
- **87 institutional knowledge entries** seeded from codebase (exploitation primitives, expert personas, methodology, LLM prompts, signal heuristics)
- **16 agent role definitions** registered on SAGE network with capabilities and domain assignments
- **Docker Compose sidecar** with SAGE + Ollama (nomic-embed-text for embeddings)

## What's included

| Component | Files | Description |
|-----------|-------|-------------|
| SDK adapter | `core/sage/` (4 files) | Async client with lazy init, graceful degradation, env var config |
| SAGE-backed memory | `core/sage/memory.py` | Drop-in FuzzingMemory replacement with SAGE persistence + JSON fallback |
| MCP config | `.mcp.json` | SSE connection to SAGE for Claude Code agents |
| Agent updates | 16 `.claude/agents/*.md` | SAGE memory instructions per agent role |
| CLAUDE.md | `CLAUDE.md` | SAGE domains, boot sequence, usage guidance |
| Docker sidecar | `docker-compose.sage.yml` | SAGE + Ollama with health checks |
| Knowledge seed | `scripts/seed_sage_knowledge.py` | Extracts 87 entries from codebase |
| Agent registration | `scripts/register_agents.py` | Registers 16 agents with role definitions |
| Unit tests | 3 test files | 38 new tests (config, client, memory) |
| Documentation | `SAGE_INTEGRATION.md` | Full setup guide with architecture diagram |

## SAGE domains

11 semantic domains for organized knowledge persistence:
`raptor-findings`, `raptor-fuzzing`, `raptor-crashes`, `raptor-forensics`, `raptor-exploits`, `raptor-methodology`, `raptor-campaigns`, `raptor-reports`, `raptor-agents`, `raptor-primitives`, `raptor-prompts`

## Test plan

- [x] 38 new unit tests passing (config, client, memory)
- [x] 7 existing tests still passing
- [x] Seed script dry-run verified (87/87 entries)
- [x] Agent registration dry-run verified (16/16 agents)
- [x] E2E: Docker SAGE sidecar starts cleanly
- [x] E2E: Knowledge seeded (87/87 OK)
- [x] E2E: Agents registered (16/16 OK)
- [x] E2E: Semantic recall verified across all domains (primitives, fuzzing, agents, personas, methodology)
- [x] E2E: SageFuzzingMemory connects and recalls from SAGE
- [ ] Test with actual RAPTOR scan workflow

## Quick start

```bash
# Start SAGE sidecar
docker compose -f docker-compose.sage.yml up -d

# Install SDK (optional, for Python integration)
pip install sage-agent-sdk httpx

# Seed institutional knowledge
python3 scripts/seed_sage_knowledge.py

# Register agents
python3 scripts/register_agents.py

# Restart Claude Code session to pick up .mcp.json
```

SAGE is purely additive — all RAPTOR commands work normally when SAGE is unavailable.


l33tdawg:
## E2E Proof: SAGE Improves Analysis Quality

Tested against OWASP NodeGoat with the agentic pipeline:

| Metric | Scan 1 (no history) | Scan 2 (with SAGE) |
|--------|--------------------|--------------------|
| Pre-scan recall | 0 memories | **8 memories recalled** |
| LLM references SAGE context | N/A | **Yes** |
| Exploitable confirmed | 3 | **4** |
| Exploits generated | 3 | **4** |

From scan 2, the LLM wrote in its reasoning for SARIF-0000:

> *"The historical context from SAGE indicates previous findings of vulnerabilities related to insecure configurations and plaintext HTTP links, implying a lack of overall security awareness."*

Cross-project recall also verified: scanning Raptor's own codebase recalled patterns from NodeGoat scans.

**Bonus fix included**: Scanner `total_files_scanned` was always 0 — now shows actual count.

55 unit tests + 7 existing tests all passing.

danielcuthbert:
oof this is a biggy! thank you @grokjc and I need to go through this

grokjc:
Thanks for identifying the total_files_scanned == 0 bug. 

I have been working on a new coverage subsystem which is about halfway done now. We've just merged #167 which adds per-tool coverage records (coverage-semgrep.json, coverage-codeql.json) that capture scanned files from SARIF at scan time. These feed project coverage which is now the canonical way to see what was examined. The total_files_scanned metric in scan_metrics.json is only consumed by a log line and I think the coverage records make it pretty much redundant.                                             
                                                                                                                                                                                                                 
Rather than adding a second extraction pass over the same SARIF, I'd prefer to either drop total_files_scanned from the metrics or have it read from the coverage record. Happy to help with that if you'd like to update the PR?

The rest of the PR (which needs a proper review when I have time!) is frankly awesome and doesn't conflict with the SARIF parser/coverage path changes as far as I can see, and is a very welcome addition. Thank you!

l33tdawg:
Morning guys - I've rebased the PR on your current main and updated the total_files_scanned fix to align with your coverage records. Instead of the SARIF extraction approach (which was essentially a second pass), the metric now derives from load_records() after the coverage records are written in scanner.py. The SARIF parser is reverted to upstream. 

I made a new commit is e78c998 if you want to review just that change. All existing tests pass (SARIF parser: 17, coverage: 39). Happy to iterate on anything — let me know when you get a chance to look at the rest of the PR :)

grokjc:
I had a look and I'm mostly thinking about how to 'focus' this into fewer files/directories. 

**Plumbing**

There's a lot of files here. It might be tidier to focus on /core/sage/ - the scripts and tests could go under there (and s/test/tests/ to match the test of our codebase) - so that everything is in one place.  Secondly, I wonder if we could just plumb this in by adding some instructions to CLAUDE.md rather than adding boilerplate to each of the .md files and repeating the same instructions in n places which creates maintenance burden as new features come online or instructions change.

**SAGE Itself**
I suppose the point of SAGE in this context is to share findings across projects, but is there any kind of namespace/project support that would prevent that if the user didn't want to pollute project A with project B's findings?

Does SAGE support authentication? Even localhost-only presents an attack surface to some degree (I wonder if that is really an issue.)

Lastly, is there any concept of 'staleness' ? Do SAGE memories time out? If we're seeding knowledge about a codebase like RAPTOR which has high churn, some of those memories won't be true even in the near future. Just curious how we handle eviction/invalidation in that scenario.

Some of these aren't really RAPTOR questions but I'm curious 😄

l33tdawg:
Thanks for the thorough review, and agreed on both points. 

I'll consolidate scripts and tests under core/sage/ (and tests/ not test/) so the integration lives in one place. And you're right about the per-agent boilerplate, so 'll collapse the SAGE instructions into CLAUDE.md once and strip the repetition from the individual agent files.

Namespace / Project Isolation 

Yes — SAGE has three layers here:

1. Domain tags: Every memory belongs to a domain (e.g. raptor-findings, raptor-fuzzing). Queries filter by domain at the SQL level, so project A's memories don't appear in project B's queries unless you explicitly ask for them.

2. Agent RBAC: Each agent has a domain_access policy — an explicit allowlist of which domains they can read/write. Domain not in the list = denied. 

3. Domain registry with ownership: Domains can be registered with an owner agent. Access grants control who gets read/write at what classification level (public → secret). Unregistered domains default to open for backward compat.

So for the Raptor use case, the raptor-* domain prefix naturally isolates findings. If a user doesn't want cross-project recall, they just scope their queries to their project's domains. 
                                                                          
Authentication

SAGE uses Ed25519 signature-based authentication, not static API keys. Every request requires three headers: X-Agent-ID (public key), X-Signature, and X-Timestamp. The signed message binds to HTTP method + path + body, so signatures can't be replayed cross-endpoint. There's also a replay cache (10k entries, 5-minute window) and optional nonce support. The signature is verified both at the REST middleware layer and again on-chain during transaction processing — the ABCI layer doesn't trust the REST layer's assertion. /health and /ready are the only unauthenticated endpoints. For the Raptor integration specifically, the SDK handles signing transparently — the user just configures their key and the client does the rest.

Staleness / Decay

Yes, this is built in. Memories decay exponentially over time: confidence(t) = initial_confidence * exp(-λ * days_elapsed) * corroboration_boost. 

Default half-life is ~139 days (λ = 0.005/day), but domains can override this — e.g. vuln_intel decays at 0.01/day (~69-day half-life) since vulnerability intel goes stale fast. Corroboration (multiple agents confirming the same finding) boosts confidence logarithmically, so validated knowledge persists longer.

On top of that, there's a background cleanup loop (every 24h by default) that deprecates:

- Memories below a configurable confidence threshold (default 0.10)
- Observations older than 7 days
- Session-context observations older than 2 days

Deprecated memories are never deleted (audit trail), just excluded from query results. For a high-churn codebase like Raptor, the decay rates can be tuned per-domain, and the cleanup TTLs are configurable through the preferences API.

Anyway, I'll push that file reorg update, and happy to answer anything else :)

grokjc:
Thanks for that. Almost there! CI/CD tests pass 😄 

Last few things of note - including only one real bug that I can see.

1. Port mismatch in config.py                                                                                                                                                                                    
  SageConfig.url defaults to http://localhost:8080 but docker-compose.sage.yml maps port 8090 and .mcp.json uses 8090. SDK callers who don't set SAGE_URL silently connect to ... nothing? Should be 8090 I think?
                                                                                                                                                                                                                   
  2. memory.py SAGE sync never fires
  save(), remember(), and record_campaign() all call asyncio.get_running_loop() which raises RuntimeError in sync code (the entire RAPTOR pipeline is sync). The except block swallows it and the SAGE sync silently fails. The _run_async() helper in hooks.py handles this correctly so the same pattern should be used here.                                                                                           
   
  3. Command file boilerplate                                                                                                                                                                                      
  The refactor commit cleaned the SAGE sections out of .claude/agents/*.md but .claude/commands/*.md still carry identical sections in all 12 files. Since CLAUDE.md is the single source of truth, these should be removed too, with the possible exception of /exploit that is phrased slightly differently?
   
  Everything else I'll handle post-merge (docs reorganisation, some minor cleanup) as I want this to fit with some other changes I have in the pipeline.  Won't take me long to get that code ready and then we can merge + follow up!

I'm also looking at a wider cleanup of dependencies and pinning specific versions, so would be good to understand your preferences there and what is actually required. I'll probably pin a specific SAGE version as part of that work and then periodically ewview.

Appreciate the efforts in getting this far, will be a solid addition.

l33tdawg:
Morning! All three addressed in cfd4af9:

1. **Port mismatch** — Fixed `SageConfig.url` default from `:8080` to `:8090`. Test updated to match.

2. **memory.py sync never firing** — Good catch, this was a real bug. Replaced the `asyncio.get_running_loop()` / `create_task()` pattern in `save()`, `remember()`, and `record_campaign()` with `_run_async()` from `hooks.py` (which creates a fresh event loop and actually executes the coroutine). All 48 SAGE tests + 19 core tests pass.

3. **Command file boilerplate** — Stripped the `## SAGE MEMORY` sections from all 12 command files. CLAUDE.md is the single source of truth now.

Re: dependency pinning — makes total sense. The Python SDK follows semver so pinning to a specific minor (e.g. `sage-sdk>=6.5,<7`) would be the safest bet. Happy to help with that when you're ready.

Let me know if anything else needs tweaking!

l33tdawg:
Morning @grokjc - Is there anything else that's not working as you expect?

grokjc:
Sorry, been AFK.

Well, as you asked 😄  ... I've noticed a few very minor issues that I was going to handle post-merge, but you're helping us out, so here we go:

**Missing import asyncio in core/sage/memory.py**

Line 131 in _sync_to_sage does await asyncio.sleep(0.3) but asyncio isn't imported at the top of the file. The inner try/except Exception swallows the NameError, so propose() still runs but we'll get a load of "SAGE sync failed: name 'asyncio' is not defined" debug messages.

**Event loop leak in core/sage/hooks.py:52**

  ```
  finally:
      loop.close()
      asyncio.set_event_loop(asyncio.new_event_loop())   # <-- this line
```
The loop created here is set as the thread's current loop but never .close()d. Over a long-running scan this leaks one FD per hook call. The docstring already explains the fresh-loop rationale, so this line can just be deleted — the loop = asyncio.new_event_loop() at the top of the next call gives you exactly the same guarantee without the leak.

  **Pin versions for reproducibility**

As you offered to help with this: docker-compose.sage.yml uses ollama/ollama:latest and ghcr.io/l33tdawg/sage:latest. A breaking upstream release would silently break SAGE for anyone who runs docker compose pull. Please pin both to specific tags/digests you've tested against.

Same concern for the # pip install sage-agent-sdk comment in requirements.txt — can you update it to # pip install sage-agent-sdk==<version> with the version you developed against? (Leaving httpx>=0.27.0 as-is is fine — matches the rest of the file.)

**Dead function in core/sage/hooks.py:109**

recall_exploit_primitives(vuln_types) is defined but not called anywhere in the tree. Can you plumb it in or nuke it please?!

Everything else is my problem and I'll get it done. I think I'll ping you a branch that you can check out and do a full-end-to-end test with your existing SAGE setup once I've got everything sorted which will rule out me screwing anything up 😄

l33tdawg:
No worries bro, thanks for the quick pass. All four addressed in `daee3aa`:

1. **Missing `import asyncio` in `core/sage/memory.py`** — added at top. 
2. **Event loop leak in `core/sage/hooks.py`** — deleted the stray `asyncio.set_event_loop(asyncio.new_event_loop())`.
3. **Pinned versions** — `docker-compose.sage.yml` now uses `ollama/ollama:0.21.0` and `ghcr.io/l33tdawg/sage:6.5.5`; `requirements.txt` comment is now `# pip install sage-agent-sdk==6.5.5`. All three tested against these versions.
4. **Dead `recall_exploit_primitives()`** - Was a leftover from an earlier design

All 47 SAGE tests still pass. Happy to run that end-to-end test once you push the branch.

grokjc:
Plan is to rebase this PR onto the new staging branch ... stand by!

grokjc:
No way to undo that mis-fire that I can see. Can you raise a new PR targeting the sage integration branch please... hopefully that isn't too painful. My follow-up changes are ready to go.

l33tdawg:
Morning @grokjc — no worries. Opened https://github.com/gadievron/raptor/pull/190 against `sage_integration`. Same content as `daee3aa`, cherry-picked cleanly onto the new base (no conflicts with #187). 47 SAGE tests green. Ready when you are.

PR #173, opened 2026-04-14
fix: validate address in GDBDebugger.examine_memory()

danielcuthbert:
Yeah, this isn't an active issue. No RAPTOR code passes unvalidated input here, and CrashAnalyser guards its own addresses upstream. But it's a public export and GDB scripts are newline-delimited, meaning a \n in the address runs whatever follows as a GDB command. GDB has a `shell` builtin. You can see where that goes. We needs to be adults and make sure badstuff© cant happen, innit? 

We want to be correct and this is the right way to do it.

- add _HEX_ADDRESS_RE (^0x[0-9a-fA-F]+$) at module level
- examine_memory() validates address, raises ValueError on mismatch
- expand module docstring with threat model
- add TestExamineMemoryAddressValidation covering valid inputs, newline injection, semicolons, and bare integers
- move test_script_contains_run_not_redirect back to TestDebuggerNoPathInjection (got orphaned during the edit)

### DIFF
--- a/packages/binary_analysis/debugger.py
+++ b/packages/binary_analysis/debugger.py
@@ -7,12 +7,26 @@ Provides programmatic interface to GDB for crash analysis.
 Security: Input files are passed via subprocess stdin, NOT via GDB's
 `run < path` in-script redirection. This prevents CWE-78 command injection
 through crafted filenames (GDB's parser interprets shell metacharacters).
+
+Address validation: examine_memory() checks the address against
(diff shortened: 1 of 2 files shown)


grokjc (commented):
Seems solid. num_bytes isn't validated (could have a newline) and I wonder if we should cap overly long hex strings? But all good, this squashes a (theoretical) vulnerability.

PR #174, opened 2026-04-14
feat: add Rust security suites for CodeQL queries

StefanFl:
While analysing a Rust project, raptor added the security suites for Rust to the CodeQL queries. This might be useful for other users to have it right from the start.


grokjc:
Thanks @StefanFl

PR #177, opened 2026-04-16
Introduce shared memory backend and wire it across RAPTOR pipelines

### LINKED ISSUE #7: Feat. Enhanced Memory
We've introduced the concept of memory for RAPTOR and this is currently only really working for the fuzzing function (saved in .raptor in users home dir and fuzzing_memory.json, which looks like so:

```
❯ more fuzzing_memory.json
{
  "knowledge": {
    "crash_pattern:05_vulnerable_function": {
      "knowledge_type": "crash_pattern",
      "key": "05_vulnerable_function",
      "value": {
        "signal": "05",
        "function": "vulnerable_function",
        "exploitable_count": 3,
        "total_count": 3
      },
      "confidence": 0.7999999999999999,
      "success_count": 3,
      "failure_count": 0,
      "last_updated": 1762853499.712527,
      "binary_hash": "03de0d8b7a71fec3",
      "campaign_id": null
    },
    "crash_pattern:05_vuln_stack_overflow": {
      "knowledge_type": "crash_pattern",
      "key": "05_vuln_stack_overflow",
      "value": {
        "signal": "05",
        "function": "vuln_stack_overflow",
        "exploitable_count": 1,
        "total_count": 1
      },
      "confidence": 0.6,
      "success_count": 1,
      "failure_count": 0,
      "last_updated": 1762850033.512656,
      "binary_hash": "d2f20dfa9acf882d",
      "campaign_id": null
    },
   ```

What this means is that our fuzzer learns the following:

- Which strategies work for certain binaries
- Which crash patterns are exploitable
- Which exploit techniques succeed
- Full campaign history

But it is somewhat isolated and only for fuzzing, which is a bit loco. All the other functions start from scratch every time and this shouldn't be the case. If we take /scan and codeql, in an ideal world, we'd have it learning:

- **Query effectiveness** - Which CodeQL rules actually find real vulns vs noise
- **False positive rates** - "py/sql-injection in Django projects is 40% FP"
- **Build commands that work** - "Maven projects: `mvn compile -DskipTests` works 90% of the time"
- **Exploitability by rule** - "cwe-089 in Java → 85% exploitable historically"
- **Language-specific patterns** - What works, what doesn't. What should it do next? 

with /crash-analysis, the same could be said:

- **Crash → root cause patterns** - "SIGSEGV in strcpy → 95% stack buffer overflow"
- **Debugging technique effectiveness** - "rr reverse-exec brilliant for UAF, less useful for stack overflows"
- **Analysis steps that worked** - What led to successful root cause identification
- **Time-to-resolution patterns** - How long different crash types take to analyse

and so on. So moving forward, what we need is something like:

├── memory.db                    # SQLite for scalability, especially if we will be running loads 
├── fuzzing_memory.json          
├── codeql_memory.json           
├── crash_analysis_memory.json   
├── web_memory.json              
└── unified_knowledge.json

This will need a lot more work and thinking but it's on my roadmap

### PR DESCRIPTION
**Pull request by northdpole**
- Introduces a canonical SQLite-backed Memory backend in packages/autonomous/memory_store.py for cross-tool knowledge and event persistence.
- Refactors fuzzing memory adapter code to use the shared backend with clearer migration/persistence semantics and improved method-level documentation.
- Adds memory export helpers and normalized export naming (memory_knowledge.json) for consistent snapshot outputs.
- Wires shared memory recording into orchestration flows (agentic, codeql, fuzzing) plus runtime scanners (web, crash_analyser) so outcomes are persisted consistently.
- Adds regression coverage in packages/autonomous/tests/test_memory.py for upsert/query behavior, secret redaction behavior, and fuzzing adapter round-trip behavior.
- Updates package exports/imports to use the normalized Memory naming throughout.

closes #7 

Note: In order to keep Raptor backwards compatible, this has code that migrates existing json memory files. When backwards compatibility is not necessary please merge pr #178

### CODE REVIEW

**Comment by danielcuthbert** [comment]
I love you for doing this, seriously thank you! @grokjc and I are on this

**Comment by northdpole** [comment]
> I love you for doing this, seriously thank you! @grokjc and I are on this

I WILL be framing this comment!
Just happy to help :-) 
LMK if there's more wishlist items

**Comment by grokjc** [comment]
Thanks for this. There are many good ideas here, just some overlap to navigate since a few things have landed while this was open.                                                                                    
                                                                                                                                                                                                                   
Storage is pretty much settled. The SAGE integration landed on main (#190 and follow-ups through #222 - a lot of work!) and covers cross-domain knowledge persistence with consensus validation, backed by its own SQLite store server-side. Adding a second client-side SQLite would mean three copies of the same data (SAGE's SQLite, this SQLite, and the JSON fallback). SAGE also inherits from FuzzingMemory, so the constructor change here (memory_file to db_path) would break SageFuzzingMemory.
                                                                                                                                                                                                                   
Secret scanning is in flight. Hinotoi-agent's #213 (merged) hardened _sanitize_log_message() with comprehensive patterns (AWS keys, GitHub tokens, PEM blocks, Bearer/Basic auth, key=value fields). Their #223 (in review) is building shared redaction infrastructure in core/security/redaction.py. The _contains_secret() patterns here are a subset of what's already in place or in flight, so any persistence-layer scanning should plug into core/security/redaction.py rather than maintaining a separate list.                                                                                                                    
                  
**But the domain knowledge shapes are extremely valuable**. SAGE's hooks currently record findings and methodology at a fairly coarse level. This PR has more granular, domain-specific ideas about what to record that SAGE doesn't cover yet:
                                                                                                                                                                                                                   
  - Crash pattern signatures with confidence scores (from the crash_analyser.py integration)                                                                                                                       
  - Web scan payload effectiveness per target (from scanner.py)
  - Build reliability per repo/language (from raptor_codeql.py)                                                                                                                                                    
  - Strategy outcome tracking tied to binary hashes (from raptor_fuzzing.py)
  
These would make cross-domain learning genuinely more useful. For example, "AFL with ASAN found 5 crashes last time we fuzzed this binary" or "CodeQL auto-detect builds reliably succeed for this repo."        
   
These fit naturally into SAGE. Rather than a new storage layer, these knowledge entries could be proposed through core/sage/hooks.py alongside the existing findings and methodology hooks. SAGE actually gives you more than SQLite would here because the embedding layer means you get semantic recall, not just exact-match lookups. A fuzzing campaign could ask "what strategies worked for similar binaries?" and get relevant results even if the binary hash doesn't match exactly. Concretely:                                                                                                                                      
                  
  - Add new propose() calls in hooks.py after crash analysis, web scans, CodeQL builds, and fuzzing runs to record the outcomes                                                                                    
  - Add matching recall_context_for_*() queries at the start of each workflow so the planner can use prior results: best strategy, known-good build commands, which payload classes were effective
  - Use SAGE's per-repo domain tags (#198) to scope the knowledge appropriately. Strategy outcomes are per-target, but build reliability might generalise across repos with similar language profiles              
                                                                                                                                                                                                                   
Happy to help if you want to take that direction. Let me know your thoughts.

**Comment by l33tdawg** [comment]
Hey guys +1 to what grokjc said - those four knowledge shapes (crash sigs, payload effectiveness, build reliability, strategy outcomes) are exactly the kind of thing SAGE's semantic recall + per-repo domain tags were built for. Happy to share propose/recall snippets if you guys need any help...

**Comment by northdpole** [comment]
Hey guys,  sadly noticed the sage work after I opened the PR.
Yeah I agree with you, sage seems super interesting and thanks for the TIL!
A second client-side SQLite would duplicate SAGE and JSON fallbacks, `SageFuzzingMemory` shouldn’t get a breaking constructor change without an explicit migration path, and anything touching secrets should go through `core/security/redaction.py` rather than a parallel detector list.

How about this: I’ll drop the standalone SQLite backend and re-scope to `core/sage/hooks.py` only.
Will route through  propose() after crash analysis / web scans / CodeQL builds / fuzzing runs, plus use `recall_context_for_*` at workflow entry so planners actually consume prior outcomes. 

I’ll scope entries with the per-repo `domain_tag` behavior you described.

**What I need from you**: can you please accept or reshape these schemas? Happy to match your existing propose payload conventions verbatim once you point me at the canonical shapes:

1) Crash analyser — crash pattern / signature

propose payload :

```
{
  "kind": "crash_analysis.pattern",
  "summary": "Short human-readable pattern (no raw secrets)",
  "signals": {
    "fault_type": "SIGSEGV|heap-uaf|stack-overflow|…", <-- what else here?
    "top_frames_normalized": ["sym_or_offset_0", "sym_or_offset_1"],
    "binary_fingerprint": "sha256|null",
    "confidence": 0.0
  },
  "outcome": {
    "exploitability_hint": "none|unknown|likely",
    "notes": "sanitized free text"
  }
}
```

Recall query (for planner context at start of crash / triage):

“Summarize prior crash patterns with similar fault type and overlapping normalized top frames for this binary fingerprint (or nearest neighbours), including confidence and exploitability hints.”

2) Web scanner

```
{
  "kind": "web_scan.payload_effectiveness",
  "target_fingerprint": "normalized_host_path_or_route_class",
  "payload_class": "sqli|xss|ssrf|…", <-- is this necessary?
  "evidence_class": "response_delta|status_delta|timing|reflection|none",
  "effectiveness": 0.0,
  "counts": { "attempts": 0, "signals": 0 },
  "notes": "sanitized free text"
}
```

Recall:

“Which payload classes produced differentiated signals on targets matching this fingerprint or semantically similar routes?”

3) CodeQL — build reliability

```
{
  "kind": "codeql.build_reliability",
  "language": "cpp|java|javascript|…",
  "auto_detect_outcome": "success|fail",
  "successful_command": ["argv", "…"],
  "failure_modes": ["missing_toolchain", "autobuild_timeout"],
  "confidence": 0.0
}
```

Recall:

“What CodeQL build approach succeeded last time for this repo/language fingerprint, and what failures should we skip retrying?”

4) Fuzzing — strategy outcome (hash-scoped + semantic)
```
{
  "kind": "fuzzing.strategy_outcome",
  "binary_fingerprint": "sha256",
  "strategy_id": "aflplusplus_asan|libfuzzer|…",
  "campaign_summary": {
    "duration_s": 0,
    "execs": 0,
    "unique_crashes": 0,
    "hangs": 0
  },
  "confidence": 0.0,
  "notes": "sanitized free text"
}
```

Recall:

“What fuzzing strategies produced crashes for this binary or similar binaries (semantic match), ordered by confidence / recency?”

If these `kind` strings, nesting, or required vs optional fields don’t match what you already persist for methodology/findings, paste your preferred `propose()` examples and I’ll align the implementation to those exactly, otherwise I’ll proceed with the shapes above in a PR

grokjc:
Hi! Before you start coding: SAGE doesn't actually take typed schemas:

```SageClient.propose() (core/sage/client.py:119)``` is ```propose(content: str, memory_type, domain_tag, confidence)``` so content is a natural-language string and recall is semantic-similarity over the embedding of that string. There is no 'kind' field on the wire. The pattern is already there in ```core/sage/memory.py:_knowledge_to_natural_language``` — that's the shape to copy.
                                                                                                                                                                                                                   
Concrete mapping for the four:
  
  - crash_analysis.pattern → domain raptor-crashes-{repo_key}. If you want binary-scoped recall, add a ```_binary_key(binary_path)``` helper alongside _repo_key in core/sage/hooks.py.                                  
  - web_scan.payload_effectiveness → new domain raptor-web-{repo_key}. Please add it to the domain list in ```core/sage/CLAUDE.md``` in the same PR.
  - codeql.build_reliability → these generalise across repos, so don't need a per-repo-scope. Either reuse raptor-methodology or add a new raptor-codeql-build (no suffix). Your call; methodology is probably fine.                                                                                                                                                                                                             
  - fuzzing.strategy_outcome → this already exists as SageFuzzingMemory in ```core/sage/memory.py```. Please extend it instead of standing up a parallel path, and don't change its constructor (you flagged this yourself, just confirming). The existing \_knowledge_to_natural_language already covers what your "strategy outcome" schema captures.                                                                             
                  
Suggest you reuse what's already there: ```_repo_key```, ```_findings_domain```, ```_throttle```, ```_get_client```, the logger.debug on exception idiom (never raise as SAGE is purely additive). Mirror ```recall_context_for_scan(repo_path, languages)``` for the four ```recall_context_for_*``` entry points: build a query string, return List[Dict] with content/confidence/domain keys, empty list when SAGE is unavailable.
                                                                                                                                                                                                                   
Confidence values are already conventionalised in hooks.py: 0.95 error / 0.85 warning / 0.75 note for severity-tiered, 0.85 for summaries, 0.90 for confirmed-exploitable. Please match those rather than picking new numbers.
                                                                                                                                                                                                                   
Redaction: assemble the natural-language content first, then run it through core.security.redaction.redact_secrets() once before propose(). Easy! :smile:

Appreciate the help, this will be a strong addition to the codebase.

northdpole:
Done and aligned  to SAGE-native patterns

## What changed

- Removed standalone autonomous SQLite memory layer and related exports/tests:
  - `packages/autonomous/unified_memory.py`
  - `packages/autonomous/memory_store.py`
  - `packages/autonomous/memory_exports.py`
  - `packages/autonomous/tests/test_unified_memory.py`
- Kept memory integration SAGE-first and additive (no hard-fail behavior).

## SAGE implementation

Added new hooks in `core/sage/hooks.py`:

### Store hooks
- `store_crash_analysis_pattern(...)`
- `store_web_payload_effectiveness(...)`
- `store_codeql_build_reliability(...)`
- `store_fuzzing_strategy_outcome(...)`

### Recall hooks
- `recall_context_for_crash_analysis(...)`
- `recall_context_for_web_scan(...)`
- `recall_context_for_codeql_build(...)`
- `recall_context_for_fuzzing_strategy(...)`

### Domain/scoping helpers
- `_crashes_domain(...)` → `raptor-crashes-{repo_key}`
- `_web_domain(...)` → `raptor-web-{repo_key}`
- `_binary_key(...)` for crash/fuzzing context hints

## Workflow wiring

- `packages/binary_analysis/crash_analyser.py`
  - stores crash patterns to SAGE
- `packages/web/scanner.py`
  - recalls web context at start, stores payload effectiveness at end
- `raptor_codeql.py`
  - recalls build context at start, stores build reliability after run
- `raptor_fuzzing.py`
  - recalls fuzzing/crash context and stores strategy/crash outcomes

## Misc

- Natural-language content only for SAGE propose (no typed wire schema)
- Redaction once before propose via `_propose_redacted(...)` + `core.security.redaction.redact_secrets(...)`
- Confidence conventions aligned with existing hook conventions
- SAGE additive behavior preserved (`[]`/no-op when unavailable, debug logging on failures)
- `core/sage/CLAUDE.md` updated with `raptor-web-{repo_key}` and domain rationale <-- Do you want this or should I remove?

grokjc:
I like the direction this is going in!

A couple of structural things:

Why edit packages/autonomous/memory.py? The thing that worries me most is load() losing its try/except: a corrupt ~/.raptor/fuzzing_memory.json now raises out of \_\_init\_\_ instead of logging and starting empty, which feels backwards - the local layer should be more resilient than the optional one, not less. A couple of smaller behaviour shifts in remember() and get_statistics() too. Could this revert, or split out into its own PR?

The four new recall calls have no consumers: recall_context_for_web_scan, _fuzzing_strategy, _crash_analysis, _codeql_build are all called as bare expressions, return discarded. When SAGE is present we'd accumulate observations forever but nothing reads them. The existing recall_context_for_scan and enrich_analysis_prompt paths do feed real consumers, so the pattern's there, so we could wire one up (the codeql build recall into the auto-build prompt feels natural)? Or pull the four sites and land them when a consumer exists?

Last: on commit 1c7a182 - "remove autonomous SQLite layer and standardize on SAGE": FuzzingMemory was already JSON-backed before this PR and sticks around as the canonical layer. Maybe "add SAGE-backed memory hooks alongside FuzzingMemory" would be more accurate? Otherwise the local layer reads as deprecated. We have to accept that not everyone will use SAGE, awesome as it is!

As this is our first real use of SAGE I'll also invite comment from @l33tdawg here to make sure we're doing it right.

l33tdawg:
Hey guys - sorry for the delay; in Europe for Superbooth hahahah - Anyway, +1 to all three of @grokjc's notes - from the SAGE side specifically:
                                                                                                                                                                                                          
memory.py - Yes this should revert to main. The changes are residue from the SQLite layer that got reverted in 1c7a182, and they degrade behaviour independently of SAGE - load() losing the try/except means a malformed fuzzing_memory.json (or any schema drift) will TypeError on unknown/missing keys now crashes instead of starting empty.                                                                                                                                                                              

- remember() no longer preserves existing confidence / success_count / failure_count when re-storing under an existing key; it just overwrites the entry. That silently breaks the learning curve FuzzingKnowledge.update_success/failure is feeding.

- get_statistics() now reloads from disk on every call, which both adds I/O and clobbers any unsaved in-memory state.                                                                                   
                                                                                                                       
SAGE's value loop is recall -> enrich prompt -> LLM -> store. Calling recall_context_for_* and discarding the result spends the round-trip without any learning happening. The existing recall_context_for_scan + enrich_analysis_prompt flow is exactly the right shape.

FuzzingMemory is the canonical local layer; SAGE is the optional cross-tool / cross-repo recall layer on top of it. 
                                                                                                                                                                                                          
Otherwise this looks great :)

grokjc:
Thanks @l33tdawg - just wondering in terms of using SAGE have we got the right idea re remember and recall with the proposed changes? i.e. I think we're calling the right functions, is this how we best use it functionality-wise, or do you see other opportunities? Thanks!

l33tdawg:
Hey @grokjc - yeah the API surface you're hitting is right. `propose` / the `store_*` wrappers / `recall_context_for_*` are the intended primitives, and the way you're shaping the content (natural-language strings, observation vs fact, per-repo domain tags for findings + global tags for methodology) matches how SAGE wants to be talked to.

Two ways to get more out of it:

**1. Use the recall return value, not just call it.** The loop SAGE was built for is `recall -> enrich the prompt / bias the decision -> LLM or planner acts -> store outcome`. The good example already lives at `packages/llm_analysis/prompts/analysis.py:206` - `enrich_analysis_prompt` returns a string that gets folded into an `UntrustedBlock` and reaches the model. That's the canonical shape.

The four sites I flagged are calls without a downstream:
- `raptor_fuzzing.py:150` - strategy planner ignores the recalled strategies.
- `raptor_fuzzing.py:284` - crash triage doesn't see prior patterns for the same `signal+function` combo (highest-leverage one IMO; same `enrich_*` block pattern from static analysis would slot in cleanly).
- `raptor_codeql.py:91` - build planner doesn't see prior build failures.
- `packages/web/scanner.py:56` - fuzzer doesn't see which payloads worked previously on that target fingerprint.

`raptor.py:87` and `raptor_agentic.py:465` capture and print but never thread it into the LLM prompt - same shape, just one step short.

**2. Treat confidence as a knob, not a label.** Every recall comes back with a 0-1 confidence. A 0.85-confidence memory deserves different treatment from a 0.6 one. e.g. for the fuzzing strategy site, a high-confidence "dictionary-based mutation beat havoc on this fingerprint" should default the strategy unless overridden - that's what makes SAGE useful as a prior, not just a log.

**Other opportunity worth flagging:** the methodology-vs-findings split is mostly there but inconsistent. Findings correctly get per-repo `raptor-findings-{key}` (so they don't leak across repos), but methodology lessons (e.g. "dlmalloc heap UAFs need a heap groom first") should live in a single global `raptor-methodology` domain so they generalise. Worth a sweep once the recall flow is wired.

Otherwise the integration is in really good shape - wire the recall sites through to their consumers and SAGE actually starts learning across runs the way you guys want.

northdpole:
1. Dropped the parallel client-side SQLite / unified-memory path and kept persistence SAGE-first plus existing JSON FuzzingMemory where appropriate — no duplicate “second memory backend.” 
2. packages/autonomous/memory.py matches main again (including resilient load / prior behaviour); the earlier drift was residue from the reverted SQLite layer, not intentional SAGE work.
3. SAGE usage matches the intended API: natural-language propose / semantic recall, _propose_redacted + redact_secrets on the way out, _throttle, per-repo domains (raptor-findings-*, raptor-crashes-*, raptor-web-*) and global raptor-methodology where we meant methodology-only recall.
4. Implemented the four store + four recall hooks in core/sage/hooks.py and wired them from crash analysis, web scan, CodeQL, and fuzzing entry/exit paths.
5. Recall is actually consumed (not only invoked) on the paths we touched: analysis / agentic precall + UntrustedBlock, CodeQL build detector input, web fuzzer ordering, fuzzing planner + AFLRunner.extra_afl_flags (with RAPTOR_SAGE_AFL_PRIOR=0 to disable mechanical AFL priors), crash agent prior recall. core/sage/CLAUDE.md documents the AFL mechanical-prior behaviour.

For a bare raptor.py scan (no agentic subprocess), we still emit formatted recall to logs and sage_precall_scan.json for downstream tools, we do not fold that into an in-process scan LLM the way /agentic does. If you want that last hop too, say the word and I’ll do it as a follow-up on another PR.

grokjc:
Hey, thanks again for all the iteration on this.

While tracing the SAGE call sites, I noticed that the defensive try/except lives at the caller and five of the call sites don't wrap at all, so an exception there would crash the whole pipeline (codeql, web scanner, etc) Two more sites are wrapped but in pretty broad try blocks that end up also catching unrelated exceptions in the surrounding code.

I'm thinking it would be cleaner to just move the protection inside the recall_* / store_* in core/sage/hooks.py — wrap each function body, return [] or 0 on failure, logger.debug the error, etc. Then every caller is safe by construction, we don't depend on each future caller remembering to wrap, and the failure log message lives in one place rather than 26. 

If you can take care of that and then rebase vs main, we're good to go.

grokjc:
OK, I've fixed the merge conflicts (that was far from straightforward) so over to you @northdpole if you want to do the last mile on this one! Thanks.

grokjc:
Thanks @northdpole - you gave us a solid foundation.  As I decided to do a general SAGE uplift today, I've rebased onto main (which now has sage-agent-sdk 11.9.2) and added a few things on top of your work:

  - **Exception safety** in `_get_client()` - bare `except` was masking connection errors
  - **SAGE 11.9.2 features** - `tags` on `propose()` and `min_confidence` on `query()` wired through all hooks
  - **Validation + code-understanding hooks** - `recall_context_for_validation`, `store_validation_verdicts`, `store_validation_disproven`, plus map/trace/hunt recall and store for `/understand`
  - **Redaction fix** - `memory.py` was bypassing `redact_secrets()` on all three `propose()` calls
  - **Test coverage** - 105 tests passing, including the new hooks and edge cases

  Merging as-is (not squashing, to preserve your commit authorship).

Thanks again for the contribution.

PR #184, opened 2026-04-20
Enrich Exploitability Feasibility with Z3 SMT Solver

unprovable:
# Overview

SMT Solvers such as Z3 are very powerful ways of using constraint programming and analysis to validate exploitability paths. Through experimentation separate from raptor, I found that examining gadgets is incredibly easy in the LLM + python orchestration + SMT solver melange. 

To that end, I worked with Claude Code to embed Z3 into raptor's workflow. I tried a number of topologies, but the best is to use it as another analytical input going directly into the LLM's final analysis. This enriches the determination of exploitability significantly in testing. 

## Structured Overview

The public APIs and smt analysis code is largely agentically influenced. These basically tie in the SMT solver into python execution layer, specifically during Stage B. Whilst CodeQL identifies potential dataflow paths and LLMs hypothesize exploitability, Z3 is used to model program constraints—e.g. complex string sanitizers, integer bounds, conditional branches, etc. , as SMT formulas. By solving these formulas, Z3 determines if a path from a tainted source to a dangerous sink is logically reachable and provides the concrete values needed to trigger the bug.

This integration should allow raptor to automate discovery of specific payloads that bypass filters, a critical part of the Proximity Tracking. Instead of relying solely on LLM heuristics to guess if a sanitizer is effective, raptor can use Z3 to "solve" for the bypass. This creates a hybrid system where the LLM provides adversarial intuition and context, while Z3 handles the rigorous constraint solving required to prove exploitability.

## Example 

I was given a copy of `raptor_testbed` by @danielcuthbert and the known format string vulnerability gets the following output which shows how Z3 is running currently, and may be possibly widely extended in the framework for some great advantage:

Command: `/analyze test/data/bin1/ --smt`

>  Finding 1 — Format String ⚠️  Exploitable (line 192)       
>
>  printf(data);  // FMT: command handler — user input as format string
>
>  Attack path:                                              
>  1. FMT:%p.%p.%p... — leak stack/PIE/libc addresses                                                                                   
>  2. FMT:%Nc%hn writes — overwrite GOT[free] @ 0xe018 (67 write targets available)                                                     
>  3. Redirect to one-gadget 0xd515f (execve("/bin/sh", rbp-0x40, r13))            
>
>  SMT analysis (enabled): z3 solved the gadget constraint set. Best gadget: 0xd515f:
>  - Constraint: rbp-0x38 is writable → satisfiable (model: any writable stack/heap address)
>  - Constraint: rdi == NULL || {"/bin/sh", rdi, NULL} → satisfiable (rdi=NULL or controlled)
>  - This overrides the heuristic selection of 0x4c139 (posix_spawn requires rsp alignment + NULL rax/rbx — harder to guarantee via format string)                                                                                                                       
>
>Why exploitable despite PIE: format string provides both the leak and the write primitive in one shot. No separate leak stage needed.
>Chain breaks: __malloc_hook/__free_hook removed in glibc 2.34+ → use GOT or .fini_array instead.

## Future Testing and Work

This of course needs a LOT of testing, but SMT solvers are well trodden for vulnerability analysis (they're something I've use a lot myself - "need an exploit? just add maths!"). As such, next steps for this PR might be:

1. Determine if the addition of `z3-solver` is reasonable to add... it's a minor hit on space and runtime for potentially large gain, but someone else should review this work. 
1. See if there are additional use cases that SMT solvers might be good for and build those into the pipeline
1. Currently this is mostly geared for OneGadget analysis. This might also be useful in other areas:
   1. ROP chain analysis
   1. CodQL Output Analysis (where I've seen some good results in other projects)

**DIsclaimer** - a lot of the code here in `smt_sanitizer.py` and a good chunk of the logic in sorting out crash data from gdb is agent written, human checked. Most of hte rest of the code was started by an agent and then added to by me, often with co-pilot autocomplete helping along the way. It's not been formally verified by itself, but it follows a well established workflow.


unprovable:
Ok - I've now exhausted my initial testing. I have added a prompted implementation of CWE-190 bugs in the last two commits. These are based on asking Claude to interpret the ideas in this paper: https://arxiv.org/html/2604.05292v1 

I asked 'what more could we do with SMT's in analysis?' and the result was this list: 

>   1. CWE-787/125 bounds analyzer — alloc-size expr vs. access-offset expr → prove OOB. Sibling module. Reuses the bitvector infra.
> 1. Path-sensitive sanitizer analysis — current code conjuncts all guards globally. Enumerate CFG paths through intermediate_steps and solve per-path; report unguarded paths explicitly.
> 1. String/regex sanitizer modeling — z3.String / z3.Re for contains/startswith/escape/allowlist patterns (XSS, SQLi, SSRF hostname checks).
> 1. Flow-trace solver for /understand --trace (flow-trace-*.json) — same bitvector engine, reused. Makes SMT an asset shared by /understand and /validate.
> 1. Concolic corpus gen in packages/autonomous/corpus_generator.py — solve unexplored branch constraints to produce AFL++ seeds.
> 1. Primitive/mitigation chain solver in exploit_feasibility — symbolic model of "do these primitives + these mitigations admit code exec?" Complements one_gadget.

So this is possibly part of a longer tail of issues, but hte OneGadget stuff is quite useful for now :) Cheers! M.

grokjc:
Yes, my initial thought is to make a /core/smt-solver/ component that is self contained and neatly abstracted so that all manner of consumers can use it and we get maximum value from the addition to the codebase. Any opportunity to offload LLM reasoning onto mechanical work that doesn't need tokens is a good one if we architect it properly.

unprovable:
Yeah, this makes a lot more sense :) Shall I make the necessary changes and update this PR? I was originally considering making it a package in`packages/smt/` but maybe adding it as a core is something that makes it a lot more portable. I have a load of tokens to play with so can see test it imminently. Let me know if I shouldn't go ahead 😄

grokjc:
Up to you but it might be easier to park this PR which is really a consumer of the solver, do a separate piece of work in making a core component, and then refactor this PR to use it once it is merged? small chunks, easily reviewed, less for you to juggle ... ?

unprovable:
Sure I can park this for now. M.

PR #188, opened 2026-04-21
Enhancing Vuln Analysis with SMT - Adding SMT Core

unprovable:
Hello, As per request by @grokjc in #184 - this PR should introduce the SMT solver core into claude, alongside a simple use case that proves onegadget usage. This runs well on the raptor_testbench.c provided by @danielcuthbert so hopefully this can be a start for building out SMT usage to enhance Raptor's capabilities and assurance of finding viable exploits. 

Cheers! M.

### DIFF
--- a/.claude/skills/exploitability-validation/SKILL.md
+++ b/.claude/skills/exploitability-validation/SKILL.md
@@ -209,6 +209,81 @@ STAGE 1:  Recompute CVSS, validate, report
 
 ---
 
+## SMT Add-ons (Z3)
+
+Three SMT-backed analyses run alongside LLM heuristics when `--smt` is passed to `raptor.py analyze`. All are soft dependencies — if `z3-solver` is not installed or `RAPTOR_SMT_ENABLED` is not `"1"`, they silently return `smt_available=False` and behaviour is identical to the non-SMT path.
+
+**Architecture.** Shared Z3 primitives live in `core.smt_solver/` (availability gate, bitvector factories, signed/unsigned comparator routing, timed solver construction, two's-complement witness formatting). Domain encodings live next to the callers in `packages/*/smt_*.py` — they import the primitives and focus purely on pattern extraction and predicate construction.
+
+| Layer | Module | Responsibility |
+|-------|--------|----------------|
+| Harness | `core.smt_solver` (`availability`, `config`, `bitvec`, `session`, `witness`) | `smt_enabled()`, `mk_var/mk_val`, `le/lt/ge/gt(…, signed=…)`, `new_solver(timeout_ms=…)`, `format_witness(model, signed=…)` |
+| Encoding | `packages/codeql/smt_sanitizer.py` | Extract guard regexes, negate them, check joint bypass SAT |
+| Encoding | `packages/codeql/smt_integer_overflow.py` | Extract arithmetic at sinks, apply `BVAddNoOverflow / BVSubNoOverflow / BVMulNoOverflow` |
+| Encoding | `packages/exploit_feasibility/smt_onegadget.py` | Parse one-gadget register/memory constraints, check feasibility against a crash state |
+
+Environment knobs (read by every call, not cached):
+
+- `RAPTOR_SMT_ENABLED=1` — master gate. All three modules no-op when unset.
+- `RAPTOR_SMT_WIDTH=32|64` — default bitvector width (default 64). Integer-overflow sessions override per-declared-type.
+- `RAPTOR_SMT_SIGNED=signed|unsigned` — default signedness (default signed). Integer-overflow sessions override per-declared-type; size_t-context sinks (`malloc`, `calloc`, `realloc`, `memcpy`, `memmove`, `memset`) default to bv64-unsigned.
+
+### One-gadget constraint satisfaction (`smt_onegadget.py`)
+
+Invoked automatically inside `analyze_binary()` when one_gadget offsets are found. Evaluates whether the register/memory constraints printed by `one_gadget` (e.g. `[rsp+0x30] == NULL`, `r14 == NULL || r14 == (char**)(&argv)`) can be satisfied — optionally given a concrete GDB crash state.
+
+Results appear under `analysis.one_gadget_info.smt_feasibility` in the exploit context:
+
+```json
+{
+  "feasible": true | false | null,
+  "reasoning": "...",
+  "satisfied_constraints": ["[rsp+0x30] == NULL"],
+  "unsatisfied_constraints": [],
+  "model": {"rsp": 140737...}
+}
+```
+
+- `feasible: true` — Z3 found satisfying values; the one-gadget is usable under the right conditions.
+- `feasible: false` — constraints are contradicted by crash state or mutually exclusive; ruling out this gadget is reliable.
+- `feasible: null` — Z3 returned `unknown` or all constraints unparseable; fall back to heuristic `note`.
+
+**Stage E:** When `smt_feasibility.feasible` is `false` for the best-ranked gadget, treat it as a strong chain break signal (stronger than the heuristic note, which remains present for comparison).
+
+### Sanitizer bypass analysis (`smt_sanitizer.py`)
+
+Invoked automatically inside `DataflowValidator.validate_dataflow_path()` before the LLM call. Reads source files to extract guard conditions (length checks, null checks, range checks) in C/C++ and Python, then asks Z3 whether the bypass condition (negation of the guard) is satisfiable.
+
+The SMT result is **combined with the LLM verdict** post-generation:
+
+| SMT result | Effect on `DataflowValidation` |
+|------------|-------------------------------|
+| `bypass_found: True` + concrete input | `confidence += 0.10`, `bypass_possible = True`, reasoning annotated |
+| `bypass_found: False` | `confidence -= 0.10`, barrier appended, reasoning annotated |
+| `bypass_found: None` (unknown/unparsed) | No adjustment |
+
+The LLM prompt also receives `SMT PRE-ANALYSIS:` with the reasoning, so the model has the SMT verdict as context.
+
+**Stage C/D:** Findings with `bypass_found: True` and a concrete bypass input in the reasoning (`SMT bypass input values: {...}`) are stronger candidates for Stage D confirmation. Findings where SMT proved all sanitizers effective get a confidence penalty that may tip borderline findings toward `ruled_out` at Stage D.
+
+### Integer overflow reachability (`smt_integer_overflow.py`)
+
+Invoked automatically inside `DataflowValidator.validate_dataflow_path()` immediately after the sanitizer pass, independently of it. Extracts binary arithmetic (`+`, `-`, `*`) at recognised sinks — `malloc/calloc/realloc` size, `memcpy/memmove/memset` length, and declared-integer assignments like `unsigned int x = a + b;` — then asks Z3 whether the expression can wrap past its fixed-width boundary under upstream guards. Uses `BVAddNoOverflow / BVSubNoOverflow / BVMulNoOverflow` (and their underflow siblings for signed arithmetic) as the overflow predicates.
+
+Width/signedness come from the declared type when present (`unsigned int` → bv32-unsigned, `size_t` → bv64-unsigned, full LP64 table in `_TYPE_WIDTH_HINTS`), otherwise from the size_t-context default for memcpy/malloc family sinks, otherwise from `RAPTOR_SMT_WIDTH` / `RAPTOR_SMT_SIGNED`. Upstream guards are re-encoded at the session's width so the overflow session can run at bv32 even while the rest of the pipeline is at bv64.
+
+The result is injected into the LLM prompt as `SMT INTEGER OVERFLOW: <reasoning>` plus (on SAT) a concrete `SMT overflow input values: {...}`. It is **advisory**: unlike the sanitizer pass, the integer-overflow verdict does not adjust `DataflowValidation.confidence` or append barriers automatically — the LLM decides what weight to give it.
+
+| SMT result | Effect |
+|------------|--------|
+| `overflow_found: True` + concrete input | Prompt annotated with witness; LLM weighs alongside other signals |
+| `overflow_found: False` | Prompt annotated as UNSAT under modeled guards |
+| `overflow_found: None` (unknown / no arithmetic / z3 off) | No annotation |
+
+**Stage C/D:** When `overflow_found: True` on a sink like `malloc(size + N)` or `memset(buf, c, size + N)`, the witness is a direct testcase input — Stage A/C PoCs should try that concrete value first. When `overflow_found: False` under guards the pipeline already modelled, treat it as evidence toward `ruled_out` at Stage D, but cross-check that the guards are actually on-path (the extractor is regex-based and does not understand scoping).
+
+---
+
 ## Integration with Exploit Feasibility
 
 Stage E automatically bridges to the `exploit_feasibility` package for memory corruption vulnerabilities.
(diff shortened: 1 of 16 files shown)


grokjc:
Thanks, I think this will be useful for many things 😄 Before I jump into a proper review: a) requirements.txt will need something adding and b) we probably need some test cases in /core/smt_solver/tests/ too!

unprovable:
ok I added some basic tests and updated the requirements.txt. These are a bit more by-hand. One test should come back as 'SAT' and the other should come back as 'UNSAT' basically validating that the solver is working. These are the only two outcomes that matter, so these are all I can think of for now... I also added a onegadget test from my testing last week that should check parsing gadgets (this is in `packages/exploit_feasibility/tests/test_smt_onegadget.py`) 

Thanks for your help and encouragement! M.

requirements.txt
```diff
@@ -19,6 +19,9 @@ instructor>=1.0.0
 # Optional: For enhanced dataflow visualization (recommended)
 tabulate>=0.9.0
 
+# for Z3 solver
+z3-solver>=4.16.0.0
```
grokjc: I think this should be present but commented out so it is optional - sorry, should have been clearer before!

requirements-dev.txt
```diff
@@ -13,3 +13,6 @@ mypy>=1.0.0
 pytest>=7.0.0
 pytest-cov>=4.0.0
 gcovr>=5.1
+
+# test for Z3 solver
+z3-solver>=4.16.0.0
```
grokjc: I'm soon starting a side quest to pin exact versions of things. You can help me by doing that here too 😄

.claude/skills/SecOpsAgentKit
grokjc, line 1: Did you mean to delete this?!?

.claude/skills/exploitability-validation/SKILL.md
grokjc, line 1: Two mismatches with the code:
  - `--smt` flag isn't in raptor.py.
  - smt_sanitizer.py and smt_integer_overflow.py (plus DataflowValidator wiring) aren't in the diff.

.claude/skills/exploitability-validation/stage-e-feasibility.md
grokjc, line 1: Same `--smt` mismatch. smt_feasibility only appears on memory-corruption findings - worth saying so

core/smt_solver/availability.py
grokjc, line 1: Drop `smt_enabled()` and the env var - `z3_available()` is the right gate.
Catch `Exception` not just `ImportError` so a broken native load degrades gracefully too.

core/smt_solver/config.py
grokjc, line 1: If RAPTOR_SMT_ENABLED goes, the width/signed env vars should too (wire these through a config dataclass or function args instead?)

core/smt_solver/session.py
grokjc, line 1: `new_solver()` is defined but smt_onegadget doesn't use it - timeouts aren't actually enforced.

packages/exploit_feasibility/tests/test_smt_onegadget.py
grokjc, line 1: The z3 mocks alias z3.unsat to solver.check.return_value, so the equality checks pass regardless of parser behaviour. 

Worth adding parser-level tests (atoms, disjunctions, unparseable input) ... ?

packages/exploit_feasibility/__init__.py
grokjc, line 1: worth logging at debug priority rather than silently dropping?

packages/exploit_feasibility/analyzer.py
grokjc, line 1: Unrelated to SMT - separate PR?

Also: ldconfig -p and the hardcoded paths read the host, not the QEMU target, so the stated use case may not actually be fixed.

packages/exploit_feasibility/api.py
grokjc, line 1: - `except Exception: pass` should log at debug.
- The inner re-import is redundant (package __init__ already exports).

Use `g['offset']` / `g['constraints']` rather than `.get(..., default)` -  the upstream serializer at api.py:306 always populates those fields, so a missing key is a real bug worth surfacing, not a default-to-zero case.

packages/exploit_feasibility/smt_onegadget.py
grokjc, line 1: - Parser only covers `== NULL` / `is NULL` / `is writable`. Real one_gadget output also has numeric equalities, `!= 0`, `valid argv`, `or` disjunctions - most real gadgets will fall into `unknown`. 
- BitVec widths hardcoded to 64; use mk_var/mk_val.
- Three raw z3.Solver() calls — use new_solver().
- Witness extraction duplicates format_witness.
- _MEM_RE accepts unbalanced brackets and rejects `[rsp]` with no offset.

grokjc (commented):
The RAPTOR_SMT_ENABLED gate duplicates what `z3_available()` already tells us. If z3 imports → SMT runs; if not → silent no-op. No env var needed. Removing it simplifies availability.py, config.py, and the docs in one go.

Also: the submodule delete (.claude/skills/SecOpsAgentKit) looks accidental.

requirements.txt
unprovable: Ok, this is done.

requirements-dev.txt
unprovable: Aha - pinned to precise version. 4.16.0.0 works well when I tried it just now, so sticking to that.

.claude/skills/SecOpsAgentKit
unprovable: No!!! Sorry... I think there was a mess up when I was moving things around. I can add it back, but it won't link to the required commit. Not sure what to do here, if I'm honest.

core/smt_solver/availability.py
unprovable: ok I think this is done.

core/smt_solver/config.py
unprovable: Made a dataclass that is paranoid af (using `Literal` and `__post_init__` to make sure someone doesn't try to pass x86_42 as a real thing). I also rejigged the tests to line up and pytest passes on `test_smt_onegadget.py` and `test_smt_solver.py`

core/smt_solver/session.py
unprovable: Great catch - this should be fixed now...

packages/exploit_feasibility/__init__.py
unprovable: Done. HAd to import get_logger so hope I didn't put the import somewhere egregious.

packages/exploit_feasibility/analyzer.py
unprovable: I hadn't noticed the diff, so it's unintentional. Please ignore!

core/smt_solver/session.py
unprovable: Fixed - good catches!

packages/exploit_feasibility/__init__.py
unprovable: Fixed. Hope I added the logger import to a reasonable location.

packages/exploit_feasibility/api.py
unprovable: Done - yeah, I was overly safe and burning cycles. Exception logger added for debug level.

packages/exploit_feasibility/smt_onegadget.py
unprovable: so dealign wit hthe regex first, it also misses `[rip]` and accepts things like `[r15-0x20`, you're right. Updated the regex with something I think should work more broadly., but maybe a bit too broadly... 

No raw z3.BitVec calls should remain, as well, using mk_var/mk_val as needed. 

Also tried to update the atomic parser with additional cases:
* numeric equalities/inequalities (`rax == 0xbeef, rsi != 0` or sim), 
* register/memory equalities (`r12 == rbx, rax == [rsi]` or sim), 
* bitmask alignment (`rsp & 0xf == 0` or sim)

packages/exploit_feasibility/tests/test_smt_onegadget.py
unprovable: So Claude and I came up with a load of new test cases based on the revamped parser logic. The result is now commited to `packages/exploit_feasibility/tests/test_smt_onegadget.py` for some parser-level checking of Z3's functionality. Passes under pytest for me.

.claude/skills/exploitability-validation/SKILL.md
unprovable: This is an error on my part. Great catch - fixed.

.claude/skills/exploitability-validation/stage-e-feasibility.md
unprovable: sdame as the SKILLS.md error... good catch! Fixed.

unprovable:
Hello! Thanks so much for such a careful review! There are two outstanding items I haven't yet been able to resolve: 

[packages/exploit_feasibility/analyzer.py](https://github.com/gadievron/raptor/pull/188/files/28c8b03040971958984169e2ebe50589df5c90eb#diff-15b32f3fa6e3efc709c54ee13fbefd1cbe448af68ff2216cd00c55592ad2eef8) <-- not sure this was wholly intentional from me? Happy to revert.

[.claude/skills/SecOpsAgentKit](https://github.com/gadievron/raptor/pull/188/files/28c8b03040971958984169e2ebe50589df5c90eb#diff-29932c21684487cc4914445317c5d1508984cafd1ab5cf837bc8ad70bfd5d4be) <-- removing this was a total accident. I'd fix it here but it's probably going to change the commit it points to so haven't touched it yet.

.claude/skills/SecOpsAgentKit
unprovable: Fixed it! With a little help from a friend. M.

unprovable:
PS - fixed .claude/skills/SecOpsAgentKit to be at the right commit (394ee0d0dd7c97e0ddc273b989cd657528544ad9) and an added submodule. Hope this resolves that issue! Just the mystery of packages/exploit_feasibility/analyzer.py left :)

unprovable:
Ok, removed that code as I could see what I was trying to do but it was related to something else and as you pointed out... didn't work! Many apologies for that confusion. Removed it, stashed a copy to work on later.

packages/exploit_feasibility/analyzer.py
unprovable: Fixed the issue - it was an attempt to get libc data to determine the grammar of what is coming into the parser. As you pointed out, it's not actually working, so will park for now.

unprovable:
Ok gone over stuff a lot and I think this is now pretty much done and ready as a first draft. Back to you, @grokjc. 🙏 M.

core/smt_solver/availability.py
```diff
@@ -15,16 +12,16 @@
 try:
     import z3  # type: ignore
     _Z3_AVAILABLE = True
-except ImportError:
+except ImportError as i:
     z3 = None  # type: ignore[assignment]
     _Z3_AVAILABLE = False
+    print(f"An import error occurred: {i}")
```
grokjc: Z3 is advertised as a soft optional dependency, but both `except` branches `print(...)` at import time. On any machine without z3 (CI, users who never touch exploit-feasibility) every Python invocation that transitively imports `core.smt_solver` prints `An import error occurred: No module named 'z3'` to stdout. Either drop the prints entirely or demote them to `logger.debug(...)`.

core/smt_solver/config.py
```diff
@@ -1,34 +1,42 @@
...
+
+@dataclass
+class SMTSolverBVConfig:
+    """Dataclass for SMT Solver Bitvector Config"""
+    width: Literal[32, 64] = 64
+    signed: bool = True
+
+    def __post_init__(self):
+        if self.width not in (32, 64):
+            raise ValueError(f"width must be 32 or 64, got {self.size}")
```
grokjc: self.size doesn't exist — attribute is self.width. The ValueError path itself raises AttributeError during f-string formatting.

packages/exploit_feasibility/smt_onegadget.py
```diff
@@ -28,11 +34,31 @@
...
+from core.smt_solver.witness import format_witness as _format_witness
 
 from .context import OneGadget
 
 
+# x86_64 registers are natively 64-bit, and the values we care about are
+# addresses — treat them as unsigned in any rendered witness.
+
+# First change the dataclass for the other functions to pick up
+_Z3.signed = False
```
grokjc: This mutates a module-private global in core.smt_solver.config at import time. Any future consumer of core.smt_solver.bv_width() / is_signed() now inherits bv64u purely as a side effect of smt_onegadget having been imported first — including the tests in core/smt_solver/tests/test_smt_solver.py.

The mutation has zero functional purpose here: every call site in this file already passes width/signedness explicitly.

You can safely delete lines 45 and 55-56; keep _REG_WIDTH = 64 / _REG_SIGNED = False as plain module constants.

grokjc (commented):
Looking good, thanks for fixing all the other stuff. There's just three things left that I'd consider blockers now (and none of them major!)

core/smt_solver/availability.py
unprovable: This should be resolved in [this commit](https://github.com/gadievron/raptor/pull/188/commits/2f115337b97e42428ce97482f28636621f6a03a8)

packages/exploit_feasibility/smt_onegadget.py
unprovable: Fixed in [this commit](https://github.com/gadievron/raptor/pull/188/commits/b45c747f57cec6ae8e797a69f1b4ae1c34198871)

core/smt_solver/config.py
unprovable: Fixed in [this commit](https://github.com/gadievron/raptor/pull/188/commits/a6712d343708de6738fb56304b3730c40a7cb7c4) 

PS - sorry about the earlier comment, I was totally mis-reading the line 😅)

unprovable:
Many thanks for your help! Really clear comments - really easy to follow :) LEt me know if there are any other wrinkles that need to be fixed. M.

core/smt_solver/availability.py
grokjc, line 1: get_logger is a factory function (core/logging.py:173 → returns RaptorLogger()), not a logger instance. logger.debug(...) is attribute access on a function object. 

Fix - match the pattern already used in packages/exploit_feasibility/__init__.py:
```python
  from core.logging import get_logger
  ...
  except ImportError as i:
      z3 = None
      _Z3_AVAILABLE = False
      get_logger().debug(f"z3 import failed: {i}")
```

grokjc (commented):
Almost! 😄 One thing left to fix and we're good to go.

core/smt_solver/availability.py
unprovable: Ah, yes! Sorry, my bad. Great catch. Fixed in [this commit](https://github.com/gadievron/raptor/pull/188/commits/d324f2d5567b313f726630538a281f0ff30c780c)

unprovable:
Aha! That's a daft error worth fixing. 😅

grokjc:
Merged! Thank you.

I have a few followup tweaks in mind, which will take me a little while and then we can start using this new core component. I have some ideas I'll run past you when I get a chance - soon!

But for now, thanks for sticking with this and fixing the issues - appreciate all the hard work in getting us this far.

unprovable:
You are most welcome, and yeah definitely open for helping out. 😄 I have a load of ideas for CodeQL use cases with a solver once it's fully integrated and everyone is happy with this baseline. `one_gadget` is the most natural place to start as it's literally a constraint-based output, so putting a constraint solver in the loop should make the LLM output much more powerful. I'm sure there will be many tweaks and extensions - indeed, SMT solvers and LLMs seem to be [having a moment](https://arxiv.org/abs/2603.06931v2) 

Looking forward to doing more with yourself and this project! M.

PR #190, opened 2026-04-21
feat: SAGE persistent memory integration (re-land against sage_integration)

l33tdawg:
Re-raising the SAGE integration against the `sage_integration` branch per @grokjc's request on #156.

Same content as PR #156 at `daee3aa`, cherry-picked onto `sage_integration` cleanly (10 commits, no conflicts). Plays nicely with #187 (cc_trust env.SAGE_* protection) which is already on this branch.

## What's in this PR
- SAGE client/config/memory + hooks under `core/sage/`
- Python pipeline wired into `raptor.py` and `raptor_agentic.py`
- 12 `.claude/commands/*.md` with memory-recall prompts (SAGE MEMORY boilerplate consolidated into CLAUDE.md)
- `docker-compose.sage.yml` pinned to `ollama/ollama:0.21.0` and `ghcr.io/l33tdawg/sage:6.5.5`
- `requirements.txt` hint pinned to `sage-agent-sdk==6.5.5`

## Addresses previous review rounds
- Round 1: file reorg into `core/sage/`, removed per-agent boilerplate, namespace/auth/staleness answers
- Round 2: port mismatch fix (8080→8090), `_run_async()` for sync-to-async, command file boilerplate stripped
- Round 3: missing `import asyncio`, event loop FD leak, pinned versions, dead `recall_exploit_primitives()` removed

## Tests
47 SAGE tests pass:
```
core/sage/tests/test_sage_client.py  ... 10 passed
core/sage/tests/test_sage_config.py  ...  9 passed
core/sage/tests/test_sage_hooks.py   ...  8 passed
core/sage/tests/test_sage_memory.py  ... 20 passed
```

Happy to run the end-to-end test you mentioned once you've pushed your follow-up changes on top.


l33tdawg:
Heads up — pushed a follow-up commit (5c5238b) after running an end-to-end against a dockerized SAGE on :8090.

**What I found during e2e**
The async bridge in the original integration was dropping every SAGE call after the first per process. Root cause: `AsyncSageClient` under the hood holds an `httpx.AsyncClient` whose connections are loop-bound, and RAPTOR's `_run_async()` creates+closes a fresh loop each call — so the cached SDK client's httpx pool referenced a closed loop. The outer `try/except` in each hook silently swallowed the `Event loop is closed` error, which is why the unit tests (all mocked at `AsyncSageClient` wholesale) never caught it.

Concretely:
- `recall_context_for_scan` does two back-to-back queries (findings + methodology) — only the first ever returned data
- `enrich_analysis_prompt` is called once per finding — only the first finding ever got SAGE context
- `memory.py`'s `save/remember/record_campaign` had the same pattern

**Fix**
Swap to `sage_sdk.client.SageClient` (the sync variant — the SDK ships both). Same method surface (`health` / `propose` / `query` / `embed` / `register_agent`), no event loop plumbing. Dropped `_run_async`, all `asyncio` imports, and the `_*_async` helpers in `hooks.py` / `memory.py`. Net: −424 / +256.

**Verified**
50 tests pass (was 47 — added a regression test that two back-to-back queries both succeed using the cached SDK client, plus a positive `store_scan_results` test). Against docker SAGE :8090 with real findings: 5 memories recalled, 416 chars of enriched context, three back-to-back `enrich_analysis_prompt` calls all return 444 chars (previously the 2nd+ returned empty).

Happy to discuss if you'd rather stay on the async SDK — this was the smaller/cleaner fix.

grokjc:
Looking good! I have a few followup PRs to land now, will keep you posted from an E2E test perspective and we are on track to get this merged into main ASAP. Thanks for all the work getting us this far.

l33tdawg:
No worries mate :) Thanks for accepting my contribution

grokjc:
In core/sage/hooks.py there's a hardcoded time.sleep(0.3) between every client.propose() call (lines 146, 225) with the comment "avoid overwhelming single-node consensus". Two questions:

1. Does sage_sdk.client.SageClient expose a batch/bulk propose API we could use instead of N separate calls?
2. Is the 300ms still needed with the sync client, or was it a workaround for the old async-bridge behaviour? On a 20-finding scan it adds ~6s of pure sleep.

If neither batch nor "always needed", I'd suggest making it configurable via SAGE_PROPOSE_DELAY_MS (default 300) so multi-node users can zero it.

Happy to make changes (don't need a PR!) but would like your advice please.

l33tdawg:
Great catches on both fronts.

**1. Batch API:** No — the SDK only exposes single-item `propose()` and the server only has `POST /v1/memory/submit`. No bulk endpoint exists today. A batched tx would be a nice addition but it's a protocol-level change, not something you can work around client-side.

**2. The 300ms:** You're right that it's a leftover. Commit `5c5238b` ("drop async bridge") mechanically swapped `await asyncio.sleep(0.3)` for `time.sleep(0.3)` and relabeled the comment — but the justification doesn't hold against the sync path. `POST /v1/memory/submit` uses CometBFT's `broadcast_tx_commit`, which already blocks until the block containing the tx is finalized. Block cadence is 1s (personal) / 3s (quorum), so the 300ms is noise on top of a gate consensus already imposes. It's not protecting anything.

**Suggestion:** go ahead and drop both sleeps. If you want the env knob as a safety valve, `SAGE_PROPOSE_DELAY_MS` with **default 0** is the right shape — the old 300 default was buying nothing.

One heads-up while you're in there: on a 20-finding scan the sleeps are ~6s, but the sequential `broadcast_tx_commit` round-trips themselves are ~20s personal / ~60s quorum (one block per propose). If that latency matters, running the proposes concurrently (`ThreadPoolExecutor`, a handful of workers) collapses them into 1-2 blocks of wall time since CometBFT will batch them into the same block. Separate change, but it's the real lever if you want the scan faster.

Appreciate you picking this up.

grokjc:
In core/sage/hooks.py keys SAGE queries/storage by Path(repo_path).name as same-basename repos at different paths cross-contaminate. Planning to switch to sha256(resolved_path)[:12] as the primary key, with active RAPTOR project name (if any) attached as a tag, not the key - as projects are campaigns, not codebase identity. 

Does SAGE have a built-in agent-identity/namespace mechanism that already does this that we should use instead? Also, does SAGE propose support tags on memories in a way that filters recall?

l33tdawg:
Good catch — fixed in `core/sage/hooks.py` (commit a86ac5a on `feat/sage-integration-v2`). Switched to `sha256(resolve(repo_path))[:12]` as the scope key, but folded it into `domain_tag` rather than tags:

- `raptor-findings` → `raptor-findings-{repo_key}` (same for `raptor-exploits`)
- `raptor-methodology` stays global — it's cross-repo generalizable knowledge, not repo identity
- `enrich_analysis_prompt` gained an optional `repo_path` param and short-circuits without it (unscoped recall would leak cross-repo, defeating the whole point)
- Tests updated, 11/11 pass

Re: per-repo `AgentIdentity` — SAGE already has agent-identity: `submitting_agent` is stamped on every write, RBAC-based visibility on reads. That's a stronger primitive than a path-hash because it's cryptographic. If you want raptor to use that instead (or in addition) — deterministic per-checkout `AgentIdentity` file, register once, let visible_agents do the namespacing — worth doing as a separate PR since it's a real shape change (key material per checkout, registration flow). Lmk if that's the direction you want and I'll open an issue / draft it.

grokjc:
Can you cherry-pick a86ac5a onto a fresh branch and open a new PR targeting the integration branch? The merged version of #190 and your working branch have diverged after the squash, so a direct push won't work.

Agent identity is strong, but you already solved *project identity* with the above commit. All good.

Also interested in your thoughts/testing re my install/uninstall script from earlier today if you get a chance. Feels like we're almost there now!

l33tdawg:
Gave `libexec/raptor-sage-setup` a full shakedown — install, uninstall, JSON state machine in 5 pre-populated scratch dirs. Overall solid; two small bugs worth picking up:

1. **Healthcheck image mismatch.** `docker-compose.sage.yml` calls `curl` but the SAGE image ships with `wget` only — container shows `(unhealthy)` forever even though `/health` is fine.
2. **Registrar method typo.** `register_agents.py:299` calls `client.register(...)`; the SDK method is `register_agent`. Bare except hides it. The 16 agent proposes still land, but the registrar identity never stamps.

Both 1-liners, opened as #199 targeting `sage_integration`.

Separately, fixing #2 surfaces a 3-way SAGE-side inconsistency on `AgentRegistration.registered_at` (Go: `int64` block height, OpenAPI: `date-time string`, Python SDK: `str | None`). That's a SAGE repo fix, not raptor — raising on the SAGE side. Our call here becomes correct automatically once that lands.

Plumbing cases all pass (deep-merge preserves other MCP servers / other env keys; empty-after-uninstall cleanup works; double-install is idempotent; uninstall-with-nothing is a no-op). Full e2e install+uninstall also clean — containers + network torn down, volumes preserved as documented.

Happy with where this is.

grokjc:
One loose end: SAGE_FALLBACK_JSON is declared in core/sage/config.py, documented in SAGE_INTEGRATION.md, and flagged by cc_trust - but grep -rn fallback_json core/sage/ shows no consumer. Intentional scaffolding for a follow-up, or orphaned scope we can delete? Happy either way.

l33tdawg:
delete it — matches current behavior. `_sage_available` is doing the whole job already; the flag is vestigial from an earlier design where we were going to support SAGE-only (no JSON cache) deployments. We never needed it because the write-both-on-success pattern turns out to be the right failure mode: if SAGE is healthy, you get SAGE + JSON; if SAGE's down, you still have JSON. Adding `fallback_json=false` would only help someone who wants to skip JSON when SAGE is healthy, which is a thin win for a trap (silent data loss if SAGE goes down mid-run).

Happy to take the cleanup PR if you want to swing at it — remove the field from `config.py`, the doc row in `SAGE_INTEGRATION.md`, the `test_cc_trust.py` allowlist entry, and the two test cases in `test_sage_config.py`. Otherwise I'll queue it.

grokjc:
Done - #214 merged.

grokjc:
sage_integration has landed eight PRs since your last round of testing. Worth another pass if you have time to test?
                                         
  1. Idempotency — run libexec/raptor-sage-setup once, then again. Second run should be mostly skipped: output with zero duplicates in SAGE.                                                                       
  2. Parallelism — compare wall-time against your prior #199 timing. asyncio.gather(Semaphore(8)) should collapse the ~100-entry seed to 1–2 CometBFT blocks total instead of ~100.
  3. Partial-failure recovery — Ctrl-C mid-seed, re-run without --force. Only the missing entries get proposed; partial agents (role stored, xref crashed) come back as partial.                                   
  4. --force — re-proposes everything regardless of existing tags.                                                                                                                                                 
  5. Thread-safety — /agentic with SAGE up; orchestrator's ThreadPoolExecutor should not blow up concurrent hooks; debug logs show exactly one SageClient construction.                                            
  6. Registration telemetry — setup's registration line now prints (on-chain height N). None there means SAGE server/SDK version drift.

l33tdawg:
On it...

l33tdawg:
Ran through all six checks against latest `sage_integration` (`0d4b080`) — the run surfaced a SAGE-side idempotency bug that swallows concurrent tag writes. Fix is in SAGE v6.6.4 (https://github.com/l33tdawg/sage/releases/tag/v6.6.4); bump PR on raptor follows.

### 1. Idempotency

**Against `sage:6.6.0` (the current pin) this FAILED.** First run stored 87/87 seed + 16/16 agents. Second run re-stored 82/87 seed (only 5 skipped) and re-stored 12/16 agents (1 skipped, 3 partial). Root cause: `propose(tags=...)` returned 200 but SAGE's post-commit `SetTags` silently failed — server logs had 197× `failed to set tags on committed memory: "clear tags: database is locked (5) (SQLITE_BUSY)"` and 85× `/v1/memory/list 500`. `async_memory_exists` then couldn't find memories via tag lookup so the script re-proposed them.

Two SAGE bugs chained together:
- SQLite DSN pragmas (`_journal_mode=WAL&_busy_timeout=15000`) silently dropped by `modernc.org/sqlite` — it wants `_pragma=journal_mode(WAL)` syntax. DB has been running in rollback-journal mode with `busy_timeout=0` since the driver switch. This is the root cause behind the 6.6.1 symptom (which papered over a different path with retries). Confirmed by `sqlite3 sage.db "PRAGMA journal_mode"` → `delete`, no `-wal` sidecar.
- `SetTags` and 5 other store methods opened transactions via raw `s.db.BeginTx`, bypassing `writeMu` that serializes writes. With busy_timeout=0, every concurrent writer hit instant SQLITE_BUSY.

**Against `sage:6.6.4`:** Run 1 87 stored, 16 stored. Run 2 `Stored: 0/87 Skipped: 87 Failed: 0` and `Stored: 0/16 Partial: 0 Skipped: 16 Failed: 0`. Zero server warnings.

### 2. Parallelism

Warm wall-time: `~3.2s` for 87 seed + 16 agents on 6.6.4 (Run 2 all-skip path). First cold run `~20s`, dominated by Ollama `nomic-embed-text` first-load. `asyncio.gather(Semaphore(8))` is doing its job — CometBFT batches the concurrent txs. Didn't pull the `#199` baseline number for direct comparison but the shape is right.

### 3. Partial-failure recovery

SIGKILL at T+8s mid-seed → 56 memories in DB. Recovery re-run (no `--force`): `Stored: 31/87 Skipped: 56 Failed: 0`. Final DB count exactly 87, no orphans.

This test **also caught a third SAGE bug**: `SetTags` and `UpdateAgentLastSeen` ran on `r.Context()`, so any client that died between `broadcast_tx_commit` returning and the tag write left an untagged orphan row. First SIGKILL pass against 6.6.4-rc2 showed 8× `failed to set tags on committed memory: "begin tx: context canceled"` and a post-recovery count of 95 (expected 87). Switched to `context.WithTimeout(context.Background(), 10s)` so post-commit finalisation survives client disconnect. Final 6.6.4 run: zero orphans, exact count.

### 4. `--force`

Clean: `Stored: 87/87 Skipped: 0 Failed: 0` on the seed path, `16/16` on the agents path. Existing-memories are re-proposed as documented.

### 5. `/agentic` thread-safety

Didn't run a full `/agentic` session but ran `pytest core/sage/tests/test_sage_hooks.py` which covers the scenario: `17 passed`. `hooks.py:_get_client` guards init with `threading.Lock` + `_client_initialised` flag; 16-worker concurrent first-call test confirms exactly one `SageClient` construction and the cached-unavailable decision sticks (no probe-storm). LGTM as-is.

### 6. Registration telemetry

**Against `sage:6.6.0`:** first register_agent call returned `on_chain_height=None` (`raptor-seed`), only the second (already_registered path) returned `on_chain_height=84` (`raptor-registrar`). Fourth SAGE bug: `handleAgentRegister` used `broadcastTx` (sync, no block height), and only the idempotent `already_registered` branch surfaced the height. Switched first-register to `broadcastTxCommitWithHeight`.

**Against `sage:6.6.4`:** Run 1 `Registered as raptor-seed (on-chain height 13)` and `Registered as raptor-registrar (on-chain height 13)`. Run 2 same two lines. Both paths surface it.

### What landed in SAGE v6.6.4

- DSN pragma fix + belt-and-braces explicit PRAGMA statements (WAL, busy_timeout=15s, synchronous=NORMAL, foreign_keys=ON)
- `beginTxLocked` helper threaded through `SetTags`, `InsertTriples`, `RotateIdentity`, `ReassignMemories{,ByTag,ByDomain}`
- `SubmitMemory` post-commit finalisation on `context.WithTimeout(context.Background(), 10s)` for both `SetTags` and `UpdateAgentLastSeen`
- `handleAgentRegister` first-register returns `on_chain_height`
- 3 regression tests: `TestSQLitePragmasApplied`, `TestSetTags_ConcurrentWithInserts`, `TestSubmitMemory_TagCtxSurvivesClientDisconnect`

Raptor-side bump PR incoming — just `docker-compose.sage.yml` 6.6.0 → 6.6.4. No raptor code changes needed; the existing `async_memory_exists` pattern works correctly once SAGE actually persists the tags.

PR #193, opened 2026-04-22
add --understand and --validate as opt-in flags to /agentic

danielcuthbert:
## What

Two new opt-in flags for `/agentic`:

- `--understand` — runs `/understand --map` before the scan, building a full context map (entry points, trust boundaries, sinks) that feeds into the analysis. Findings get evaluated with architectural knowledge rather than in isolation.
- `--validate` — after the pipeline completes, runs the full 8-stage `/validate` pipeline on anything that came back exploitable or confirmed.

## Why

Right now `/agentic` analyses findings in isolation. If you're doing a proper security review on a workstation rather than a quick CI scan, you want the richer context that `/understand` and `/validate` add. The `/validate` bridge already picks up `context-map.json` automatically if it's in the project directory, so the two flags compose nicely.

Both flags are opt-in so nothing breaks for existing workflows. It also means this works better in a pipeline or as an agent you'd call in chat window within the IDE, which is pretty sweet imho 

## How it works

The flags are consumed at the Claude Code layer (`agentic.md`), not by the Python pipeline. `raptor.py` strips them before passing args to `raptor_agentic.py` as a safety net.

## Usage

```bash
# full enriched review
python3 raptor.py agentic --repo /path/to/code --understand --validate

# just pre-scan context mapping
python3 raptor.py agentic --repo /path/to/code --understand

# just post-scan validation on exploitable findings
python3 raptor.py agentic --repo /path/to/code --validate
```

## Files changed

- `.claude/commands/agentic.md` — documents the flags and the orchestration steps Claude follows for each
- `CLAUDE.md` — updated the agentic note to mention both flags
- `raptor.py` — strips `--understand` and `--validate` before they reach `raptor_agentic.py`

### DIFF
--- a/.claude/commands/agentic.md
+++ b/.claude/commands/agentic.md
@@ -9,15 +9,71 @@ description: Full autonomous security workflow — scan, dedup, prep, analyse, c
 2. Deduplicate findings
 3. Prep findings (read code, extract dataflow)
 4. **Validate + analyse** each finding (exploitation-validator methodology, Stages A-D)
-5. **Self-review** — catch contradictions, retry low confidence (Stage F)
-6. **Consensus** — multi-model second opinion (if configured)
+5. **Self-review**: catch contradictions, retry low confidence (Stage F)
+6. **Consensus**: multi-model second opinion (if configured)
 7. **Generate exploit PoCs** for exploitable findings
 8. **Generate secure patches** for confirmed vulnerabilities
 9. **Cross-finding analysis** (structural grouping, shared root causes)
 
-Nothing will be applied to your code - only generated in out/ directory.
+Nothing will be applied to your code - only generated in the out/ directory.
 
 Execute: `python3 raptor.py agentic --repo <path>`
+
+## Optional enrichment flags
+
+By default, `/agentic` scans and analyses findings in isolation. Two optional flags add richer context for more thorough results. They are opt-in because they add time and cost, but if you are doing a proper security review rather than a quick scan, they are well worth it.
+
+| Flag | What it does |
+|------|-------------|
+| `--understand` | Runs `/understand --map` before scanning to build a full context map: entry points, trust boundaries, sinks. This feeds directly into the analysis so findings are evaluated with architectural knowledge rather than in isolation. |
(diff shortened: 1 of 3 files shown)


.claude/commands/agentic.md
grokjc, line 1: The bit that says "findings where is_exploitable is true or confidence is high" is a bit vague - is "high" a string like "high", a number above some threshold, or something else? And if those field names get renamed later, Claude will just keep reading this markdown and quietly validate the wrong findings, with nothing flagging that it's gone wrong.                                                                  
                  
Could we point at the actual schema file and spell out exactly which field values count? That way it's unambiguous now, and if the schema ever changes someone will notice.

raptor.py
```diff
@@ -182,6 +182,11 @@ def mode_agentic(args: list) -> int:
         print(f"✗ Agentic workflow script not found: {agentic_script}")
         return 1
 
+    # --understand and --validate are consumed by the Claude Code agentic
+    # command and never reach this function. Strip them here as a safety net
+    # so raptor_agentic.py doesn't receive unknown flags.
+    args = [a for a in args if a not in ('--understand', '--validate')]
```
grokjc, line 189: The strip is exact-match, so --understand=true would slip through. Neither flag takes a value today so it's not a bug, but a one-line comment saying "these flags must never take values" would lock that in for whoever touches it next (and stop me breaking this later 😄)

grokjc (commented):
One thing worth tightening before merge: the Usage examples in the PR body and agentic.md all show _python3 raptor.py agentic --understand --validate_, but those flags only fire when invoked via the /agentic slash command - the raw python3 form strips them and runs plain agentic. 

Could we swap the examples to the slash-command form so people don't copy-paste the one that silently drops the functionality?

.claude/commands/agentic.md
danielcuthbert: 200% right on this, vaguess is meh. I will be updating this to do the schema option, good call

danielcuthbert: swapped to /agentic --understand --validate slash-command form with a note that the flags don't work via the Python CLI

yes i did remove the high vagueness with exact field names, types, and enum values pointing at packages/llm_analysis/prompts/schemas.py

raptor.py
danielcuthbert: added the one-line comment locking in the boolean-only invariant

Issue #159, opened 2026-04-10
Installing via docker fails with an error


thezuck:
**Running this:**
`docker build -f .devcontainer/Dockerfile -t raptor-devcontainer:latest .`

**fails with that:**
```
=> ERROR [ 2/13] RUN apt-get update && export DEBIAN_FRONTEND=noninteractive     && apt-get install -y --no-install-recommends     python3-pip     python3-venv     git     build-essential     gcc     g++     make     cmake     autoconf  4.7s
...
2.895 Get:7 https://dl.yarnpkg.com/debian stable InRelease
2.985 Err:7 https://dl.yarnpkg.com/debian stable InRelease
2.985   The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 62D54FD4003F6525
2.991 Reading package lists...
3.385 W: GPG error: https://dl.yarnpkg.com/debian stable InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 62D54FD4003F6525
3.385 E: The repository 'https://dl.yarnpkg.com/debian stable InRelease' is not signed.
```
**Reason as per Chatgpt:**

> Root cause (precise)
> The base image:
> 
> FROM mcr.microsoft.com/devcontainers/python:1-3.12-bookworm
> 
> already includes the Yarn APT repository:
> 
> https://dl.yarnpkg.com/debian
> 
> But it does NOT include the corresponding GPG key in a way compatible with modern APT (signed-by).
> 
> So when the first RUN apt-get update executes, APT sees:
> 
> NO_PUBKEY 62D54FD4003F6525
> 
> and aborts.
> 
> This is not the Dockerfile’s fault — it’s inherited state from the base image.

Suggested fix (confirmed):
Replace the following line in the Dockerfile:

`RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \`

with:
```
RUN rm -f /etc/apt/sources.list.d/yarn.list \
    && apt-get update \
    && export DEBIAN_FRONTEND=noninteractive \
```

This worked for me, don't know if yarn is needed or not but for now I did not run into any issues after removing it.

grokjc:
Thanks! See #71

grokjc:
Fixed in 72c3c2d. Thanks for reporting and proposing the rm fix @thezuck.

PR #71, opened 2026-03-07
Fix devcontainer Dockerfile - add Yarn GPG key and repository

martclau:
Building on Ubuntu 24.04.4 using Docker 29.3.0 hits a missing requirement. The change fixes this by adding the Yarn repo and PGP key.

$ sudo docker build --no-cache -f .devcontainer/Dockerfile -t raptor-devcontainer:latest .
[+] Building 2.4s (6/17)                                                                                                                                                                                                                 docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                                               0.0s
 => => transferring dockerfile: 5.86kB                                                                                                                                                                                                             0.0s
 => [internal] load metadata for mcr.microsoft.com/devcontainers/python:1-3.12-bookworm                                                                                                                                                            0.1s
 => [internal] load .dockerignore                                                                                                                                                                                                                  0.0s
 => => transferring context: 2B                                                                                                                                                                                                                    0.0s
 => CACHED [ 1/13] FROM mcr.microsoft.com/devcontainers/python:1-3.12-bookworm@sha256:7876580dc67fd460fd962f004cbeb480027e9bbc0657096f1087db11f9eaff39                                                                                             0.0s
 => => resolve mcr.microsoft.com/devcontainers/python:1-3.12-bookworm@sha256:7876580dc67fd460fd962f004cbeb480027e9bbc0657096f1087db11f9eaff39                                                                                                      0.0s
 => [internal] load build context                                                                                                                                                                                                                  0.0s
 => => transferring context: 187B                                                                                                                                                                                                                  0.0s
 => ERROR [ 2/13] RUN apt-get update && export DEBIAN_FRONTEND=noninteractive     && apt-get install -y --no-install-recommends     python3-pip     python3-venv     git     build-essential     gcc     g++     make     cmake     autoconf       2.0s
------                                                                                                                                                                                                                                                  
[...]

PR #194, opened 2026-04-22
fix dockerfile for bookworm: drop dead packages, fix libgcc

### LINKED ISSUE #183: Dockerfile has old dependencies
Debian Bookwork removed `libgconf-2-4` and `libgcc1` was renamed to `libgcc-s1`. Also the `yarn` apt repo has an expired key, so you have to fix that with `RUN rm -f /etc/apt/sources.list.d/yarn.list`. 

As such, the Dockerfile should probably look like the attached. Acknowledgement to the LLM that helped. #ThanksBeToClaude

Cheers! M.

[Dockerfile-raptor-updated.txt](https://github.com/user-attachments/files/26893190/Dockerfile-raptor-updated.txt)

danielcuthbert:
## What

Three fixes to get the devcontainer building cleanly on Debian Bookworm:

- Remove `libgcc1` (gone in Bookworm, replaced by `libgcc-s1`)
- Remove `libgconf-2-4` (GNOME 2 package, hasn't existed for years)
- Add `RUN rm -f /etc/apt/sources.list.d/yarn.list` before the main apt block to stop apt falling over on the conflicting repo

## Why

The current Dockerfile fails to build on Bookworm because both `libgcc1` and `libgconf-2-4` no longer exist in the package repositories. This has been blocking devcontainer usage.

Fixes #183. Thanks to @unprovable for the updated Dockerfile.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -11,6 +11,8 @@ FROM mcr.microsoft.com/devcontainers/python:1-3.12-bookworm
 # Organized by component that needs them
 # ============================================================================
 
+RUN rm -f /etc/apt/sources.list.d/yarn.list
+
 RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     && apt-get install -y --no-install-recommends \
     #
@@ -60,8 +62,7 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
     libdbus-1-3 \
     libexpat1 \
     libfontconfig1 \
-    libgcc1 \
-    libgconf-2-4 \
+    libgcc-s1 \
     libgdk-pixbuf2.0-0 \
     libglib2.0-0 \
     libgtk-3-0 \


grokjc:
LGTM — all three fixes are correct for Bookworm:                                                                                                                                                                 
                                                                                                                                                                                                                   
- `libgconf-2-4` was removed from Debian 12 (GNOME 2 lib; gone since Bullseye deprecation). Current builds must be failing on this line unless they're hitting a cached apt layer.                               
- `libgcc1` → `libgcc-s1` is the correct rename in Bookworm.
- `rm -f /etc/apt/sources.list.d/yarn.list` matches the root cause thezuck captured in #159 (`NO_PUBKEY 62D54FD4003F6525` from a stale yarn repo baked into the `devcontainers/python:1-3.12-bookworm` base).    
                                                                                                                                                                                                                   
Worth adding `Fixes #159` to the PR body so it auto-closes alongside #183 on merge.                                                                                                                              
                                                                                                                                                                                                                   
This also supersedes #71 — that PR re-imports the yarn GPG key but never installs yarn, and with the stale `yarn.list` removed the key is no longer needed at all. #71 can be closed as superseded once this lands.          
                                                                                                                                                                                                                   
One follow-up (**not** this PR — keep the scope tight): `packages/codeql/build_detector.py:104,127,360` expects `yarn` on PATH for CodeQL JS targets. After this, `yarn` and `pnpm` should be added via the existing Node step (`npm install -g yarn pnpm`). Happy to send that as a stacked diff once #194 lands.

grokjc:
72c3c2df371cfa392ea9089794ee88d64cfed2de partially resolves (yarn issue etc I described above)

PR #196, opened 2026-04-22
Ship semgrep rules locally for offline/air-gapped pipeline support

danielcuthbert:
## What's this?

Turns out if you run RAPTOR in a pipeline with no outbound connectivity, semgrep just sits there quietly dying trying to reach semgrep.dev. Not ideal when you've just told someone \"yes it works in your air-gapped environment\". This fixes that.

## What changed

**Offline-first rule loading**
- Cached all 7 registry packs we actually use into `engine/semgrep/rules/registry-cache/` — they ship with the repo now
- `get_semgrep_config()` in `RaptorConfig` maps any pack ID to a local file first, falls back to the registry only if there's no cache hit
- Every pack we reference has a cache file. Zero network calls needed to run a full scan.

**Removed dead pack references**
Five entries in `POLICY_GROUP_TO_SEMGREP_PACK` were pointing at packs that simply do not exist on semgrep.dev (confirmed: all 404). They've been removed:
- `ssrf` (p/ssrf: 404, also had no local rules directory — was silently doing absolutely nothing)
- `crypto` (p/crypto and category/crypto: both 404 — local rules still run fine)
- `deserialisation`, `filesystem`, `logging` (same story)

**Fixed `--policy-groups all`**
The `all` expansion was picking up `registry-cache/` as a rule directory and attempting to scan with it. It is now excluded. Also added a proper warning when someone specifies an unknown policy group rather than silently producing zero findings.

**27 new crypto rules from OWASP ASVS V11**
While we were in the neighbourhood, the crypto rule set had some gaps against ASVS V11. Four new files:

| File | V11 Req | What it catches |
|------|---------|----------------|
| `weak-prng.yaml` | V11.5.1 | `Math.random()`, `new Random()`, `math/rand`, `rand()` across JS/Python/Java/Go/C/Ruby |
| `pkcs1v15-padding.yaml` | V11.3.1 | Bleichenbacher-vulnerable RSA padding across Java/Python/Go/Node |
| `insecure-password-hash.yaml` | V11.4.2 | Passwords hashed with MD5/SHA-256 etc. instead of bcrypt/argon2/scrypt |
| `weak-hash-extended.yaml` | V11.4.1 | MD5/SHA-1 coverage extended to JS, Go, Ruby, PHP, Rust, C |

## The irony

GitHub push protection blocked this PR because the cached `p/secrets` rules contain Slack webhook URL regex patterns that look like real secrets. A secrets detection rule file getting blocked by secrets detection is, frankly, peak security tooling.

## Test plan

- [ ] Run a scan with no network access and verify it completes without hanging
- [ ] Confirm `--policy-groups all` does not include `registry-cache` in the scan
- [ ] Confirm specifying `--policy-groups ssrf` logs a warning rather than silently producing nothing
- [ ] Validate new crypto rules with `semgrep --validate` on each file


core/config.py
```diff
@@ -158,6 +158,18 @@ class RaptorConfig:
...
+    def get_semgrep_config(cls, pack_id: str) -> str:
+        """Return local cached path for a registry pack if available, else the registry ID.
+
+        Naming: p/secrets -> c.p.secrets.json (mirrors semgrep.dev /c/p/ URL path).
+        Falls back to the registry identifier so online scans still work.
+        """
+        cache_file = cls.SEMGREP_REGISTRY_CACHE_DIR / ("c." + pack_id.replace("/", ".") + ".json")
+        if cache_file.exists():
+            return str(cache_file)
+        return pack_id
```
grokjc, line 171: Returning the registry ID on miss means a future-added pack without a cache file silently degrades to a network fetch. On the air-gapped hosts this PR is written for, semgrep will hang until SEMGREP_TIMEOUT, and run_single_semgrep will then write {"runs": []} on line 146 (empty output) or 165 (exception) — the exact silent-failure mode the PR is trying to eliminate. Suggest logger.warning on miss, or a strict mode so CI fails loud.

grokjc (commented):
There's no way to refresh the pack, do we need a mechanism to do so or some kind of manifest? How would we know if the cache was outdated/invalid?

(and do we need tests to that effect to ensure the cache matches reality?)

core/config.py
danielcuthbert: That is the next thing im working on, a proper refresh approach. This solves a bug right now where it fails when it cannot connect to semgrep, which isn't ideal at all as then a scan happens and returns 0 results.

grokjc:
Suggest we merge this and then follow up with the additional stuff when you're ready.

PR #197, opened 2026-04-22
Tighten SMT one-gadget checker, seed shared harness primitives

grokjc:
Fixes two correctness gaps: check_onegadget returned feasible=True when every constraint was unparseable (no actual verification) and named all pending constraints as "unsatisfied" on joint-unsat when only the unsat-core subset actually conflicted. Adds scoped() for push/pop hypothesis probing and explain.track/core_names for naming which assertions Z3 used to derive unsat — both adopted by the refactored check_onegadget and exposed for future encoders. Also reg-gates non-x86_64 tokens, logs unparseable constraints, surfaces SMT status on analyze_binary() output and the /validate report header, and clarifies the z3-unknown reasoning to name the solver timeout. Test coverage roughly doubled (30 → 57).


unprovable:
Side note - this is fire 🔥🔥🔥

PR #203, opened 2026-04-23
Bugfix in SMT - Multiple calls on `track()` might corrupt unsat-core names

unprovable:
Calling `track()` twice on the same solver with the same `rev` dict must not produce colliding labels.  The bug was that both calls would generate `_c0`, `_c1`, ... — corrupting the `rev` map so `core_names` returned wrong (or missing) constraint names, especially for large projects. This would make the output unreliable. 

What's added:
* A Fix in `core/smt_solver/explain.py`
* Tests in `core/smt_solver/tests/test_smt_solver.py`

**NB** - this bug isn't actually triggered in the existing `packages/exploit_feasibility/smt_onegadget.py` usage of `track()` but came up in my testing for other stuff and it seems prudent to lay the groundwork for some flexibility here ahead of time. 😊

### DIFF
--- a/core/smt_solver/explain.py
+++ b/core/smt_solver/explain.py
@@ -15,12 +15,16 @@ Usage::
 """
 from __future__ import annotations
 
-from typing import Any, Dict, List, Sequence, Tuple
+from typing import Any, Dict, List, Optional, Sequence, Tuple
 
 from .availability import z3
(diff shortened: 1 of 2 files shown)


danielcuthbert (approved):
LGTM

PR #206, opened 2026-04-23
Enhancement of CodeQL Analysis with SMT Solver

unprovable:
## Using SMT Solvers where LLMs Struggle

**IDEA** - incorporate the SMT Solver logic from #188 into the CodeQL analysis workflow. This is potentially very useful as bug classes (listed below) that are based in integer bounds/arithmetic/etc. are hard to LLMs to get right, whilst they seem to excel at string-based checking. So, this PR adds the existing SMT Solver core to CodeQL, providing parsers, tests, and analysis shims.

Paging @grokjc for review as this is another biggie (sorry!). 🚀 

On first analysis with Claude, the following bug classes stood out as points where SMT might really help: 

  1. CWE-190 (integer overflow) — pure bitvector arithmetic, Z3 is ideal here!
  2. CWE-122/120 (buffer overflow/size checks) — `size < buf_len` constraints, Z3 should help
  3. CWE-193 (off-by-one) — boundary condition arithmetic
  4. CWE-476 (null deref) — `ptr != NULL` equality

**Note** - bugs like CWE-89 (SQLi) are still best left to something that can parse language strings and  has some inherent abstract notion of string/meaning mutability… you know, like an LLM :-P But LLM’s notoriously fall over when it comes to arithmetic, linear constraints, etc. So this is where an SMT Solver should actually add a lot of value. 

To that end, here’s what’s in the PR box - new files first:

* `packages/codeql/smt_path_validator.py` - this imports the SMT solver harness and adds parsers for codeql output that allows it to then run the solver on the constraints identified through codeql queries. 
* `packages/codeql/tests/test_smt_path_validator.py` - tests for the above, like any responsible citizen. 

Then the modified files:
* `packages/codeql/dataflow_validator.py` - shimming the analysis and parser chain from `smt_path_validator.py` into the dataflow validation pipeline. Before the LLM makes a determination, the code runs the parsers and SMT solver code, and passes the output of the run (any unsat and sat results) to the LLM to enrich exploitability decisions. 
* `.claude/commands/codeql.md`, `.claude/skills/exploitability-validation/stage-e-feasibility.md`, and `CLAUDE.md` - essentially updating the agent with what new toys it has to play with. 

## New SMT-CodeQL Testbench

There is also a new test bench in `test/data/smt_codeql_testbench` that presents the following test case groups for CodeQL analysis:

### Group 1 — SAT with non-obvious PoC values

Z3 finds concrete satisfying assignments that require reasoning about unsigned bitvector wraparound or compound inequality constraints.

| Case | File | Sink line | Guard (visible) | Bug | Z3 PoC value |
|------|------|-----------|-----------------|-----|--------------|
| `ALLOC` | `smt_codeql_testbench.c:93` | `memset(&records[i], ...)` | `count < 0x40000000` | `count * 16` overflows uint32 | `count = 0x10000001` → alloc = 16 bytes, loop writes 4 GB |
| `SUM` | `smt_codeql_testbench.c:132` | `memcpy(shared_buffer + offset, ...)` | `offset + length <= buffer_size` | unsigned sum wraps | `offset=0xFFFF0000, length=0x10010` → sum=0x10 ≤ 64 |
| `OBO` | `smt_codeql_testbench.c:165` | `buf[index] = value` | `index > INDEX_LIMIT` (should be `>=`) | off-by-one | `index=128` passes `> 128` check, writes `buf[128]` |
| `MASK` | `smt_codeql_testbench.c:200` | `memcpy(heap_region + base, ...)` | `base + size <= HEAP_SIZE` | unsigned base+size wrap | `base=0xFFFFE000, size=4000` → sum=0x3E80 ≤ 8192 |

### Group 2 — UNSAT (provably dead paths)

CodeQL reports a flow to the sink; Z3 proves the path conditions are mutually exclusive and suppresses the LLM call.

| Case | File | Sink line | Conditions Z3 checks | Conflict |
|------|------|-----------|----------------------|----------|
| `DEAD_RANGE` | `smt_codeql_testbench.c:243` | `strcpy(dead_buf, data)` | `x > 100` AND `x < 50` | ⊥ immediately |
| `DEAD_NULL` | `smt_codeql_testbench.c:280` | `strcpy(dst, ptr)` | `ptr != NULL` AND `ptr == NULL` | ⊥ immediately |
| `DEAD_MASK` | `smt_codeql_testbench.c:308` | `memcpy(flag_buf, payload, len)` | `flags & 0x1 == 0` AND `flags & 0x1 == 1` | ⊥ immediately |

### Group 3 — Indeterminate (Z3 falls through to LLM)

Conditions contain function calls the bitvector parser cannot encode. Z3 returns `feasible=None` and full LLM analysis runs unchanged.

| Case | File | Why Z3 returns `None` |
|------|------|-----------------------|
| `LLM` | `smt_codeql_testbench.c:327` | `strlen(input)` call — parentheses rejected by bitvector parser |
| `PARTIAL` | `smt_codeql_testbench.c:350` | `validate(ptr)` call — same reason; `size < 256` is parseable but alone is indeterminate |

## What’s missing?

In the parser, `~` (unary NOT) is left as unknown. It requires a unary parser path (no RHS operand), and ~mask in real conditions is almost always wrapped in a bitmask form like `addr & ~PAGE_MASK == 0`. I have essentially chickened out of this level of complexity for now. 🐣

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,3 @@ out/
 
 # Runtime artifacts
 codeql_dbs/
-
-# Testbench data to ignore
-test/data/testbench
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ out/
 
 # Runtime artifacts
 codeql_dbs/
+
+test/data/
--- a/.gitignore
+++ b/.gitignore
@@ -22,4 +22,5 @@ out/
 # Runtime artifacts
 codeql_dbs/
 
-test/data/
+# Testbench data to ignore
+test/data/testbench
--- a/README.md
+++ b/README.md
@@ -92,7 +92,9 @@ demonstrates how Claude Code can be adapted for **any purpose**, with RAPTOR pac
 - **Cost Management:** Budget enforcement, real-time callbacks, and intelligent quota detection
 - **Enhanced Reliability:** Multiple bug fixes improving robustness across CodeQL, static analysis, and LLM providers
 - **Code Understanding** We wanted to build more adversarial code comprehension, which allows you to map attack surface, trace those vital data flows &amp; hunt for vulnerability variants
-- **Z3 SMT Engine:** Two-layer SMT integration, both optional (`pip install z3-solver`) and gracefully absent-safe. (1) **Dataflow pre-screening:** CodeQL path conditions are checked for joint satisfiability before any LLM call — provably unreachable paths are discarded, satisfiable paths get concrete candidate inputs injected into the prompt. (2) **One-gadget constraint analysis:** during binary feasibility assessment, Z3 checks whether a one-gadget's register/memory constraints are satisfiable given the concrete crash state, ranking gadgets by actual reachability rather than heuristics alone.
+- **Z3 SMT Engine:** Two-layer SMT integration, both optional (`pip install z3-solver`) and gracefully absent-safe.
+  - **Dataflow pre-screening:** CodeQL path conditions are checked for joint satisfiability before any LLM call — provably unreachable paths are discarded, satisfiable paths get concrete candidate inputs injected into the prompt.
+  - **One-gadget constraint analysis:** during binary feasibility assessment, Z3 checks whether a one-gadget's register/memory constraints are satisfiable given the concrete crash state, ranking gadgets by actual reachability rather than heuristics alone.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ demonstrates how Claude Code can be adapted for **any purpose**, with RAPTOR pac
 - **Cost Management:** Budget enforcement, real-time callbacks, and intelligent quota detection
 - **Enhanced Reliability:** Multiple bug fixes improving robustness across CodeQL, static analysis, and LLM providers
 - **Code Understanding** We wanted to build more adversarial code comprehension, which allows you to map attack surface, trace those vital data flows &amp; hunt for vulnerability variants
+- **Z3 SMT Engine:** Two-layer SMT integration, both optional (`pip install z3-solver`) and gracefully absent-safe. (1) **Dataflow pre-screening:** CodeQL path conditions are checked for joint satisfiability before any LLM call — provably unreachable paths are discarded, satisfiable paths get concrete candidate inputs injected into the prompt. (2) **One-gadget constraint analysis:** during binary feasibility assessment, Z3 checks whether a one-gadget's register/memory constraints are satisfiable given the concrete crash state, ranking gadgets by actual reachability rather than heuristics alone.
 
 ---
 
@@ -308,7 +309,7 @@ docker build -f .devcontainer/Dockerfile -t raptor-devcontainer:latest .
 /fuzz     - Binary fuzzing (AFL++ + crash analysis)
 /web      - Web application security testing (STUB - treat as alpha)
 /agentic  - Full autonomous workflow (analysis + exploit/patch generation)
-/codeql   - CodeQL-only deep analysis with dataflow
+/codeql   - CodeQL-only deep analysis with dataflow + Z3 SMT path pre-screening
 /analyze  - LLM analysis only (no exploit/patch generation - 50% faster & cheaper)
 /validate - Exploitability validation pipeline
 ```
--- a/packages/codeql/dataflow_validator.py
+++ b/packages/codeql/dataflow_validator.py
@@ -12,6 +12,12 @@ from dataclasses import dataclass, asdict
 from pathlib import Path
 from typing import Dict, List, Optional
 
+from packages.codeql.smt_path_validator import (
+    PathCondition,
+    PathSMTResult,
+    check_path_feasibility,
+)
+
 # Add parent directory to path for imports
 # packages/codeql/dataflow_validator.py -> repo root
 sys.path.insert(0, str(Path(__file__).parents[2]))
@@ -55,6 +61,11 @@ class DataflowValidation:
     barriers: List[str]
     prerequisites: List[str]
 
+
+PATH_CONDITIONS_SCHEMA = {
+    "path_conditions": "list of {step_index: int, condition: str, negated: bool}",
+    "unparseable": "list of strings — conditions too complex to express as simple predicates",
+}
 
 # Dict schema for LLM structured generation (consistent with other callers)
 DATAFLOW_VALIDATION_SCHEMA = {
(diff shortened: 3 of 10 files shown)


.gitignore
```diff
@@ -20,4 +20,7 @@ out/
 .claude/raptor.env
 
 # Runtime artifacts
-codeql_dbs/
\ No newline at end of file
+codeql_dbs/
+
+# Testbench data to ignore
+test/data/testbench
```
grokjc: Ignoring a file that doesn't exist! (do we need this?)

packages/codeql/smt_path_validator.py
```diff
@@ -0,0 +1,357 @@
...
+# Tokenise: identifiers, hex literals, decimal literals, operators.
+# '>>' and '<<' appear before '[<>&|]' so they are matched as two-char tokens
+# rather than as two separate single-char tokens.
+_TOKEN_RE = re.compile(
+    r'(0x[0-9a-f]+|\d+|[a-z_][a-z0-9_]*|[+\-*]|<=|>=|!=|==|>>|<<|[<>&|])',
+    re.IGNORECASE,
+)
+
+
+def _parse_expr(text: str, vars_: Dict[str, Any]) -> Optional[Any]:
```
grokjc, line 112: Left-to-right arithmetic silently mis-encodes mixed-operator expressions
  
  No precedence + parens rejected means `a + b * 2 == 10` parses as `(a+b)*2 == 10`.
  With a=2, b=3 the real semantics give 8≠10 (infeasible), parser gives 10 and Z3
  reports feasible=True with a bogus witness. Silent wrong verdict.
  
  Fix: reject expressions containing both `+`/`-` and `*`/`>>`/`<<` — send to
  `unknown`. One-line guard. Real precedence or paren support in a follow-up.

packages/codeql/smt_path_validator.py
```diff
@@ -0,0 +1,357 @@
...
+  count * 16 < max_alloc       (bitvector mul — wraparound at 2^64, not 2^32)
+  n >> 1 < limit               (logical right shift)
+  n << 3 == buf_size           (left shift)
+  flags | 0x1 != 0             (bitwise OR)
+  ptr != NULL  /  ptr == NULL
+  index >= 0
+  flags & 0x80000000 == 0
+  value == 42
+
+Variables are created as 64-bit bitvectors.  Signed comparisons are used
```
grokjc: Docstring says signed, code is unsigned-only
                                                            
  Docstring: "Signed comparisons are used by default". Code: `_SIGNED = False`,                                                                                                                                    
  every relational op is `z3.ULT/ULE/UGT/UGE`. No way for a caller to flip it.                                                                                                                                     
                                                                                                                                                                                                                   
  Update the docstring to match reality: unsigned 64-bit semantics throughout,                                                                                                                                     
  signed-integer predicates around the sign bit go to `unknown`.

packages/codeql/tests/test_smt_path_validator.py
```diff
@@ -0,0 +1,331 @@
...
+    @_requires_z3
+    def test_multiplication_makes_path_infeasible(self):
+        """count * 4 == 8 AND count == 3 is unsatisfiable (3*4 = 12, not 8)."""
+        r = check_path_feasibility([
+            PathCondition("count * 4 == 8", step_index=0),
+            PathCondition("count == 3", step_index=1),
+        ])
+        assert r.feasible is False
+
+    def test_multiplication_with_unsupported_op_goes_to_unknown(self):
```
grokjc: Test doesn't test what its name claims                                                                                                                                                                           
                                                            
  `a * b | c` parses fine — both ops are supported. The test only passes because                                                                                                                                   
  `_z3_available` is mocked to False. Delete it (the `TestNoZ3` class covers the                                                                                                                                   
  no-z3 path) or rewrite against a genuinely unsupported form.

packages/codeql/dataflow_validator.py
```diff
@@ -205,6 +278,35 @@ def validate_dataflow_path(
...
+        conditions = self._extract_path_conditions(dataflow, repo_path)
+        smt_result = check_path_feasibility(conditions)
+
+        if smt_result.feasible is False:
+            self.logger.info(
+                f"SMT: path infeasible — {smt_result.reasoning}"
+            )
+            return DataflowValidation(
+                is_exploitable=False,
+                confidence=0.9,
```
grokjc: confidence=0.9 too high for an LLM→SMT chain
  
  Verdict depends on LLM extracting all relevant conditions faithfully. Missed or
  wrongly-negated conditions produce false infeasibility — dangerous direction.
  Drop to 0.7 and note the contingency in reasoning.

test/data/smt_codeql_testbench/README.md
```diff
@@ -0,0 +1,33 @@
+# SMT Testbench
+
+Cases designed to exercise Z3 path condition analysis in CodeQL dataflow findings.
+
+## Group 1 — SAT with non-obvious PoC values
```
grokjc: Group 1 SAT claims don't match 64-bit encoder semantics
  
  ALLOC, SUM, MASK all rely on 32-bit unsigned wraparound for their stated PoC
  values. The encoder is fixed at 64-bit BV — no wrap at those values. Z3 will
  find SAFE witnesses instead (ALLOC: count=1; SUM: offset=1,length=1), so the
  LLM prompt gets innocuous inputs, not the overflow case.
  
  OBO works fine. Groups 2 and 3 unaffected.
  
  Fix here: mark Group 1 as aspirational — note that full wraparound detection
  needs the width-parametric encoder (follow-up PR). Properly handled when
  signedness/width go parametric.

README.md
```diff
@@ -92,6 +92,7 @@ demonstrates how Claude Code can be adapted for **any purpose**, with RAPTOR pac
 - **Cost Management:** Budget enforcement, real-time callbacks, and intelligent quota detection
 - **Enhanced Reliability:** Multiple bug fixes improving robustness across CodeQL, static analysis, and LLM providers
 - **Code Understanding** We wanted to build more adversarial code comprehension, which allows you to map attack surface, trace those vital data flows &amp; hunt for vulnerability variants
+- **Z3 SMT Engine:** Two-layer SMT integration, both optional (`pip install z3-solver`) and gracefully absent-safe. (1) **Dataflow pre-screening:** CodeQL path conditions are checked for joint satisfiability before any LLM call — provably unreachable paths are discarded, satisfiable paths get concrete candidate inputs injected into the prompt. (2) **One-gadget constraint analysis:** during binary feasibility assessment, Z3 checks whether a one-gadget's register/memory constraints are satisfiable given the concrete crash state, ranking gadgets by actual reachability rather than heuristics alone.
```
grokjc: Style mismatch with surrounding list
  
  Single long paragraph in a list of 1–2-line bullets. Split into a parent bullet
  with two sub-bullets (CodeQL dataflow pre-screen, one-gadget analysis) to match.

grokjc (commented):
Thanks - this is definitely in 'I need Claude to review this' territory ... so that's what happened 😄 

There's a design conversation I want to have with you after this lands around parametric width/signedness and a shared csem module. Not a blocker for this PR. Will pick that up separately after this merges but I have some followups in mind for the both of us. This is growing arms and legs, in a good way!

.gitignore
unprovable: Sorry no - this is an artefact from my testing. I'll get rid of this line now (and try to remember not to trust `git status` so blindly)... Fixed in [this commit](https://github.com/gadievron/raptor/pull/206/commits/8f9bbb58abc13e00f0b8158fa2ec91fc49bfdf5b)

packages/codeql/smt_path_validator.py
unprovable: Good catch! Fixed in [this commit](https://github.com/gadievron/raptor/pull/206/commits/9a6d91cf0c653532ffa943d92a2fa63d844d501b)

packages/codeql/dataflow_validator.py
unprovable: Yeah, this is totally right. Fixed in [this commit](https://github.com/gadievron/raptor/pull/206/commits/54ef76cd92a5961de0c037f937b4bfdb904dce73)

unprovable: LMK if you want to chagne the reasoning wording there, btw.

README.md
unprovable: A good shout - fixed in [this commit](https://github.com/gadievron/raptor/pull/206/commits/aabd63fb5876b553433b5dbee2247ea2106db6c8)

packages/codeql/tests/test_smt_path_validator.py
unprovable: Not sure what was intended there, really. I think I got some confused hallucination of a human kind there. REmoved in [this commit](https://github.com/gadievron/raptor/pull/206/commits/28473bce45e4a557ed9ee158320a03a936e3c588)

test/data/smt_codeql_testbench/README.md
unprovable: totally get what is meant here. Updated the descriptions as much as I can, and asked a few bots to double check. When we make the discussion for more broadly getting signedness and bitwidth to be parametric, then we can consider how to fix this. Resolved in [this commit](https://github.com/gadievron/raptor/pull/206/commits/1c72c615fef1d3366c127078193626a45362ff61)

grokjc:
Thanks - can you also update the README for the testbench to match the ASPIRATIONAL markers that you put in the .c? Other than that, all good!

packages/codeql/smt_path_validator.py
unprovable: Fxied in [this commit](https://github.com/gadievron/raptor/pull/206/commits/ffbbb7e2ab6ff7ad9e08a62e496ca01e5e63c18f)

unprovable:
I hope this is alright? 

So, this is actually where I wanted to get to - building a framework for automating the parsing and analysis of CodeQL results. Deferring it to an agent is a bonus, but getting these parsers and analytical logic started is the main aim. We can coordinate more on the overall structure moving forward as you suggest, of course! 

Many thanks for your help! M.

grokjc:
Merged — thanks for the fast turnaround on the review items, especially the precedence guard which ended up better than I'd originally suggested.

Hold off on further SMT work for a moment. I've got a consolidation PR in the works that will:

  - Add `core/smt_solver/csem.py` with width/signedness-aware primitives (wrap/extend/truncate/shifts)
  - Make `check_path_feasibility()` and `check_onegadget()` accept `width` / `signed` parameters, sourced from the finding's type context rather than hardcoded
  - Unblock the Group 1 cases here and signed-overflow predicates generally

I'll ping you when it lands ...

Separately I want to have a conversation about the enhancement backlog (Stage-E integration, /understand flow-trace as a second condition source, per-CWE encoders) so we can carve up that work between us.

unprovable:
This sounds great! I'll look forward to it. And yes, fire over lists of the enhancement backlog. Cheers! M.

PR #208, opened 2026-04-23
SMT: parametric BVProfile across encoders

grokjc:
Introduces a frozen BVProfile dataclass and named constants (BV_X86_64, BV_C_UINT32, BV_C_INT32, …) as the single parametric handle both SMT encoders now accept, replacing the ad-hoc width/signed kwargs and the unused _Z3_BVConfig default.

Adds core/smt_solver/csem.py with C-semantics primitives (signed and unsigned overflow predicates, width coercion, arithmetic vs logical right shift, C-style cast) that future encoders need to reason about real C arithmetic rather than abstract bitvector math.

Wires the CodeQL dataflow validator so the extraction LLM emits per-path width/signed hints and a rule-id heuristic (CWE-190 family → 32-bit unsigned) fills in when hints are absent. The chosen profile flows through to check_path_feasibility, so CWE-190 32-bit wraparound — previously aspirational in the testbench — now works end-to-end.

Reasoning strings spell out the modelled semantics ("32-bit unsigned") instead of bv{32,64}{s,u} shorthand, so Stage E can spot profile / vuln-type mismatches without decoding jargon.

45 new tests; full repo suite passes at 2014.

### DIFF
--- a/.claude/commands/codeql.md
+++ b/.claude/commands/codeql.md
@@ -40,9 +40,10 @@ pre-check before the full LLM analysis (`packages/codeql/smt_path_validator.py`)
 
 Requires `z3-solver` (`pip install z3-solver`). Degrades gracefully when absent.
 
-**Best coverage:** CWE-190 (integer overflow), CWE-120/122 (buffer size checks),
-CWE-193 (off-by-one), CWE-476 (null deref). String-based findings (CWE-89) fall
-through to LLM analysis.
+**Best coverage:** CWE-190 (integer overflow, **including 32-bit wraparound** —
+the extraction LLM emits per-path width/signedness hints so Z3 models the right
+C type semantics), CWE-120/122 (buffer size checks), CWE-193 (off-by-one),
+CWE-476 (null deref). String-based findings (CWE-89) fall through to LLM analysis.
 
 ## Examples
 
--- a/.claude/skills/exploitability-validation/stage-e-feasibility.md
+++ b/.claude/skills/exploitability-validation/stage-e-feasibility.md
@@ -127,8 +127,17 @@ structured predicates; Z3 checks joint satisfiability:
 - `feasible: null` — Z3 unavailable or all conditions unparseable; full LLM
   analysis runs unchanged.
 
-Best coverage: CWE-190 (integer overflow — Z3 will find wraparound bypass paths),
-CWE-120/122 (buffer size arithmetic), CWE-193 (off-by-one), CWE-476 (null deref).
+The `reasoning` string carries a mode tag in parentheses — e.g.
+`"feasible (32-bit unsigned): ..."`. This tells you the bitvector semantics
+used for the verdict. **If the tag's width doesn't match the vulnerability's
+real C type**, the verdict is suspect: a CWE-190 (integer overflow) path
+analysed under `64-bit unsigned` misses 32-bit wraparound and the `false`
+verdict should be treated as incomplete — defer to your own C-semantics
+reasoning rather than trusting it as a chain-break.
+
+Best coverage: CWE-190 (integer overflow — Z3 finds wraparound bypass paths
+when the profile matches the real C type width), CWE-120/122 (buffer size
+arithmetic), CWE-193 (off-by-one), CWE-476 (null deref).
 
 **SMT feasibility (one-gadget):** When the optional `z3-solver` package is installed, `analysis.one_gadget_info.smt_feasibility` contains a Z3-backed verdict on whether the best gadget's register/memory constraints can be satisfied:
 
--- a/core/smt_solver/__init__.py
+++ b/core/smt_solver/__init__.py
@@ -1,40 +1,80 @@
 """SMT solver framework for RAPTOR.
 
 A thin, optional Z3 harness shared by domain encoders in ``packages/``.
-Handles availability gating, bitvector configuration, signed/unsigned
+Handles availability gating, bitvector primitives, signed/unsigned
 comparison routing, witness extraction, and solver construction with a
 default timeout.
 
-Domain-specific encodings (sanitizer patterns, integer overflow predicates,
-one-gadget constraints, ...) live in their respective ``packages/`` modules
-and import primitives from here.
+Encoders parametrise width and signedness via ``BVProfile`` — one
+handle instead of two flags, with pre-made profiles for common
+architecture and C-type flavours (``BV_X86_64``, ``BV_C_UINT32``,
+``BV_C_INT32``, ...).
+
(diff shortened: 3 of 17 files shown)


grokjc:
@unprovable - see PR content as promised.

## What landed
  
  - **BVProfile abstraction** — replaces the ad-hoc width/signed kwargs with a single frozen dataclass. Pre-made profiles for common architectures (`BV_X86_64`, `BV_AARCH64`) and C types (`BV_C_UINT32`, `BV_C_INT32`, …) read as intent at call sites: `profile=BV_C_UINT32` vs `width=32, signed=False`.                                                                                                                                                                                   
  - **csem.py** — C-semantics primitives landed properly now that they had a real consumer: signed/unsigned overflow predicates, width coercion, arithmetic-vs-logical shift disambiguators, C-style cast.                                                                                                                                            
  - **CodeQL wiring** — the extraction LLM now emits `path_width` / `path_signed` hints; a rule-id heuristic (CWE-190 family → 32-bit unsigned) fills in when the hint is missing. Profile threads through (previously ASPIRATIONAL) work end-to-end.                                                                                                                                                                     
  - **Reasoning strings** spell out "32-bit unsigned" rather than`bv32u`, so Stage E can spot profile/vuln-type mismatches.                                                                                                                                                     
                                                                                                                                                                                                                   
  ## What's open — in rough value-per-effort order             

This is what I'm thinking would be good additions - what do you reckon?                                                                                                                                                    
                                                                                                                                                                                                                   
  1. **Wire SMT into Stage E as an LLM-callable tool**. Stage E already reads source and reasons about memory-corruption feasibility heuristically; give it `check_path_feasibility()` / `check_overflow()` as functions the Stage E LLM can invoke ad-hoc, same pattern as `analyze_binary()`. Doesn't need pipeline plumbing — the LLM extracts operands itself. Biggest Stage E value unlock per effort.                                                                                                                                                                      

  2. **/understand flow-trace bridge** — flow-trace JSON already carries branch conditions + taint metadata. Feed it into `check_path_feasibility` as a second condition source (CodeQL paths are the first). Needs a small schema pact with /understand but otherwise reuses everything we've built.

  3. **Standalone integer-overflow encoder** (`smt_intoverflow`) as a first-class function — could reuse the Shape-A pattern from (1) so Stage E calls `check_overflow(operands, op, check="unsigned_wrap")` directly. Composes cleanly on csem.
  4. **One-gadget witness → concrete PoC bytes** — when `OneGadgetSMTResult.model` is populated, emit the register-setup stub as actual payload bytes. Closes the "verdict → PoC" loop.                                                                                                                                                
                                                                                                                                                                                                                   
  ## Division of labour — open for discussion
  
  - 1 (Stage E wiring) and 2 (/understand bridge) touch RAPTOR-internal glue I know well — happy to take either.
  - 3 (int-overflow encoder) and 4 (one-gadget witness bytes) are natural extensions of the SMT work you've been driving — yours if you want them, but no obligation.
 
 No urgency on any of this. Ping me when you have a view on which you want, and I'll pick up the rest.

unprovable:
This is amazing! Let me take a deeper look at see what magic we can weave. Inclined to go with the wiring of SMT into Stage E as something the LLM can call to effect analysis, and then the `/understand` flow-trace stuff is also a great shout. Then I think doing the witness -> poc bytes usage is actually more useful from a validation/exploit creation point of view. How best are we to do this?

grokjc:
On reflection a lot of that is simple(ish) wiring of the new SMT capability into the existing codebase. I'll take a first stab at these and let you know when they are on a local branch so you can see what you think.

Having spent some Claude tokens on the bigger picture, have identified things that definitely need your expertise that are probably a better use of your time!

  1. Optimisation queries — Z3 Optimize wrapper. Right now check_path_feasibility returns some satisfying witness. For exploit work we usually want the best witness — smallest input that triggers, largest count that still bypasses the guard, lowest-byte payload that satisfies bad-byte constraints. Z3's Optimize solver does this natively but the API is different (push hard/soft constraints, optimise expression). Wrap it as check_path_feasibility_optimal(conditions, minimise=expr) / maximise=.... Direct PoC-quality win.
  2. Multiple-witness enumeration. Find N distinct satisfying assignments ranked by usefulness (e.g., smallest payload, then distinct register patterns). Useful when the first witness Z3 picks isn't the one we
  want. Build on top of the existing solver via solver.add(z3.Not(model_so_far)) iteration with care for variable identity. Good companion to (1).
  3. Quantifier helpers in csem. Patterns like "for all bytes in payload[0..len], byte ∉ bad_bytes" or "exists offset such that overflow occurs and result < limit". Common shapes that domain encoders need but
  few people get right (esp. quantifier instantiation strategy in z3). Each helper would be ~5 lines but they're 5 correct lines that we'd get wrong on our own.
  4. Array theory primitives. For B2 (buffer OOB encoder), heap modelling, array-bounds reasoning. mk_array(name, idx_width, val_width), select, store, plus combined-theory tactics for QF_ABV. Without these, B2
  has to reinvent array reasoning poorly.
  5. Soundness audit / property-based tests for csem. Are our overflow predicates SMT-equivalent to what they claim? uadd_overflows(a, b) should be exactly equivalent to (a + b) < a in unsigned semantics. A property-based test that randomly samples inputs and compares csem against ground-truth z3 encoding would catch any encoding bugs and lock in correctness. SMT-experts do this kind of thing routinely.
  6. Tactics tuning for our workload. Our queries are mostly QF_BV with some QF_ABV. There's likely a Z3 tactic combination (e.g., simplify > propagate-values > solve-eqs > smt) that's significantly faster than
  default. Pure Z3 expertise, no RAPTOR context required.

How about I clear the backlog of ideas and you see what you can do re those 6 items? That would really make a profound difference. No obligation - but the first one in particular (optimisation) would be very helpful re the PoC generation stuff. I can do it without, but would have much higher quality results I think once that lands.

(And once we've done all that ... Symbolic-execution-driven path extraction as an alternative to the LLM-based extraction in dataflow_validator. That's a story for another day 😄)

grokjc:
See also #225 and #227.

Following those (I need to get #225 merged ...) I also have another PR almost finished:
                                                                                      
  - /understand --trace JSON gains optional path_conditions and path_profile fields (the LLM has full source context when it runs; better-positioned than Stage E to extract these correctly)                                                                                                                                                     
  - core/understand_bridge.py forwards both fields through to attack-paths.json, dropping malformed values with a warning                                                                                                                                                    
  - raptor-smt-validate-path gains --from-trace <id> --workdir <path> so Stage E passes a trace id; script reads attack-paths.json, picks conditions+profile, runs the solver. No bash pipelines on the LLM.

PR #209, opened 2026-04-23
Add release workflow: versioned archives and changelog on tag

danielcuthbert:
## What's this?

Adds a GitHub Action that fires whenever a `vX.Y.Z` tag is pushed (or manually triggered). It builds a clean release and posts it to GitHub Releases automatically.

## What it does

1. Finds the previous tag and generates a changelog from commits in between, grouped by prefix: `feat:`, `fix:`, `security:`, `docs:`, and everything else
2. Runs `git archive` to get a clean export, then strips `out/`, `codeql_dbs/`, `__pycache__`, `.venv`, `.vscode`, and `docs/img` -- things nobody needs in a release tarball
3. Builds both `.zip` and `.tar.gz`
4. Generates a `SHA256` checksums file
5. Creates a GitHub Release with all three files attached and the changelog as the release notes

## How to cut a release

```bash
git tag v3.1.0
git push origin v3.1.0
```

That's it. The Action handles the rest.

You can also trigger it manually from the Actions tab if you want to release from a specific tag without pushing it fresh.

## Commit prefix convention

The changelog groups commits by prefix. Not enforced, but `feat:`, `fix:`, `security:`, and `docs:` will land in their own sections. Anything without a prefix goes into "Other changes". This is documented in the Contributing section of the README.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,20 +28,33 @@ jobs:
         id: tag
         env:
           INPUT_TAG: ${{ github.event.inputs.tag }}
+          EVENT_NAME: ${{ github.event_name }}
         run: |
-          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+          if [ "$EVENT_NAME" = "workflow_dispatch" ]; then
             echo "name=${INPUT_TAG}" >> "$GITHUB_OUTPUT"
           else
             echo "name=${GITHUB_REF#refs/tags/}" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Validate tag
+        env:
+          TAG: ${{ steps.tag.outputs.name }}
+        run: |
+          if ! echo "${TAG}" | grep -qE '^v[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::Tag '${TAG}' does not match vX.Y.Z format"
+            exit 1
+          fi
+          if ! git merge-base --is-ancestor HEAD origin/main 2>/dev/null; then
+            echo "::warning::Tagged commit is not on main — verify this is intentional"
           fi
 
       - name: Generate changelog
         id: changelog
         env:
           TAG: ${{ steps.tag.outputs.name }}
+          REPO: ${{ github.repository }}
         run: |
-          # Find the previous tag
-          PREV_TAG=$(git tag --sort=-version:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | grep -v "^${TAG}$" | head -1)
+          PREV_TAG=$(git tag --sort=version:refname | grep -E '^v[0-9]+\.[0-9]+\.[0-9]+$' | awk -v tag="${TAG}" '$0 == tag {exit} {last=$0} END {print last}')
 
           if [ -z "$PREV_TAG" ]; then
             echo "No previous tag found, using all commits"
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,156 @@
...
+permissions:
+  contents: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
grokjc, line 22: workflow_dispatch will release the wrong commit.
  
  actions/checkout@v4 defaults to github.ref, which on a manual dispatch from the default branch is refs/heads/main — not the tag the operator typed in. Downstream you git log … HEAD and git archive HEAD, so a manual dispatch with input v3.0.5 would actually package up whatever's currently on main. The push-tag path is fine because Actions resolves the tag ref for you.
  
 ```
uses: actions/checkout@v4
  with:
    fetch-depth: 0
    ref: ${{ github.event.inputs.tag || github.ref }}
```  

You'll need to reference inputs.tag directly here since the resolve-tag step runs after checkout.

File: .github/workflows/release.yml [hunk 2]
```diff
@@ -0,0 +1,156 @@
...
+          ls -lh /tmp/${NAME}.*
+
+      - name: Create GitHub release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          TAG="${{ steps.tag.outputs.name }}"
+          NAME="raptor-${TAG}"
+
+          gh release create "${TAG}" \
```
grokjc, line 179: Add --verify-tag ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,156 @@
...
+          if [ -z "$PREV_TAG" ]; then
+            echo "No previous tag found, using all commits"
+            COMMITS=$(git log --pretty=format:"%s" HEAD)
+          else
+            echo "Comparing ${PREV_TAG}..${TAG}"
+            COMMITS=$(git log --pretty=format:"%s" "${PREV_TAG}..HEAD")
+          fi
+
+          # Group commits into sections
+          FEATURES=$(echo "$COMMITS" | grep -iE '^feat(\(.+\))?:' | sed 's/^feat[^:]*: /- /' || true)
```
grokjc: grep -iE '^feat((.+))?:' matches feat:no-space-here, but sed 's/^feat[^:]*: /- /' requires the literal space after the colon. Result: that commit lands in the changelog still wearing its feat: prefix.
  
  Either tighten the grep, or relax the sed:
  
  sed 's/^feat[^:]:[[:space:]]/- /'

.github/workflows/release.yml
```diff
@@ -0,0 +1,156 @@
...
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            echo "name=${{ github.event.inputs.tag }}" >> "$GITHUB_OUTPUT"
+          else
+            echo "name=${GITHUB_REF#refs/tags/}" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Generate changelog
+        id: changelog
+        run: |
+          TAG="${{ steps.tag.outputs.name }}"
```
grokjc: Untrusted-input hardening — switch ${{ … }} to env: / $VAR.
                                                                                                                                                                                                                   
  Actions string-substitutes ${{ … }} before the shell sees the line, so a tag containing "; …; " would execute on the runner. Threat model is restricted (anyone who can push a tag or run dispatch), so it's not urgent — but the fix is so cheap it's worth doing:                                                                                                                                                               
                                                                                                                                                                                                                   
  env:            
    TAG: ${{ steps.tag.outputs.name }}
  run: |                                                                                                                                                                                                           
    echo "$TAG"

(Applies to several places in this file)

grokjc (commented):
Two questions: 
- .github/ ends up inside the archive. Intentional? Some projects strip it so workflow internals don't ship with the release.
- No --draft / approval gate — pushing a tag publishes immediately. Fine if that's what you want, just flagging.

.github/workflows/release.yml
danielcuthbert: damn sed skills are getting worse ;( fixed

.github/workflows/release.yml
danielcuthbert: good call, should have

grokjc (commented):
**security: sed pattern mismatch**

The grep catches both security: and sec:, but the sed only strips sec[^:]*:. A commit like security(auth): fix token leak would render as urity(auth): fix token leak in the changelog.                          
                                                                                                                                                                                                                   
```
# current                                                                                                                                                                                                        
SECURITY=$(echo "$COMMITS" | grep -iE '^security(\(.+\))?:|^sec(\(.+\))?:' | sed 's/^sec[^:]*:[[:space:]]*/- /' || true)

# fix
SECURITY=$(echo "$COMMITS" | grep -iE '^security(\(.+\))?:|^sec(\(.+\))?:' | sed 's/^sec\(urity\)\?[^:]*:[[:space:]]*/- /' || true)
```                                                                                                                                                                                                                   

**Exclude .github/ from release archives**
                                                                                                                                                                                                                   
CI internals don't need to ship. Cleanest way is a .gitattributes at the repo root:                                                                                                                              
   
 ```.github/ export-ignore```

  git archive picks this up automatically — no workflow changes needed.

**Exclude stray dev files from archives**                                                                                                                                                                         
   
.env, *.patch, .claude/ etc. would end up in the tarball if they exist at tag time. Same .gitattributes approach:                                                                                                
 ```                                                                                                                                                                                                                
.github/       export-ignore                                                                                                                                                                                     
.env           export-ignore                                                                                                                                                                                   
*.patch        export-ignore
.claude/       export-ignore
```
  **One remaining ${{ }} in a run: block**                                                                                                                                                                          

The "Full diff" URL still interpolates ${{ github.repository }} inline. Low risk since it's GitHub-controlled, but for consistency with the other hardening in 0f72054:                                          

 Add to the "Generate changelog" step's env: block
```
REPO: ${{ github.repository }}                                                                                                                                                                               
```                                                                                                                                                                                                                   
then replace
```
CHANGELOG="${CHANGELOG}**Full diff:** https://github.com/${{ github.repository }}/compare/${PREV_TAG}...${TAG}"
```
with                                                                                                                                                                                                           
```
CHANGELOG="${CHANGELOG}**Full diff:** https://github.com/${REPO}/compare/${PREV_TAG}...${TAG}"
```

grokjc:
We should probably start editing the version in raptor-offset when we push a release too. Perhaps that can be automated or templated as part of the workflow. Anywhere else that mentions version number? 🤔

grokjc:
All remaining review items addressed:

  sed mismatch (Comment 4a): Replaced all four prefix-specific seds with a generic sed -E 's/^[^:]+:[[:space:]]*/- /'. Since grep already filters lines by prefix, the sed just strips everything up to the first
  colon. Handles security(auth):, SEC:, and any future prefix variant without separate patterns.

  .gitattributes (Comment 4b/4c): Added with export-ignore for .github/, .claude/, .devcontainer/, .vscode/, .env, *.patch, out/, codeql_dbs/, docs/img/, test/, conftest.py. The manual rm -rf and find cleanup in
   the archive step is now gone — git archive respects the attributes natively.

  ${{ github.repository }} (Comment 4d): Moved to REPO env var. Also moved github.event_name to EVENT_NAME. The only ${{ }} remaining are in env: blocks, which Actions evaluates before the shell touches them.
The workflow now stamps the version into two places:

  - raptor-offset — the banner line (Based on Claude Code - vX.Y.Z), with padding recalculated to keep the box aligned
  - core/config.py — RaptorConfig.VERSION (v-prefix stripped, so VERSION = "3.1.0")

  The stamp happens on a detached HEAD, committed as release: stamp vX.Y.Z, and the tag is force-updated to point at that commit (annotated, not lightweight). Main stays untouched. The regex matches generically
  — doesn't depend on the current version text being any specific string.

  Re-runs are idempotent: if the version is already stamped, git diff --cached --quiet skips the commit, and gh release delete + gh release create replaces the release cleanly.

  A few more things fixed while in here:

  - Tag validation: New step fails fast if the tag doesn't match vX.Y.Z (catches workflow_dispatch typos like main or v1.2). Also warns if the tagged commit isn't an ancestor of main.
  - PREV_TAG for out-of-order releases: Changed from "highest tag that isn't current" to "tag immediately before current in version order" using ascending sort + awk. If you ever cut a v3.1.1 hotfix after v3.2.0
   exists, the changelog now correctly compares against v3.1.0 instead of v3.2.0.
  - Changelog noise filter: release:, chore:, ci:, test:, build:, style: prefixes are now excluded from "Other changes". Without this, every release's stamp commit would appear in the next release's changelog.
  - Annotated tags: The retag step now creates annotated tags (git tag -a -f) with a release message, instead of lightweight tags.
  - Test script: test/test_release_workflow.sh — 60 assertions covering tag validation, provenance, PREV_TAG resolution, changelog categorization, prefix stripping, version stamping (both files, box alignment,
  idempotency), and archive exclusions. Runs against a disposable temp repo in ~2 seconds.

PR #211, opened 2026-04-23
ci: expand Python test coverage

Hinotoi-agent:
## Summary
- add a compile step for RAPTOR Python sources in CI
- expand the unit-test job to run `core` tests in addition to the existing package tests
- fix a startup banner f-string syntax issue that the new compile step catches

## Why
The current unit-test job only runs `packages/*/tests`, so regressions in `core/` can be missed. Adding `python -m compileall` gives CI a fast syntax/import-preflight gate before pytest starts, and including `core` in the pytest target brings the project-management, reporting, run metadata, coverage, SARIF, security, and startup helpers into the normal PR check path.

## Validation
- `python -m compileall -q core packages raptor.py raptor_agentic.py raptor_codeql.py raptor_fuzzing.py` passes under Python 3.12
- `pytest core -q` passes locally under Python 3.12: 687 passed, 9 skipped
- `pytest core packages/*/tests -q` was also run locally under Python 3.12; the expanded suite reaches 1967 passed / 33 skipped, with 4 existing macOS-specific package failures unrelated to this CI change. The GitHub workflow runs on Ubuntu.
- `git diff --check` passes

## Notes
This is intentionally a small CI hardening PR. It avoids overlapping with the larger open sandbox/release PRs and only broadens the existing Python verification surface.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -71,10 +71,15 @@ jobs:
       - name: Restore venv permissions
         run: chmod +x $VENV_PATH/bin/*
 
+      - name: Compile Python sources
+        run: |
+          source $VENV_PATH/bin/activate
+          python -m compileall -q core packages raptor.py raptor_agentic.py raptor_codeql.py raptor_fuzzing.py
+
       - name: Run unit tests
         run: |
           source $VENV_PATH/bin/activate
-          pytest packages/*/tests
+          pytest core packages/*/tests
 
   test-workflows:
     needs: deps
--- a/core/startup/banner.py
+++ b/core/startup/banner.py
@@ -50,7 +50,7 @@ def format_banner(logo, quote, tool_results, tool_warnings, llm_lines,
         lines.append("")
 
     # Tools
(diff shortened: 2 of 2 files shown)


grokjc:
This is exactly the kind of quiet CI improvement that pays for itself. 🙏                                                                                                                             
                                                                                                                                                                                                                   
  A few notes for anyone following along:                                                                                                                                                                          
                                                                                                                                                                                                                   
  - core/ was a real coverage gap: the existing core/coverage, core/inventory, core/json, core/project, core/reporting suites weren't being exercised in CI, so this closes a genuine hole. Bonus: in-flight work on core/sandbox/ and core/security/ test suites will pick up CI coverage automatically once rebased on this.
  - compileall preflight is cheap and catches a class of bugs pytest wouldn't (bad imports, missing deps, f-string syntax snafus) before the slower test stage runs.                                               
  - The banner.py escape change is effectively cosmetic on our 3.12 (PEP 701 allows backslashes inside f-string expressions from 3.12 onward), but it's a defensive improvement for anyone running on ≤3.11 and doesn't hurt readability.
                                                         
Thanks for the contribution!

PR #212, opened 2026-04-23
ci: cover remaining Python test files

Hinotoi-agent:
## Summary
- broaden the Python unit-test job from `pytest core packages/*/tests` to `pytest core packages`
- compile all top-level Python entry points with `python -m compileall -q core packages *.py`
- make the deeply nested Claude trust test portable by avoiding `json.dumps()` recursion before the scanner runs

## Why
PR #211 closed the biggest CI gap by adding `core` tests. This follow-up closes the remaining Python collection gap: `packages/*/tests` misses tests that live directly under a package directory, such as `packages/codeql/test_database_manager.py`.

The compile step also now covers top-level helper scripts such as `build_inventory.py` and `generate_diagram.py`, not only the `raptor*.py` launchers.

## Validation
- `python -m compileall -q core packages *.py` passes locally
- `pytest core -q` passes locally: 687 passed, 9 skipped
- `pytest packages/codeql/test_database_manager.py -q` passes locally: 11 passed
- `git diff --check upstream/main...HEAD` passes

## Notes
A full local `pytest packages -q` run on macOS still reaches the existing platform-specific package failures documented in #211. The GitHub workflow runs on Ubuntu; this PR is intended to make CI collect the remaining Python tests there rather than silently skipping them.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -74,12 +74,12 @@ jobs:
       - name: Compile Python sources
         run: |
           source $VENV_PATH/bin/activate
-          python -m compileall -q core packages raptor.py raptor_agentic.py raptor_codeql.py raptor_fuzzing.py
+          python -m compileall -q core packages *.py
 
       - name: Run unit tests
         run: |
           source $VENV_PATH/bin/activate
-          pytest core packages/*/tests
+          pytest core packages
 
   test-workflows:
     needs: deps
--- a/core/security/tests/test_cc_trust.py
+++ b/core/security/tests/test_cc_trust.py
@@ -226,12 +226,12 @@ class TestEnvInjection:
         fail-closed scan wrapper must catch it."""
         claude = tmp_path / ".claude"; claude.mkdir()
         depth = sys.getrecursionlimit() + 500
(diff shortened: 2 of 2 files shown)


grokjc:
Nice follow-up to #211 — this scoops up the loose ends neatly. 

  - pytest core packages catches packages/codeql/test_database_manager.py, which was the one orphaned test file living outside a tests/ subdir and silently skipped by the old glob.
  - compileall ... *.py is exactly the glob-over-hardcoded-list tidy-up we needed - now build_inventory.py and generate_diagram.py are covered too.
  - The test_deeply_nested_env_value_does_not_crash refactor is a real latent bug fix. 
 
The old version built a getrecursionlimit() + 500 deep Python dict and handed it to json.dumps() — but json.dumps recurses on nested containers and would RecursionError before the file was ever written, meaning the scanner's fail-closed path was never actually exercised. Because core/*/tests wasn't running in CI until #211, nobody noticed. Constructing the JSON as a string sidesteps the encoder recursion and lets the scanner hit its str()/repr() path, which is what the test is supposed to cover. 

Thanks again!

PR #213, opened 2026-04-23
test: strengthen LLM log secret redaction

Hinotoi-agent:
## Summary
- Strengthen `_sanitize_log_message()` tests with realistic fake token shapes.
- Extend redaction coverage for Bearer/JWT credentials, Basic auth headers, GitHub tokens, AWS access key IDs, JSON/key-value secret fields, and PEM private-key blocks.
- Add a false-positive regression check so LLM usage telemetry like `prompt_tokens` / `completion_tokens` remains visible.
- Keep fake tokens constructed from string fragments so the tests do not introduce realistic-looking literal secrets.

## Why
RAPTOR handles LLM provider errors and external tool output where credentials can appear in exception messages, auth headers, environment-style key/value lines, JSON-ish SDK errors, or subprocess output. Stronger redaction tests help prevent API keys and tokens from leaking into logs during scans while preserving useful non-secret diagnostics.

## Test plan
- `python -m pytest packages/llm_analysis/tests/test_llm_callbacks.py::TestSanitizeLogMessage -q`
- `python -m pytest packages/llm_analysis/tests -q`
- `python -m compileall -q packages/llm_analysis/llm/client.py packages/llm_analysis/tests/test_llm_callbacks.py`
- `python -m compileall -q core packages raptor.py raptor_agentic.py raptor_codeql.py raptor_fuzzing.py`
- `git diff --check`
- Added-line security grep for hardcoded secret literals and dangerous Python sinks

## Notes
A broader local run of `pytest core packages/llm_analysis/tests -q` had pre-existing environment/Python-version related failures outside this patch. The touched package suite passes cleanly.

### DIFF
--- a/packages/llm_analysis/llm/client.py
+++ b/packages/llm_analysis/llm/client.py
@@ -89,8 +89,9 @@ def _sanitize_log_message(msg: str) -> str:
     # such as PASSWORD_POLICY, SECRET_ROTATION_DAYS, MAX_API_KEY_LENGTH, or
     # pagination cursors like page_token/next_token.
     secret_field = (
-        r'(?:[A-Za-z0-9_-]*(?:API[_-]?KEY|PASSWORD|SECRET|'
+        r'(?:[A-Za-z0-9_-]*(?:API[_-]?KEY|PASSWORD|'
         r'SECRET[_-]?KEY|SECRET[_-]?ACCESS[_-]?KEY)'
+        r'|(?:CLIENT|APP|SHARED|API|CONSUMER)[_-]?SECRET'
         r'|(?:ACCESS|AUTH|BEARER|ID|REFRESH|SESSION|SERVICE)[_-]?TOKEN)'
     )
     # Quoted values may be short or contain spaces/commas; the field name marks them sensitive.
--- a/packages/llm_analysis/llm/client.py
+++ b/packages/llm_analysis/llm/client.py
@@ -59,7 +59,17 @@ def _sanitize_log_message(msg: str) -> str:
     # Redact Google API keys (AIza*)
     msg = re.sub(r'AIza[a-zA-Z0-9-_]{30,}', '[REDACTED-API-KEY]', msg)
     # Redact Bearer tokens (Mistral and others in auth headers/error messages)
-    msg = re.sub(r'Bearer [a-zA-Z0-9-_]{20,}', 'Bearer [REDACTED]', msg)
+    msg = re.sub(
+        r'Bearer [a-zA-Z0-9._~+/-]{20,}={0,2}',
+        'Bearer [REDACTED]',
+        msg,
+        flags=re.IGNORECASE,
+    )
+    # Redact GitHub tokens that may appear in git/gh subprocess output
+    msg = re.sub(r'gh[oprsu]_[a-zA-Z0-9_]{36,}', '[REDACTED-API-KEY]', msg)
+    msg = re.sub(r'github_pat_[a-zA-Z0-9_]{20,}', '[REDACTED-API-KEY]', msg)
+    # Redact AWS access key IDs that commonly appear in tool output/traces
+    msg = re.sub(r'\b(?:AKIA|ASIA)[A-Z0-9]{16}\b', '[REDACTED-API-KEY]', msg)
     return msg
 
 
--- a/packages/llm_analysis/llm/client.py
+++ b/packages/llm_analysis/llm/client.py
@@ -51,9 +51,10 @@ def _sanitize_log_message(msg: str) -> str:
     Searchable tags: #SECURITY #API_KEY_PROTECTION #LOG_SANITIZATION
     Related: Cursor Bot Bug #2, PR #32, defense-in-depth best practice
     """
-    # Redact multiline private key material before shorter generic patterns.
+    # Redact private key material before shorter generic patterns. If a log line
+    # is truncated before the END marker, redact through the end of the message.
     msg = re.sub(
-        r'-----BEGIN [A-Z ]*PRIVATE KEY-----.*?-----END [A-Z ]*PRIVATE KEY-----',
+        r'-----BEGIN [A-Z ]*PRIVATE KEY-----.*?(?:-----END [A-Z ]*PRIVATE KEY-----|$)',
         '[REDACTED-PRIVATE-KEY]',
         msg,
         flags=re.DOTALL,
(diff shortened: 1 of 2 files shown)


packages/llm_analysis/llm/client.py
```diff
@@ -51,15 +51,57 @@ def _sanitize_log_message(msg: str) -> str:
...
-    msg = re.sub(r'Bearer [a-zA-Z0-9-_]{20,}', 'Bearer [REDACTED]', msg)
+    # Redact common authorization header schemes from SDK/tool errors.
+    msg = re.sub(
+        r'Bearer [a-zA-Z0-9._~+/-]{20,}={0,2}',
+        'Bearer [REDACTED]',
+        msg,
+        flags=re.IGNORECASE,
+    )
+    msg = re.sub(
+        r'Basic\s+\S+',
```
grokjc: ```Basic\s+\S+``` with IGNORECASE will match normal English like "basic error" or "basic setup" in error messages. Maybe tighten to base64-shaped content, something like ```Basic\s+[A-Za-z0-9+/]{8,}={0,2}```.

packages/llm_analysis/llm/client.py
```diff
@@ -51,15 +51,57 @@ def _sanitize_log_message(msg: str) -> str:
...
+        msg,
+        flags=re.IGNORECASE,
+    )
+    # Redact GitHub tokens that may appear in git/gh subprocess output
+    msg = re.sub(r'gh[oprsu]_[a-zA-Z0-9_]{36,}', '[REDACTED-API-KEY]', msg)
+    msg = re.sub(r'github_pat_[a-zA-Z0-9_]{20,}', '[REDACTED-API-KEY]', msg)
+    # Redact AWS access key IDs that commonly appear in tool output/traces
+    msg = re.sub(r'\b(?:AKIA|ASIA)[A-Z0-9]{16}\b', '[REDACTED-API-KEY]', msg)
+    # Redact key/value or JSON-ish assignments such as API_KEY=... or "token": "...".
+    secret_field = (
```
grokjc, line 91: The containment match on SECRET/PASSWORD/API_KEY will catch metadata fields that describe secrets but aren't secrets themselves -SECRET_ROTATION_DAYS=90, PASSWORD_POLICY="strict", MAX_API_KEY_LENGTH=64 all get redacted. Might be worth anchoring to start/end of the field name, or just a comment acknowledging the over-redaction tradeoff.                                                                              
   
Same thing for TOKEN - correctly skips prompt_tokens and completion_tokens (trailing s breaks the word boundary, nice), but pagination fields like page_token and next_token get caught.

packages/llm_analysis/tests/test_llm_callbacks.py
grokjc, line 1: No test for ASIA AWS keys -- the regex covers both AKIA and ASIA but only AKIA is tested. Quick extra line.

No test for a partial PEM block -- if a log has the BEGIN marker but the message gets truncated before END, the regex won't match and key material leaks. Worth a test to document that known gap even if you don't fix it now.

The github token splits feel a bit arbitrary -- "gh" + "p_" + "e" * 36 is avoiding a literal ghp_ prefix which isn't itself a secret. Splitting at the boundary between prefix and random part ("ghp_" + "e" * 36) would read more naturally. Same for "github" + "_pat_".

No boundary-length tests -- Bearer requires 20+ chars, GitHub tokens require 36+. A test with 19 or 35 chars confirming no match would catch off-by-one regressions if someone tweaks the thresholds later.

grokjc (changes requested):
Overall this is a solid expansion - PEM-first ordering is right, the JWT bearer fix is a real gap closed, and building test tokens from fragments is a nice touch. The false positive surface on Basic and the field-name containment matching are the main things worth tightening.

Tests are thorough - good coverage of the new patterns and the preservation check for token metrics is a nice touch. A few small gaps: no test for ASIA AWS keys (only AKIA), no boundary-length tests (e.g. 19 chars not matching the 20-char Bearer threshold), and might be worth a test for a truncated PEM block without the END marker to document that edge case.

packages/llm_analysis/llm/client.py
Hinotoi-agent: Tightened Basic auth redaction to base64-shaped credentials (`Basic\s+[A-Za-z0-9+/]{8,}={0,2}`) and added a regression test confirming plain-English `basic error` / `basic setup` messages are preserved. Fixed in 3d7f5dde1633309fe0f8e27c6945c646202b9dcd.

packages/llm_analysis/llm/client.py
Hinotoi-agent: Tightened the key/value field matching so it no longer redacts metadata fields like `SECRET_ROTATION_DAYS`, `PASSWORD_POLICY`, `MAX_API_KEY_LENGTH`, or pagination cursors like `page_token` / `next_token`, while still covering explicit secret fields. Added regression coverage in 3d7f5dde1633309fe0f8e27c6945c646202b9dcd.

packages/llm_analysis/tests/test_llm_callbacks.py
Hinotoi-agent: Addressed these test gaps in 3d7f5dde1633309fe0f8e27c6945c646202b9dcd: added ASIA AWS key coverage, boundary-length checks for Bearer/GitHub token thresholds, a truncated PEM regression that now redacts through message end, and adjusted GitHub token test construction to split at the natural prefix boundary.

grokjc:
Thanks for the quick turnaround - the Basic regex fix, the TOKEN prefix list, and the new tests all look good. One last thing: bare SECRET as a standalone alternative still catches fields like IS_SECRET that aren't credentials. Same fix you already did for TOKEN works here - explicit prefix list:

  ```r'(?:CLIENT|APP|SHARED|API|CONSUMER)[_-]?SECRET'```

That gets the real ones (CLIENT_SECRET, APP_SECRET, SHARED_SECRET, etc.) without the false positives. Happy to approve once that's in.

Hinotoi-agent:
Thanks — fixed in `f4857a8466f720e812ce99324763f71a54d8b8c0`.

I removed the bare `SECRET` field alternative and replaced it with the explicit prefix form:

```python
r'(?:CLIENT|APP|SHARED|API|CONSUMER)[_-]?SECRET'
```

I also updated the regression coverage so `IS_SECRET` is preserved as metadata/non-credential context, while `CLIENT_SECRET` remains redacted.

grokjc:
Thanks again! Appreciate the contribution(s) ❤️

PR #219, opened 2026-04-24
[security] fix(web): enforce WebClient target scope across redirects

Hinotoi-agent:
## Summary

This PR hardens RAPTOR's web scanner so HTTP requests stay within the operator-configured target origin, including across redirects.

Before this change, `WebClient` built request URLs with `urljoin()` and delegated redirect handling to `requests` with `allow_redirects=True`. The crawler constrained discovered links to the same netloc, but a target-controlled HTTP redirect could still move the scanner to an off-scope host after the in-scope request was sent.

This patch adds explicit target-origin validation for direct request URLs and manually follows only same-origin redirects.

## Security issues covered

| Issue | Status in this PR |
| --- | --- |
| Cross-origin redirect scope escape in web scanner requests | Fixed |
| Protocol-relative / absolute URL target-scope escape through `WebClient` callers | Fixed |
| Cookie exposure to off-scope redirect targets when cookies are configured on `WebClient` | Fixed by blocking the off-scope redirect before the redirected request is made |

## Before this PR

- `WebClient.get()` and `WebClient.post()` accepted URLs built with `urljoin(self.base_url, path)`.
- Absolute or protocol-relative paths could replace the configured target origin.
- `requests` followed redirects automatically.
- The crawler's same-netloc filtering applied to discovered links before requests, but not to HTTP redirect targets after the request.
- If callers configured cookies through `WebClient.set_cookies()`, an off-scope redirect could receive those cookies.

## After this PR

- `WebClient` normalizes target scope as `(scheme, host, port)`.
- Direct request URLs are rejected if they leave the configured base origin.
- Redirects are handled manually with `allow_redirects=False`.
- Same-origin redirects continue to work.
- Cross-origin redirects are blocked before any request is made to the redirected host.
- Regression tests cover absolute URLs, protocol-relative URLs, same-origin redirects, cross-origin redirects, and cookie non-leakage.

## Why this matters

RAPTOR is a pentest tool, so active scanning and fuzz payloads are expected behavior. The security boundary here is the operator's authorized target scope.

A malicious or compromised in-scope target should not be able to silently redirect RAPTOR's scanner traffic to off-scope third-party or internal services. Keeping web scanner requests inside the configured target origin helps prevent accidental unauthorized scanning and prevents configured session cookies from being sent to an attacker-controlled redirect destination.

## Attack flow

```text
1. Operator configures RAPTOR to scan https://authorized.example
2. authorized.example returns a redirect such as:
   Location: http://127.0.0.1:8080/internal
3. Prior behavior: requests follows the redirect automatically
4. RAPTOR sends traffic to the off-scope host
5. If WebClient cookies were configured, those cookies can accompany the redirected request
```

## Affected code

| Area | Files |
| --- | --- |
| Web scanner HTTP client | `packages/web/client.py` |
| Regression coverage | `packages/web/tests/test_client_scope.py` |

## Root cause

- The crawler scoped discovered links before requesting them, but redirect handling was delegated to `requests`.
- `requests` follows redirects unless told otherwise.
- `WebClient` did not re-check target scope after receiving a redirect `Location` header.
- `WebClient` also did not reject absolute or protocol-relative URL inputs that changed the target origin.

## CVSS assessment

- Issue: Web scanner cross-origin redirect scope escape
- CVSS v3.1: 5.4 Medium
- Vector: `AV:N/AC:L/PR:N/UI:R/S:U/C:L/I:N/A:L`

Rationale: an attacker-controlled web target can trigger the behavior remotely after an operator scans that target. User interaction is required because the operator must initiate the scan. Impact is bounded to off-scope scanner traffic and potential low-confidentiality cookie exposure when cookies are configured through the client API.

## Safe reproduction steps

A safe local reproduction uses two loopback HTTP servers:

1. Start an in-scope server on one local port.
2. Start an off-scope sink server on a different local port.
3. Configure the in-scope server to return:
   `302 Location: http://127.0.0.1:<off-scope-port>/internal-metadata`
4. Create a `WebClient` for the in-scope server.
5. Call `client.get("/")`.
6. On vulnerable code, the off-scope sink receives the redirected request.
7. On this PR, `WebClient` raises `ValueError` before the off-scope request is made.

The new tests implement this flow without external network access.

## Expected vulnerable behavior

On vulnerable code:

- `WebClient` follows a cross-origin redirect automatically.
- The final response URL is the off-scope URL.
- The off-scope sink receives the request.
- If cookies were configured with `WebClient.set_cookies()`, the sink can receive those cookies.

## Changes in this PR

- Added target-origin normalization and scope checks in `WebClient`.
- Added URL construction helper that rejects URLs outside the configured target origin.
- Replaced automatic redirect following with manual redirect handling.
- Allowed same-origin redirects to preserve existing crawler behavior.
- Blocked cross-origin redirects before a request is made to the redirected host.
- Added regression tests for direct URL scope checks and redirect scope checks.

## Files changed

| File | What changed |
| --- | --- |
| `packages/web/client.py` | Adds request URL scope checks and scoped redirect handling for GET/POST |
| `packages/web/tests/test_client_scope.py` | Adds local HTTP-server regression tests for target scope and cookie non-leakage |

## Maintainer impact

- Existing relative URL scans continue to work.
- Same-origin redirects continue to work.
- Intentional cross-origin redirect crawling is now blocked by default to preserve the configured target boundary.
- No external services are used by the tests.

## Fix rationale

Target scope should be enforced at the lowest HTTP client layer, not only in the crawler. This keeps future callers of `WebClient` covered even if they bypass crawler link discovery or pass absolute URLs directly.

Manual redirect handling makes the security boundary explicit and prevents sensitive scanner state, such as cookies, from being sent to off-scope redirect targets.

## Type of change

- [x] Security hardening
- [x] Bug fix
- [x] Tests
- [ ] Documentation only
- [ ] Refactor only

## Test plan

Commands run locally:

```text
python3.11 -m pytest packages/web/tests/test_client_scope.py -q
python3.11 -m pytest packages/web/tests -q
python3.11 -m pytest core/security/tests packages/llm_analysis/tests -q
python3.11 -m compileall -q packages/web/client.py packages/web/tests/test_client_scope.py
git diff --check
```

Results:

```text
6 passed — packages/web/tests/test_client_scope.py
10 passed — packages/web/tests
431 passed, 5 skipped — core/security/tests packages/llm_analysis/tests
compileall touched files — passed
git diff --check — passed
```

I also attempted the full suite with `python3.11 -m pytest -q`. It completed with unrelated existing environment/platform failures outside this change area, including tree-sitter C parameter extraction, a subprocess path that invoked `python3` with unsupported `str | None` syntax, macOS `/private/var` temp path differences, Solaris cross-platform detection expectations, and local glibc confidence detection on macOS.

## Disclosure notes

This PR is intentionally scoped to RAPTOR's web scanner target-boundary enforcement. It does not claim that scanning or fuzz payload generation is inherently unsafe; those are expected features for a pentest tool. The issue addressed here is the narrower boundary problem where a configured target can redirect scanner traffic outside the operator's intended authorization scope.


danielcuthbert:
good catch, we have a mega sandbox PR currently being reviewed but yeah this helps. thank you

grokjc:
Things worth flagging:

  1. No test for 307/308 method preservation — the tests cover 302 but the method-preservation path for 307/308 is untested. Would be good to add since it's the more complex branch.
  2. No test for _MAX_REDIRECTS exhaustion — the TooManyRedirects path isn't exercised.
  3. Merge conflict with PR #223 — both PRs modify packages/web/client.py (imports, __init__ signature, _log_request). Whichever merges second will need conflict resolution.
  4. test_same_origin_redirect doesn't use the _server() context manager — manually creates/tears down the server while the other tests use the helper. Minor inconsistency.
  5. Logged URL is always the initial request URL, not the final destination — _log_request('GET', url, ...) in get() uses the pre-redirect URL. After a redirect chain, the request history says the client hit /start when the response actually came from /final. Pre-existing behavior but now more visible with manual redirects.

  Nothing blocking. Solid contribution — I'd approve with a suggestion to add the 307/308 and max-redirect tests.

Hinotoi-agent:
Thanks for the review. I pushed cb6f674127093b1c47e4c4f0bc2b9f3fb1158c49 addressing the redirect follow-up items:

- Added 307/308 redirect coverage asserting POST method/body preservation.
- Added `_MAX_REDIRECTS` exhaustion coverage for `TooManyRedirects`.
- Switched `test_follows_same_origin_redirect` to the shared `_server()` context manager.
- Updated request history logging to record the final response URL after an allowed redirect.

Validation run:
- `uv run ruff check packages/web/client.py packages/web/tests/test_client_scope.py` passed.
- `python3.11 -m pytest packages/web/tests/test_client_scope.py -q` passed: 8 passed.
- `python3.11 -m pytest packages/web -q` passed: 12 passed.
- `git diff --check` passed.
- `uv run pytest -q` ran; it still has 5 pre-existing unrelated local/macOS failures outside the touched web client area (`core.inventory` tree-sitter C params and `packages.exploit_feasibility` platform/temp-path checks).

grokjc:
Thanks for the quick turnaround on the follow-up items, the 307/308 coverage and exhaustion test are exactly what was needed. The redirect-aware logging is a nice touch too.

The 5 test failures in your run were pre-existing cross-platform issues (macOS /proc paths, tree-sitter-c param extraction, tmpdir symlink resolution) that affected all branches, fixed in 3785de3.

Back to @danielcuthbert for the functional review.

grokjc:
CI failure on the merge is a conflict between this PR and #223 (now on main). The redaction tests monkeypatch session.get to inject exceptions, but this PR changes get() to go through _send_scoped_request() which calls session.request() instead. The monkeypatch misses, the real request fires, and DNS resolution fails on example.test.                                                                                 
                                                                                                                                  
Fix: update the two redaction tests (test_web_client_redacts_timeout_urls_in_logs and test_web_client_redacts_request_exception_urls_in_logs) to monkeypatch session.request instead of session.get.

Hinotoi-agent:
Thanks for the heads-up — this was the #219/#223 merge interaction you identified.

Root cause:
- #219 routes `WebClient.get()` / `post()` through `_send_scoped_request()`.
- `_send_scoped_request()` uses `session.request(...)` directly so it can enforce redirect scope consistently.
- The redaction log tests added from #223 were still monkeypatching `session.get` / `session.post`.
- That monkeypatch no longer intercepted the request path after the merge, so the tests made a real request to `example.test` and failed in DNS resolution.

I pushed `13f2719e791acf76f5716042ec4519d60d691538` with the minimal fix:
- `test_web_client_redacts_timeout_urls_in_logs` now monkeypatches `client.session.request`.
- `test_web_client_redacts_request_exception_urls_in_logs` now monkeypatches `client.session.request`.

Validation:

```text
uv run pytest packages/web/tests/test_secret_redaction.py::test_web_client_redacts_timeout_urls_in_logs packages/web/tests/test_secret_redaction.py::test_web_client_redacts_request_exception_urls_in_logs -q
# 2 passed

uv run python -m compileall -q core packages *.py
uv run pytest packages/web/tests/test_client_scope.py packages/web/tests/test_secret_redaction.py packages/web -q
# 21 passed

uv run pytest core/security/tests packages/llm_analysis/tests -q
# 477 passed, 5 skipped

git diff --check
```

GitHub Actions are green again on the latest head.

danielcuthbert (approved):
Legit issue and one that I think many would be impacted by, so appreciate you helping with this. LGTM

PR #221, opened 2026-04-24
Set Codeql `--build-mode=none` through `--no-build` in raptor codeql

unprovable:
Fairly self explanatory - sometimes codeql will fail when building a database from building code (not so much an issue for languages like python, but if c/c++ fail then it's annoying). One way through is to allow codeql to have `build-mode: none` set through command line. This slightly clunky PR aims to do just that by adding the option and exposing it to the agent in control.


unprovable:
Actually, this is a _terrible_ idea. Closing as I can't find a way to make it work without Claude trying to bypass the full analysis toolchain every time. Reverting, removing, hoping the godz forgive us.

PR #210, opened 2026-04-23
Subprocess sandbox: namespaces, Landlock, seccomp, egress proxy, rlimits

grokjc:
New core/sandbox/ package wraps subprocess calls with layered isolation for any command processing untrusted content (LLM PoCs, target builds, scan/fuzz runs). Per-layer graceful fallback — missing kernel features narrow isolation rather than break callers.

Layers (full profile):
- user/pid/ipc/net namespaces via unshare
- mount-ns via ctypes + newuidmap in a fork-based spawn path (mount ops run BEFORE landlock_restrict_self so kernel 6.15+ Landlock does not block them)
- Landlock FS + TCP-port rules, ABI-aware, fail-closed on restrict_self failure
- seccomp-bpf blocking keyctl/bpf/userfaultfd/perf_event_open/ io_uring/ptrace(full)/kcmp/handle_at, AF_UNIX/PACKET/NETLINK/ SOCK_RAW sockets, TIOCSTI/TIOCCONS/TIOCSCTTY ioctls
- rlimits (mem/cpu/fsize/core + nproc via prlimit inside the unshare chain so RLIMIT_NPROC counts against the ns-local UID)

Entry points: run_untrusted() (strict default for attacker-derived code: restrict_reads + fake_home + block_network + full seccomp), sandbox() context (per-call config with use_egress_proxy for HTTPS-CONNECT hostname allowlist), run_trusted() (rlimits only), --sandbox {full,debug,network-only,none} / --no-sandbox CLI.

Egress proxy (core/sandbox/proxy.py): async CONNECT tunnel on loopback-only ephemeral port, hostname allowlist (UNION across callers), resolved-IP screening (private/loopback/link-local rejected), upstream HTTPS_PROXY + NO_PROXY honoured, per-sandbox event buffer with caller_label stamping. Pairs with seccomp UDP block to close DNS/UDP exfil.

core/security/ (new): centralised env_sanitisation (allowlist + blocklist applied by RaptorConfig.get_safe_env and at the sandbox boundary) and log_sanitisation (escape_nonprintable for cmd_display, ASAN bug_type, proxy CONNECT target — defends against ANSI-escape injection).

Observability: result.sandbox_info carries crashed/signal/ resource_exceeded/sanitizer/blocked/proxy_events; proxy events also persisted to {output}/proxy-events.jsonl. Startup banner shows active layers (sandbox ✓ (net+mount+landlock+seccomp)) plus one-time warnings for partial-sandbox runs.

[...]

PR #223, opened 2026-04-25
[security] feat(web): make secret redaction operator-configurable

Hinotoi-agent:
## Summary

This PR makes secret redaction operator-configurable while keeping RAPTOR safe by default.

RAPTOR now redacts common secrets from web scanner request history, web fuzzer finding URLs, and LLM/provider log sanitization unless the operator explicitly opts into preserving exact values for local debugging/troubleshooting.

Operators can opt in with either:

```bash
RAPTOR_REVEAL_SECRETS=true
```

or, for web scans:

```bash
python raptor.py web --url https://example.test --reveal-secrets
```

## Security issues covered

| Area | Issue | Change |
| --- | --- | --- |
| Web scanner artifacts | Request history and finding URLs can contain credentials from query strings or URL userinfo | Redact sensitive URL components by default |
| LLM/provider logs | Existing sanitizer always redacted, with no explicit local-debug override | Centralized operator opt-in can preserve exact values when intentionally enabled |
| Operator workflow | Debugging sometimes needs exact URLs/tokens to reproduce scanner behavior | Add explicit reveal mode instead of silently retaining secrets |

## Before this PR

- `packages/web/client.py` stored full request URLs in `request_history`.
- `packages/web/fuzzer.py` stored full target URLs in findings.
- `_sanitize_log_message()` had no central operator-level reveal setting.
- Operators had no consistent configuration knob for choosing redacted vs exact artifacts.

## After this PR

- Secret redaction is default-on.
- `RAPTOR_REVEAL_SECRETS=1/true/yes/on` enables exact-value retention.
- `/web` exposes `--reveal-secrets` for explicit local troubleshooting.
- Web request history and fuzzer finding URLs share the same redaction policy.
- LLM/provider log sanitization honors the same central setting.

## Why this matters

RAPTOR is a security scanner, so its logs and findings are often shared with maintainers, customers, or collaborators. Those artifacts can naturally include credentials found during authorized testing, especially in query strings, URL userinfo, and authorization headers.

The safe default should be to redact secrets from shareable artifacts. At the same time, operators sometimes need exact values to reproduce scanner behavior locally. This PR supports both needs with an explicit opt-in.

## How this differs from related issue/PR

This is not a target-scope or sandboxing change. It is separate from the open WebClient target-scope work in #219 and the broader subprocess sandboxing work in #210.

This PR focuses only on artifact/log secret handling and an operator-controlled debugging override.

## Attack flow

This is defensive hardening rather than a standalone vulnerability claim:

1. RAPTOR scans an authorized target.
2. Target URLs or responses contain credentials in URL/query/header-like strings.
3. RAPTOR records scan artifacts or logs.
4. Those artifacts may be shared externally for triage.
5. Default redaction reduces accidental credential disclosure.
6. Operators can explicitly reveal secrets when exact local debugging data is required.

## Affected code

| File | Purpose |
| --- | --- |
| `core/security/redaction.py` | Shared secret redaction helper |
| `core/config.py` | Central `RAPTOR_REVEAL_SECRETS` setting |
| `packages/web/client.py` | Redacts request-history URLs by default |
| `packages/web/fuzzer.py` | Redacts finding URLs by default |
| `packages/web/scanner.py` | Adds `--reveal-secrets` and threads the setting through |
| `packages/llm_analysis/llm/client.py` | Lets existing log sanitizer honor the central reveal setting |

## Root cause

- Web artifacts treated full URLs as harmless evidence even though scanner targets often contain credentials.
- Redaction behavior was not centralized across RAPTOR components.
- There was no explicit operator-level choice between share-safe artifacts and exact local-debug artifacts.

## CVSS assessment

No CVSS score is assigned. This PR is a security hardening and operator-safety improvement for scanner artifacts, not a claim of an independently exploitable vulnerability in RAPTOR.

## Safe reproduction steps

On current `main`, run code paths that log or store URLs containing sensitive parameters such as:

```text
https://example.test/path?api_key=<value>&debug=true
https://example.test/search?access_token=<value>
https://user:password@example.test/path
```

Observe that web request history and findings can retain the exact sensitive values.

## Expected vulnerable behavior

Before this PR, web scanner artifacts could preserve full credential-bearing URLs. After this PR, those values are replaced with `[REDACTED]` unless reveal mode is explicitly enabled.

## Changes in this PR

- Adds shared `redact_secrets()` helper.
- Redacts sensitive query parameters while preserving non-sensitive context such as host, path, and non-secret query params.
- Redacts URL userinfo passwords.
- Redacts Bearer and Basic authorization values.
- Adds `RaptorConfig.reveal_secrets_enabled()`.
- Adds `RAPTOR_REVEAL_SECRETS` environment toggle.
- Adds `/web --reveal-secrets` CLI flag.
- Threads reveal configuration through `WebScanner` → `WebClient` → `WebFuzzer`.
- Adds regression tests for default redaction and explicit reveal behavior.

## Files changed

| Category | Files |
| --- | --- |
| Shared helper | `core/security/redaction.py` |
| Config | `core/config.py` |
| Web scanner | `packages/web/client.py`, `packages/web/fuzzer.py`, `packages/web/scanner.py` |
| LLM logs | `packages/llm_analysis/llm/client.py` |
| Tests | `core/security/tests/test_redaction.py`, `packages/web/tests/test_secret_redaction.py`, `packages/llm_analysis/tests/test_llm_callbacks.py` |

## Maintainer impact

- Default behavior becomes safer for shared artifacts.
- Operators who need exact values can opt in explicitly.
- Existing callers of `WebClient` and `WebScanner` remain compatible because the new `reveal_secrets` parameter defaults to `None`.
- The redaction helper preserves useful evidence context instead of replacing whole URLs.

## Fix rationale

Redaction belongs at artifact/log boundaries, not in scanner request generation. RAPTOR should still send the operator-authorized request as configured, but stored or logged evidence should be share-safe by default.

The explicit reveal setting keeps troubleshooting possible without making exact credential retention the default.

## Type of change

- [x] Security hardening
- [x] Feature / operator configuration
- [x] Tests
- [ ] Documentation-only change
- [ ] Breaking change

## Test plan

Passed locally on Python 3.11.15:

```bash
python -m ruff check core/config.py core/security/redaction.py core/security/tests/test_redaction.py packages/web/client.py packages/web/fuzzer.py packages/web/scanner.py packages/web/tests/test_secret_redaction.py packages/llm_analysis/llm/client.py packages/llm_analysis/tests/test_llm_callbacks.py
python -m compileall -q core packages *.py
python -m pytest core/security/tests/test_redaction.py packages/web/tests/test_secret_redaction.py packages/llm_analysis/tests/test_llm_callbacks.py::TestSanitizeLogMessage -q
python -m pytest core/security/tests packages/web/tests packages/llm_analysis/tests/test_llm_callbacks.py -q
git diff --check
```

Results:

```text
ruff: All checks passed
compileall: passed
targeted redaction tests: 59 passed
related core/web/LLM callback tests: 195 passed
diff check: passed
```

I also ran the broader suite:

```bash
python -m pytest core packages -q
```

Result on local macOS/Python 3.11.15:

```text
1942 passed, 122 skipped, 6 failed
```

The failures appear unrelated to this PR and match existing platform/environment-sensitive areas, including tree-sitter C parameter extraction, Python-version syntax in exploitability validation schemas, `/private/var` path resolution, Solaris cross-platform checks, and glibc confidence detection on Darwin.

An added-line scan for obvious secret assignments and dangerous sink additions found no hits.

## Disclosure notes

This PR is intentionally scoped to scanner artifact/log secret handling. It does not claim RAPTOR has a standalone remotely exploitable vulnerability, and it does not change scanner request behavior or target-scope enforcement.


packages/llm_analysis/llm/client.py
```diff
@@ -51,6 +51,9 @@ def _sanitize_log_message(msg: str) -> str:
     Searchable tags: #SECURITY #API_KEY_PROTECTION #LOG_SANITIZATION
     Related: Cursor Bot Bug #2, PR #32, defense-in-depth best practice
     """
+    if RaptorConfig.reveal_secrets_enabled():
```
grokjc: This early return bypasses all defense-in-depth API key sanitisation — Anthropic keys, OpenAI keys, Google keys, GitHub tokens, AWS access keys, PEM private key blocks, and structured key=value patterns. That sanitisation exists to protect RAPTOR's own operator credentials from leaking in error messages.
                                                                                                                   
RAPTOR_REVEAL_SECRETS is designed for target credentials in web scan artifacts — a different concern. An operator debugging a web scan shouldn't have to also expose their sk-ant-* keys in log output.

Please remove this early return. The reveal_secrets toggle already works correctly in the web client and fuzzer paths via redact_secrets() — that's where it belongs.

packages/llm_analysis/tests/test_llm_callbacks.py
```diff
@@ -379,6 +379,13 @@ def test_redacts_truncated_private_key_block(self):
         assert "n" * 64 not in result
         assert "[REDACTED-PRIVATE-KEY]" in result
 
+    def test_can_preserve_secrets_for_operator_debugging(self, monkeypatch):
```
grokjc: This test validates the _sanitize_log_message bypass that should be removed (see comment on client.py). Please drop this test — operator API keys should always be redacted regardless of RAPTOR_REVEAL_SECRETS.

core/config.py
```diff
@@ -107,6 +107,7 @@ class RaptorConfig:
     ENV_OUT_DIR = "RAPTOR_OUT_DIR"
     ENV_JOB_ID = "RAPTOR_JOB_ID"
     ENV_LLM_CMD = "RAPTOR_LLM_CMD"
+    ENV_REVEAL_SECRETS = "RAPTOR_REVEAL_SECRETS"
```
grokjc: Project convention is to avoid new RAPTOR_* env vars where CLI flags or config objects suffice. The web scanner already has --reveal-secrets on the CLI, which threads through to WebClient and WebFuzzer cleanly.                                                                                                                                                                                                         
   
If the env var stays as a convenience default for the web path, consider renaming to RAPTOR_REVEAL_TARGET_SECRETS to make clear this is about target-site credentials in scan artifacts, not operator credentials.

packages/web/scanner.py
```diff
@@ -156,7 +162,7 @@ def main():
 
     # Run scan
     verify_ssl = not args.insecure
-    scanner = WebScanner(args.url, llm, out_dir, verify_ssl=verify_ssl)
+    scanner = WebScanner(args.url, llm, out_dir, verify_ssl=verify_ssl, reveal_secrets=args.reveal_secrets or None)
```
grokjc: False or None evaluating to None works but is a Python gotcha that'll trip up readers. Clearer as:
                                                                                                                                                                                                                   
  ```reveal_secrets=True if args.reveal_secrets else None```

packages/web/client.py
```diff
@@ -51,16 +55,17 @@ def _rate_limit_wait(self) -> None:
     def _log_request(self, method: str, url: str, response: requests.Response,
                      duration: float) -> None:
         """Log request details."""
+        log_url = redact_secrets(url, reveal_secrets=self.reveal_secrets)
         self.request_history.append({
             'method': method,
-            'url': url,
+            'url': log_url,
```
grokjc, line 64: request_history now holds redacted URLs rather than originals. That's fine for statistics and reporting (its current consumers), but it means the history can't be used to replay requests if that's ever needed later. If you wanted to preserve both, you could store the original as a separate field and redact only the logged/displayed URL — but this is a design choice, not a bug, and the current approach matches the  PR's intent.

grokjc (changes requested):
Good work on the core redaction module — redaction.py is clean, the _SECRET_QUERY_KEYS set is well-scoped, and the web client/fuzzer integration threads reveal_secrets through correctly. Tests are comprehensive.

One security concern needs fixing before merge: the _sanitize_log_message bypass conflates "show me target credentials in scan artifacts" with "disable all defense-in-depth API key sanitisation." See inline comments.

packages/llm_analysis/llm/client.py
Hinotoi-agent: Removed the `RaptorConfig.reveal_secrets_enabled()` early return in `_sanitize_log_message()`, so LLM/operator API-key sanitization now always runs regardless of the web artifact reveal setting. Commit: cdcec6e

packages/llm_analysis/tests/test_llm_callbacks.py
Hinotoi-agent: Dropped the test that validated the removed LLM sanitizer bypass. Commit: cdcec6e

core/config.py
Hinotoi-agent: Renamed the convenience environment opt-in to `RAPTOR_REVEAL_TARGET_SECRETS` so the scope is target/web artifacts, not operator credentials. Commit: cdcec6e

packages/web/scanner.py
Hinotoi-agent: Updated this to the clearer `True if args.reveal_secrets else None` form and wrapped the call for readability. Commit: cdcec6e

grokjc:
The follow-up commit addresses the LLM bypass nicely. One thing: --reveal-secrets currently only exists on the /web scanner, but WebClient falls back to RAPTOR_REVEAL_TARGET_SECRETS when nothing is passed explicitly. Could we close that gap by wiring the flag through the other entry points that construct a WebClient (e.g. /agentic) rather than relying on the env var as a catch-all? Then the env var can go away and every invocation is explicit.

Happy to defer that to a follow-up if you'd rather land this now and address it separately.

Hinotoi-agent:
Thanks — I pushed `fd7aaedf71174ea7b743540aa914c37ce6675521` with the explicit-reveal fix plus one extra hardening/check pass.

What changed for the requested gap:
- Removed the `RAPTOR_REVEAL_TARGET_SECRETS` fallback entirely.
- Removed `RaptorConfig.reveal_secrets_enabled()` and the config constant/tests for that env var.
- Made `WebClient(..., reveal_secrets=False)` and `WebScanner(..., reveal_secrets=False)` explicit by default.
- Kept `/web --reveal-secrets` as the explicit operator opt-in and threaded that boolean directly into `WebScanner`/`WebClient`.
- Added regression coverage that the legacy env var no longer enables reveal behavior.
- Added scanner constructor coverage proving `reveal_secrets=True/False` is passed to `WebClient`.

Additional hardening added after rechecking the PR:
- Centralized `WebClient` display/log redaction through `_redact_for_logging()`.
- Redacted timeout and request-exception log messages, so URL secrets do not leak on failure paths.
- Redacted the `WebFuzzer.fuzz_parameter()` start log while preserving non-secret URL context.
- Added tests for timeout logs, exception logs, and fuzzer start logs.

Validation run locally with `uv run` / Python 3.11:

```text
uv run python -m compileall -q core packages *.py
uv run pytest core/security/tests/test_redaction.py packages/web/tests/test_secret_redaction.py packages/web/tests/test_scanner_none_llm.py packages/llm_analysis/tests/test_llm_callbacks.py packages/web -q
# 72 passed

git diff --check

source scan:
reveal_secrets_enabled: []
ENV_REVEAL_TARGET_SECRETS: []
RAPTOR_REVEAL_TARGET_SECRETS: []
```

GitHub Actions are also green on the latest head.

grokjc:
Nicely done. Env var removed, explicit default, regression test for the old var. Good to merge. Thanks!

PR #225, opened 2026-04-25
SMT: Stage E LLM-callable path-feasibility helper + parser hardening

grokjc:
Adds raptor-smt-validate-path, a libexec entry point Stage E invokes when the LLM has read source for a finding and wants Z3 to verify whether the visible guards actually block exploitation. The script takes condition strings as positional args (or a JSON array via --stdin), accepts profile by stdint-style name (uint32, int32, uint64, ...), and emits a JSON verdict to stdout. No Python visible to the LLM.

Underlying: a thin Python wrapper (validate_path) over the existing check_path_feasibility, with profile-name resolution, condition normalisation, and JSON-serialisable output. Stage E's stage-e-feasibility.md skill prompt gains an [E-3c] section documenting when to invoke, profile choice guidance, output interpretation, and the trivial-witness gotcha (Z3 picks the smallest satisfying assignment by default — add a lower-bound condition to force the dangerous range).

Adversarial testing of the wrapper surfaced four classes of latent parser bug, all silent encoding mismatches that produced wrong verdicts. The path-condition parser now rejects them cleanly to the unknown bucket:

  - Characters silently dropped by the tokeniser (~, ^, /, %, ?:, =) used to vanish from the token stream and produce mis-encoded expressions; now caught by a full-input-consumed sanity check.
  - Literals exceeding the profile's bitvector width (e.g. 0x100 at uint8) silently wrapped to zero in z3.BitVecVal, producing fake tautologies; now rejected.
  - Leading-zero decimal literals (01234 — octal in C) were interpreted as base-10, or crashed with a Python ValueError when encountered in the bitmask path; now rejected uniformly.
  - The bitmask path (flags & MASK == VAL) had its own literal extraction that bypassed atom-level validation; literal handling is now hoisted into a shared _parse_literal_value helper used by both paths.

Negated conditions in the unsatisfied / unsat-core display now read "NOT (ptr != NULL)" rather than "ptr != NULL" — the previous form showed the original text even though the solver asserted the negation, confusing readers.

Validator docstring's "Known limitations" section rewritten to accurately describe what's rejected and why.

36 wrapper tests + 18 regression tests for the parser hardening. Full repo suite: 2084 passed.

### DIFF
--- a/.claude/skills/exploitability-validation/stage-e-feasibility.md
+++ b/.claude/skills/exploitability-validation/stage-e-feasibility.md
@@ -147,6 +147,45 @@ arithmetic), CWE-193 (off-by-one), CWE-476 (null deref).
 
 The heuristic `note` field is always preserved alongside `smt_feasibility` — compare both when they conflict. When `smt_feasibility.feasible` is `false`, prefer the SMT verdict for the `chain_breaks` entry.
 
+### [E-3c] Optional: SMT path-condition verification
+
+When you've read the source for a finding and formed a hypothesis about whether the visible guards actually block exploitation — e.g. a CWE-190 where `count < MAX_RECORDS` looks protective but `count * sizeof(record)` might wrap on 32-bit unsigned arithmetic — you can ask Z3 to verify whether the path conditions are jointly satisfiable:
+
+```bash
+libexec/raptor-smt-validate-path --profile uint32 \
+  'alloc_size == count * 16' \
+  'alloc_size < 0x8000' \
+  'count > 0x10000000' \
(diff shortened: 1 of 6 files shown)


grokjc:
@unprovable for visibility and/or comment

unprovable:
Oh this is outstanding! yeah, this tightens up the parser's logic magnificently. I'll sync everything up and see if there's more to make the edge case sets as thin as possible.

grokjc:
Thanks - I'm going to merge this and we can iterate as needed.

PR #234, opened 2026-04-26
Harden devcontainer: pin versions, replace hand-rolled deps, run as vscode

grokjc:
Pins semgrep and @anthropic-ai/claude-code so rebuilds are reproducible. Replaces a hand-rolled Playwright apt list (32 packages, drifting out of sync with upstream Chromium) with `playwright install --with-deps chromium`. Captures the nodesource setup script before executing it so curl failures aren't silently swallowed by `curl | bash`, and bumps Node 20→22 since 20 hit EOL April 2026. Sets USER vscode so the runtime container isn't root-by-default; build steps still run as root by inheritance, and vscode has passwordless sudo for tools that genuinely need it.

Also dedupes a `rm -f /etc/apt/sources.list.d/yarn.list` line that landed twice when #194 stacked on top of #226.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -15,8 +15,6 @@ RUN rm -f /etc/apt/sources.list.d/yarn.list
 # Organized by component that needs them
 # ============================================================================
 
-RUN rm -f /etc/apt/sources.list.d/yarn.list
-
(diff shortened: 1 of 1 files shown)


danielcuthbert (approved):
Alles gud, we will need to potentially track said versions but this is suitable for now

PR #237, opened 2026-04-26
bridge: function-level enrichment, normalisation, and log sanitisation

grokjc:
Reworks core/orchestration/understand_bridge.enrich_checklist so the priority markers it writes into checklist.json reflect the granularity the context map actually provides:

  - Match on (file, function name) when the context map names a function, falling back to (file, None) only when name is absent. Files that contain both an entry point and a sink now retain both signals via a "+"-concatenated priority_reason ("entry_point+sink") rather than the second one overwriting the first.
  - Clear prior priority markers before re-applying so re-runs reflect the current context map instead of accumulating stale markers.
  - Resolve unchecked-flow IDs to full entry/sink details under checklist["priority_targets"] so Stage B has the resolved record in front of it, not a bare ID lookup.

Adds normalize_context_map() — a mechanical pre-pass that strips leading "./", converts absolute paths under target_path to relative, backfills missing function names from line ranges (smallest-span for nested functions), warns on file/line references the checklist disproves, and validates unchecked_flow ID cross-references.  Exposed to the standalone /understand --map flow via a new MAP-5 step in the skill markdown calling the libexec/raptor-normalize-context-map shim, so non-validate runs benefit from the same fixups the bridge applies.

Adopts core/security/log_sanitisation.escape_nonprintable (shipped on main with sandbox #210) for every operator-facing logger call in this file that interpolates an LLM-emitted string — file paths, entry/sink ID references, trace file paths, stale-file lists.  Without this, a malicious target repo could inject ANSI escapes, BIDI overrides, or BEL into raptor's own warning output.  Two regression tests lock the contract in.

Also fixes a CI flake in test_happy_path_runs_lifecycle_and_writes_ selection_file — the assertion that finding IDs do not appear in the post-pass prompt used 1-2 char IDs ("f1", "f2", "f3"), which can collide with the 8-char random suffix tempfile.mkdtemp generates. Renamed to FNDA-EXPLOITABLE / FNDB-HIGHCONF / FNDC-IGNORED so the hyphen + uppercase rules them out of mkdtemp's character set.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -9,7 +9,7 @@
       "Bash(libexec/raptor-agentic *)",
       "Bash(libexec/raptor-build-checklist *)",
       "Bash(libexec/raptor-coverage-summary *)",
-      "Bash(libexec/raptor-validation-helper *)",
+      "Bash(libexec/raptor-normalize-context-map *)",
       "Bash(libexec/raptor-project-manager *)",
       "Bash(libexec/raptor-render-diagrams *)",
       "Bash(libexec/raptor-run-feasibility *)",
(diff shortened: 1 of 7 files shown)


danielcuthbert (approved):
Solid throughout. love the defensive approach we have taken with log sanitisation ;)

PR #239, opened 2026-04-27
SMT Parser; Set Reasons for better feedback to agent

unprovable:
Split the `unknown` bucket by reason in SMT Parser. The largest loss of coverage from the parsers are things sinking into `unknown` without any feedback to the agent. If we provide this feedback, then we should be able to do much better parsing and thin out the pathways to `unknown` to things that we generally have no clue about. 

To do this, we make `unknown: List[str]` --> `unknown: List[UnknownReason]` carrying `(text, kind, hint)` as follows:
  - `LEX_ERROR`, 
  - `UNSUPPORTED_OPERATOR`, 
  - `PARENS_NON_CALL`, 
  - `MIXED_PRECEDENCE`, 
  - `LITERAL_OUT_OF_RANGE`, 
  - `UNKNOWN_REGISTER`, 
  - `SOLVER_TIMEOUT`,
  - `SOLVER_UNKNOWN_FRAGMENT`

  This lets:
  - The LLM/extractor retry only on the textually-fixable cases (massive coverage gain at the source).
  - Tests assert on why something was unknown, not just that it was.
  - Reports show a researcher exactly what to rephrase.

Tests are also included for checking and completeness - thanks be to Claude for the insight on these.

### DIFF
--- a/.claude/skills/exploitability-validation/stage-e-feasibility.md
+++ b/.claude/skills/exploitability-validation/stage-e-feasibility.md
@@ -184,7 +184,34 @@ Output is JSON to stdout:
 
 **Profile mismatch is a silent failure mode.** If you analyse `unsigned int` arithmetic under `uint64`, Z3 misses 32-bit wraparound and the `feasible: false` verdict you get is wrong (the path IS reachable via 32-bit wrap). Match the profile to the dominant C type's width. The reasoning string spells out what was used (`"feasible (32-bit unsigned): ..."`) — if it doesn't match the vuln's real C type, treat the verdict as suspect.
 
-**Iterate on `unknown`.** If the verdict comes back with a non-empty `unknown` bucket, the parser couldn't encode those forms. Re-read the source, re-express the rejected forms in parser-compatible shape (introduce a helper variable, drop the cast, split a complex condition, replace a function call with the predicate it computes), and re-invoke. One refinement pass usually closes the gap. If a second pass still has unknowns, the conditions are genuinely outside the bitvector fragment — stop iterating and fall back to qualitative reasoning rather than chasing further.
+**Iterate on `unknown` using `unknown_reasons`.** When the verdict comes back with non-empty `unknown`, the parallel `unknown_reasons` array tells you *exactly why* each condition was dropped — read those rather than guessing. Each entry has the shape:
+
+```json
+{
+  "text": "01234 < limit",
+  "kind": "literal_ambiguous",
+  "detail": "leading-zero decimal is ambiguous with C octal",
+  "hint": "rewrite as hex (0x...) or strip the leading zero"
+}
+```
+
+The `kind` values you'll see and the surgical fix for each:
+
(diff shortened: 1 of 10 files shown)


unprovable:
Two other PR's are built on this (but are separate from each other): #241 , which adds some find/replace logic for English-isms ("greater than zero" becomes "> 0", etc) and #244 , which adds a validator free-variable fallback whilst parens are being juggled around. This should allow the solver to run in more scenarios.

grokjc:
Thanks — Rejection/RejectionKind is a good improvement. Three things need fixing before merge:
  
1. The branch was forked before #225 landed and never rebased — please rebase onto current main. That will surface three regressions from #225 you'll need to keep:
  - _parse_literal_value removed from smt_path_validator.py (silent width-wrap reintroduced — x == 0x100 at uint8 → 0)
  - Silent-character-drop guard removed from _parse_expr (~mask == 0xFF mis-parses as mask == 0xFF)
  - The [E-3c] section + iterative-refinement paragraph dropped from stage-e-feasibility.md
  
  All three need to come back, ideally producing structured Rejections instead of None (LITERAL_OUT_OF_RANGE, LITERAL_AMBIGUOUS, plus a new LEX_DROPPED_CHARS kind).
  
2. Helper duplication. _propagate and _classify_solver_unknown are defined verbatim in both smt_path_validator.py and smt_onegadget.py. Move them to core/smt_solver/rejection.py so future encoders share one source. Same goes for _parse_literal_value once it's restored.
  
3. Smaller items:                                         
  - assert not isinstance(reg_expr, Rejection) in check_onegadget is fragile — asserts disappear under python -O. Use raise RuntimeError.
  - test_trailing_tokens_rejection actually asserts UNRECOGNIZED_FORM, not TRAILING_TOKENS. Misnamed; please rename and add a real TRAILING_TOKENS test (e.g. "a + b c == 0").                                     
  - The second LEX_EMPTY branch in _parse_constraint (if not or_exprs:) is unreachable — re.split on a non-empty string always yields ≥1 element.                             
  - No test for _classify_solver_unknown. Mock solver.reason_unknown() returning "timeout" / "canceled" / etc.                                                                                                     
  - Stage E skill should teach the LLM to consume unknown_reasons[].kind and .hint for targeted retry.

unprovable:
1. - rebased to curent main with #225 _et al._ included. Also added rejection reasons as much as I can foresee. Happy to add mroe fixes. 
2. Moved helpers to `rejection.py` which is now working well. 

Working on 3 right now...

unprovable:
Ok not quite 1-per-issue but I think these are now resolved in these commits, @grokjc . 🙏

grokjc:
All five items + the bonus are addressed cleanly. Helpers consolidated, assert→raise, real TRAILING_TOKENS test, unreachable branch gone, skill paragraph updated with the JSON shape, and a proper TestClassifySolverUnknown covering timeout / canceled (US/UK) / uppercase / empty / None / missing-attr / exception paths. 151 SMT-related tests pass locally. LGTM — ready to merge.

Thanks!

PR #241, opened 2026-04-27
SMT Parser - Add English-aliased Pre-canonicalization step

unprovable:
Building on coverage widening in #239 - this invokes rather simple regex canonicalization of LLM output.

Adds a small shared pre-pass (`core/smt_solver/canonicalise.py`) that rewrites common english operator phrases to their symbolic equivalents *before* the SMT encoder parsers see the input. Wired into both `smt_path_validator._parse_condition` (CodeQL dataflow) and `smt_onegadget._parse_atom` (exploit feasibility). LLM output frequently uses phrases like *"size is greater than 0"*, *"count is at least 16"*, *"ptr is non-null"* — these previously fell through to the `unknown` list as `UNRECOGNIZED_FORM`. They now parse cleanly.                                                                                                                              
   
  ## What's rewritten                                                                                                                   
                                                                                   
  | English phrase                                  | Canonical form |
  | ----------------------------------------------- | -------------- |
  | `is greater than [or equal to]`                 | `>` / `>=`     |
  | `is less than [or equal to]`                    | `<` / `<=`     |
  | `is at least` / `is at most`                    | `>=` / `<=`    |
  | `equals` / `is equal to`                        | `==`           |
  | `is not equal to` / `does not equal`            | `!=`           |
  | `is non-zero` / `is nonzero` / `is non zero`    | `!= 0`         |                                                                  
  | `is non-null` / `is nonnull` / `is non null`    | `!= NULL`      |
                                                                                                                                        
  ## Design notes                                                                  
                                                                                                                                        
  - **Additive, not invasive.** Existing dedicated parser paths (`is NULL`, `is writable` in one-gadget) are deliberately *not*
  canonicalised — those keep their domain-specific Z3 encodings and rejection messages.                                                 
  - **Order matters.** Longer phrases match first (`is greater than or equal to` before `is greater than`).
  - **Identifier-safe.** `\b` word boundaries keep `equalsValue == 1` and `is_greater_than_zero` from being mangled.                    
  - **Idempotent.** Already-symbolic input passes through unchanged.                                                                    
  - **Original text preserved for diagnostics.** The `Rejection.text` field still carries the user's original input so callers can match
   failures back to the source.                                                                                                         
                                                                                                                                        
  ## Test plan                              
                                                                                                                                        
  - [x] 23 unit tests covering each rewrite, idempotence, identifier safety, case-insensitivity, whitespace collapse                    
  - [x] Smoke tests through both parsers (`check_path_feasibility`, `check_onegadget`) confirm canonicalised input reaches              
  `feasible=True` with empty `unknown`                                                                                                  
  - [x] Full SMT suite green: 190/190 (existing 116 + 23 canonicalise + 51 rejection tests from earlier branch)
                                            
  ## Files                   
                                                                                                                                        
  core/smt_solver/init.py 
  core/smt_solver/canonicalise.py
  core/smt_solver/tests/test_canonicalise.py
  packages/codeql/smt_path_validator.py 
  packages/exploit_feasibility/smt_onegadget.py      
                                      
  > **Note:** this PR is stacked on #239, so the PR diff against `main` will include the rejection-reasons commits.

### DIFF
--- a/core/smt_solver/__init__.py
+++ b/core/smt_solver/__init__.py
@@ -22,6 +22,7 @@ predicates, one-gadget constraints, ...) live in their respective
 
 from .availability import z3, z3_available
 from .bitvec import ge, gt, le, lt, mk_val, mk_var
+from .canonicalise import canonicalise
 from .config import (
     BVProfile,
     BV_AARCH64,
@@ -90,4 +91,6 @@ __all__ = [
     "propagate",
     "parse_literal_value",
     "classify_solver_unknown",
+    # English-aliased pre-canonicalisation
+    "canonicalise",
 ]
--- a/core/smt_solver/canonicalise.py
+++ b/core/smt_solver/canonicalise.py
@@ -1,0 +1,61 @@
+"""English-aliased pre-canonicalisation for SMT encoder parsers.
+
+LLM output frequently uses english operator phrases — "is greater than",
+"equals", "is at least" — instead of the canonical symbolic forms.
+This module applies a small ordered set of regex rewrites *before* the
+parser sees the input, mapping common english forms to ``>``, ``<``,
+``>=``, ``<=``, ``==``, ``!=`` (with NULL / 0 specialisations for
+non-zero / non-null phrasings).
+
+Design intent:
+- Rewrites are *additive*: phrases the per-parser grammar already
+  recognises (``is NULL``, ``is writable`` in one-gadget) are NOT
+  touched here, so existing parse paths keep their dedicated
+  rejection messages.
+- Order matters — longer, more-specific phrases are tried before
+  shorter ones (``is greater than or equal to`` before ``is greater
+  than``).
+- Word boundaries (``\\b``) keep rewrites from firing inside
+  identifiers (``equalsValue`` must NOT become ``==Value``).
+
+Used by:
+  packages/codeql/smt_path_validator.py :: _parse_condition
+  packages/exploit_feasibility/smt_onegadget.py :: _parse_atom
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Tuple
+
+# (pattern, replacement) pairs.  Replacements include surrounding spaces
+# because the english phrases don't always sit next to whitespace; the
+# trailing collapse-whitespace pass tidies up.
+_REWRITES: Tuple[Tuple[re.Pattern[str], str], ...] = (
+    (re.compile(r'\bis\s+greater\s+than\s+or\s+equal\s+to\b', re.IGNORECASE), ' >= '),
+    (re.compile(r'\bis\s+less\s+than\s+or\s+equal\s+to\b',    re.IGNORECASE), ' <= '),
+    (re.compile(r'\bis\s+at\s+least\b',                       re.IGNORECASE), ' >= '),
+    (re.compile(r'\bis\s+at\s+most\b',                        re.IGNORECASE), ' <= '),
+    (re.compile(r'\bis\s+greater\s+than\b',                   re.IGNORECASE), ' > '),
+    (re.compile(r'\bis\s+less\s+than\b',                      re.IGNORECASE), ' < '),
+    (re.compile(r'\bis\s+not\s+equal\s+to\b',                 re.IGNORECASE), ' != '),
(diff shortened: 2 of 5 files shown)


grokjc:
Nice — pre-canonicalising English phrases is a real win for LLM-produced conditions.
                                                                                                                                                                                                                   
Rebase first — this branch carries the same #225 regressions as #239. Once #239 is fixed, rebase onto it.
                                                                                                                                                                                                                   
Coverage gaps to consider:                                
  - Positive forms missing: is zero and is null don't rewrite, but is non-zero and is non-null do. Asymmetric. Path_validator has no native is NULL grammar so x is null will fail there (one_gadget has it natively, so works there). Either rewrite the positive forms too, or document that positive forms must be symbolic.                                                                                              
  - Common synonyms not covered: exceeds (→ >), below (→ <), does not exceed (→ <=), up to (→ <=). Easy to extend; not blocking.
                                                                                                                                                                                                                   
Otherwise sound — order is correct (long forms first), \b boundaries prevent false matches in identifiers, idempotent on symbolic input.

unprovable:
This should now be rebased with changed made to #239

unprovable:
confilcts should now be resolved.

unprovable:
> Coverage gaps to consider:
>
>Positive forms missing: is zero and is null don't rewrite, but is non-zero and is non-null do. Asymmetric. Path_validator has no native is NULL grammar so x is null will fail there (one_gadget has it natively, so works there). Either rewrite the positive forms too, or document that positive forms must be symbolic.
>Common synonyms not covered: exceeds (→ >), below (→ <), does not exceed (→ <=), up to (→ <=). Easy to extend; not blocking.

Ok, these should now be at least slightly better covered: 

 **Summary of changes**                                                                                                                
                                                                      
  | Phrase            | Maps to    | Rationale                                                                              |           
  | ----------------- | ---------- | -------------------------------------------------------------------------------------- |
  | `is null`         | `== NULL`  | closes asymmetry with `is non-null` (path validator had no native handling)            |           
  | `is zero`         | `== 0`     | closes asymmetry with `is non-zero`                                                    |           
  | `exceeds`         | `>`        | common LLM phrasing                                                                    |           
  | `below`           | `<`        | common LLM phrasing                                                                    |           
  | `does not exceed` | `<=`       | common LLM phrasing; ordered before `exceeds` so longer wins                           |
  | `up to`           | `<=`       | inclusive reading documented at the rewrite site                                       |           
                                                                      
  **Ordering safeguards**: every new phrase is placed so that longer/more-specific forms win against shorter prefixes. Comments at the rewrite list and on each multi-token phrase explain the constraint. 
                                                                                                                                        
  **Cross-encoder behaviour for `is null`**: one-gadget previously had its own dedicated `is NULL` suffix-form parser. Now the canonicalise pre-pass converts `... is null` → `... == NULL` before that parser runs, so the dedicated handler is bypassed — but the resulting Z3 expression (`lhs == 0`) is identical, so behaviour is unchanged. Documented in the module docstring. 

Probably ready for @grokjc's wizard-like whispers of wisdom.

grokjc:
Canonicalisation additions look great — both review items closed cleanly, including the ordering-collision tests. Two things before merge:
                                                                                                                                                                                                                   
  1. Duplicate _classify_solver_unknown in smt_path_validator.py — both an import (line 98 classify_solver_unknown as _classify_solver_unknown) AND a local def _classify_solver_unknown (line 521ish) exist; the local def shadows the import, partially undoing #239's helper consolidation for this file. Looks like a rebase artefact from merging the new core helper with the pre-existing local copy. Delete the local def — the import alone is what's wanted. (smt_onegadget.py is fine — import only, no shadowing.)
  2. Rebase onto current main — #239's squash-merge gave it a different SHA than your branch's copy, so the diff against main is larger than the actual change. A simple rebase will collapse it back to the canonicalise additions.                                                                                                                                                                                          
   
  Once those two are addressed, ready to merge.

unprovable:
Removed the dupe, and also rebased, then `git push --force-with-lease` so I think everything should line up, now.

grokjc:
Thanks! LGTM.

unprovable:
Fixed with the fix from #248

PR #243, opened 2026-04-27
core/http + core/json: HttpClient chokepoint + JsonCache

grokjc:
Adds core/http/ (HttpClient Protocol with UrllibClient + EgressClient backends) and core/json/cache.py (TTL'd JSON disk cache). EgressClient routes through core.sandbox.proxy with hostname allowlist enforcement, closing the exfil-to-attacker-host hole that direct urlopen leaves open.

urllib3-backed for connection pooling and to avoid the no_proxy bypass stdlib's ProxyHandler silently performs. Bounded backoff with wall- clock deadline; URL scheme + userinfo validation; size caps; conditional requests via NotModified; pid+tid tempfile suffixes for safe concurrent cache writers.

Validated against a 17-scenario adversarial stress harness and a real- network E2E through the full chokepoint chain.

### DIFF
--- a/core/http/__init__.py
+++ b/core/http/__init__.py
@@ -1,0 +1,270 @@
+"""HTTP client interface — single chokepoint for outbound HTTP from RAPTOR.
+
+Two backends ship today:
+
+  - :class:`~core.http.urllib_backend.UrllibClient` — urllib3-backed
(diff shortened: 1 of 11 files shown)


danielcuthbert (approved):
All solid, love this idea and yeah this will form the basis of good things moving forward

PR #244, opened 2026-04-27
SMT path validator: free-variable fallback for function-call subterms

unprovable:
Adds a free-variable fallback to the CodeQL path validator's parser. Function-call-shaped subterms (`strlen(input)`, `getpid()`, `min(a, b)`, ...) are now replaced with fresh `_anon_N` Z3 variables before parsing, so the rest of the condition can still drive feasibility analysis. Previously, any input containing parens fell through to the `unknown` list with `PARENS_NOT_SUPPORTED`; this was the single biggest source of dropped conditions in real LLM-extracted dataflow output.

  ## Behaviour changes

  | Input | Before | After |
  | --- | --- | --- |
  | `strlen(input) < 1024`                       | unknown (`PARENS_NOT_SUPPORTED`) | parses (`_anon_0 < 1024`) |
  | `0 < strlen(input)`                          | unknown                           | parses                     |                     
  | `getpid() != 0`                              | unknown                           | parses                     |                     
  | `min(a, b) > 0`                              | unknown                           | parses                     |                     
  | `f(g(x), h(y)) < 100`                        | unknown                           | parses (single `_anon_0`)  |                     
  | `compute(a + b * c, lookup(t, k)) > 0`       | unknown                           | parses                     |                     
  | `strlen(s) & 0xff == 0`                      | unknown                           | parses                     |                     
  | `(a + b) > 0`                                | unknown                           | unknown (non-call grouping — still unsupported) |
  | `strlen(x < 1024`                            | unknown                           | unknown (unbalanced parens still rejected)      |

Critically, the fallback **never inverts a verdict from infeasible to feasible based on missing information**. A path with a real contradiction plus a fallback-recovered call is still reported infeasible:                                                            

```python                                                 
  size > 100, size < 50, strlen(input) < 1024  →  feasible=False
```

## Design notes
                                                                                                                                        
  - Conservative semantics. Two textually-identical calls `(strlen(s) == strlen(s))` get distinct `_anon_N` placeholders. The parser doesn't know which calls are pure, so it picks the safe stance: never claim infeasibility from an elided subterm.
  - Nested calls collapse. `f(g(x), h(y))` becomes one `_anon_0` — the outer call drives the balanced-paren walk; the inner structure is discarded along with everything else inside the call.                                                                                 
  - Counter is path-scoped. The `_anon_N` allocator is seeded from existing `_anon_*` keys in the shared vars_ dict, so calls in different conditions of the same path don't accidentally collide on a shared free variable.                                                     
  - Unbalanced parens fall through. `strlen(x < 1024` doesn't match the balanced walk, so the existing PARENS_NOT_SUPPORTED path still flags it.                                                                                                                             
  - Hint message updated. When non-call grouping parens `((a + b) * c)` are rejected, the hint now points at the fallback so the LLM understands which paren shape is supported.                                                                                           
                                                            
**Test plan**

- 12 new tests in TestFreeVariableFallback: simple LHS/RHS calls, zero-arg, multi-arg, nested, two-call distinct vars, bitmask LHS, fallback-doesn't-mask-real-conflict, unbalanced-still-rejected, counter-progress-across-conditions
  - 4 existing tests updated to use non-call grouping parens (the original `validate(ptr, len) == 0` inputs now legitimately parse via the fallback)                                                                                                                         
  - packages/codeql/tests/test_smt_path_validator.py: 58/58 pass (was 46)
  - Full SMT sweep (path validator + one-gadget + core/smt_solver): 240/240 pass                                                        

**Files:**
- packages/codeql/smt_path_validator.py
- packages/codeql/tests/test_smt_path_validator.py 
                                                                                                                                        
Specifically:                                                                                                                         
  - packages/codeql/smt_path_validator.py                   
    - New `_substitute_calls(text, vars_, *, profile)` helper `(balanced-paren walk, fresh-var allocator)`.
    - `_parse_condition` now runs `_substitute_calls` before the parens check.                                                              
    - Updated module `docstring` (call-shape behaviour + new "Known limitations" entry on non-call grouping).                             
    - Updated `PARENS_NOT_SUPPORTED` rejection hint.                                                                                      
  - packages/codeql/tests/test_smt_path_validator.py                                                                                    
    - New TestFreeVariableFallback class (12 tests).                                                                                    
    - 4 existing tests updated to use grouping-paren inputs.  

**NOTE** - this is built on #239 as is #241, except this is distinct from #241 and does not require canonicalization.

### DIFF
--- a/.claude/skills/exploitability-validation/stage-e-feasibility.md
+++ b/.claude/skills/exploitability-validation/stage-e-feasibility.md
@@ -180,9 +180,38 @@ Output is JSON to stdout:
 
 **When to skip:** if `analyze_binary` returned `smt_available: false`, z3-solver isn't installed and this tool would just return `feasible: null` — skip the call and rely on your own reasoning. Also skip when the finding already carries `smt_feasibility` from the CodeQL pre-check (don't redo work).
 
-**Conditions the parser rejects** (these land in `unknown` and the verdict will lean toward `null`): function calls (`strlen(input)`), type casts (`(uint32_t)x`), struct/pointer access (`obj.field`, `s->len`), array indexing (`arr[0]`), pointer deref (`*p`), unary NOT (`~`), XOR (`^`), division (`/`), modulo (`%`), ternary (`?:`). Stick to bare variables, integer/hex literals, `NULL`, and the operators `+ - * | & << >>` plus relational `< <= > >= == !=`. If you have to express something more complex, simplify it first or omit it — a `null` verdict from a contaminated condition set is worse than no SMT call at all.
+**Conditions the parser rejects** (these land in `unknown` and the verdict will lean toward `null`): type casts (`(uint32_t)x`), struct/pointer access (`obj.field`, `s->len`), array indexing (`arr[0]`), pointer deref (`*p`), unary NOT (`~`), XOR (`^`), division (`/`), modulo (`%`), ternary (`?:`), and non-call grouping parens (`(a + b) * c`). Stick to bare variables, integer/hex literals, `NULL`, and the operators `+ - * | & << >>` plus relational `< <= > >= == !=`. If you have to express something more complex, simplify it first or omit it — a `null` verdict from a contaminated condition set is worse than no SMT call at all.
+
+**Function calls are recovered, not rejected** — see the soundness warning below.
 
 **Profile mismatch is a silent failure mode.** If you analyse `unsigned int` arithmetic under `uint64`, Z3 misses 32-bit wraparound and the `feasible: false` verdict you get is wrong (the path IS reachable via 32-bit wrap). Match the profile to the dominant C type's width. The reasoning string spells out what was used (`"feasible (32-bit unsigned): ..."`) — if it doesn't match the vuln's real C type, treat the verdict as suspect.
+
+**Free-variable fallback over-approximates in the SAT direction.** Function-call subterms (`strlen(input)`, `getpid()`, `min(a, b)`, ...) parse — the parser substitutes each balanced `<ident>(...)` with a fresh `_anon_N` Z3 free variable so the rest of the condition still drives feasibility. The catch: two textually-identical calls become *independent* free variables, because the parser doesn't know which calls are pure. So:
+
+```
+strlen(s) > 100 AND strlen(s) < 50
+```
+
+returns `feasible: true` with a model like `{_anon_0: 101, _anon_1: 49}` — even though for any pure `strlen` the path is unsat (`strlen(s)` can't be both >100 and <50). The model is mathematically correct for the *encoded* problem but not for the real C path.
+
+**The asymmetry that matters for validation:**
+
+- `feasible: false` with `_anon_N` in the unsat core is **still sound** — adding free variables only *relaxes* constraints, so an unsat verdict can't become sat by removing them. Trust it.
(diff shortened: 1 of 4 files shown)


grokjc:
The free-variable fallback is a big coverage win — recovering strlen(input) < 1024 instead of dropping it. Conservative semantics (distinct placeholders for identical calls) is the right call.
                                                                                                                                                                                                                   
  Rebase first — same #225-regression issue as #239 / #241.
                                                                                                                                                                                                                   
  One soundness concern needs surfacing in the skill prompt: identical calls become independent free variables, so strlen(s) > 100 AND strlen(s) < 50 will return feasible: true even though (with strlen treated  as a pure function) it's unsat. That's an over-approximation in the sat direction — feasible: true on paths containing function calls is necessarily a hint, not proof. Please update stage-e-feasibility.md [E-3c] to teach the LLM: when the model contains _anon_N variables, the feasibility verdict is over-approximate; the actual path may be infeasible if the function calls are pure and constrained.               
                                                            
  Smaller items:                                                                                                                                                                                                   
  - Placeholder counter sum(1 for k in vars_ if k.startswith('_anon_')) is brittle if anyone ever deletes from vars_. Use max(...) + 1 over existing indices.
  - Worth a test for the canonicalise + substitute_calls interaction (when both #241 and this land), e.g. "strlen(input) is greater than 0" should parse end-to-end.                                               
  - Confirm tests cover the unbalanced-parens fallthrough ("strlen(x").

unprovable:
Ok, will get on the conflicts.

unprovable:
Working on conflicts

unprovable:
Ok added a note to `.claude/skills/exploitability-validation/stage-e-feasibility.md` to try and stymie the soundness concern raised by @grokjc . There's a more graceful fix I'm sure in the future.

unprovable:
> Use max(...) + 1 over existing indices

Done.

unprovable:
> Confirm tests cover the unbalanced-parens fallthrough ("strlen(x").

Done - well, I extended the tests in `packages/codeql/tests/test_smt_path_validator.py:775+` to cover the following: 

```
    @pytest.mark.parametrize("expr", [
        "strlen(input < 1024",  # missing close, mid-expression
        "strlen(x",             # missing close, at EOI
        "strlen(x))",           # extra close from the right
        "f(g(x)",               # nested with one missing close
        "strlen x)",            # orphan close, no preceding (
    ])
```

> Worth a test for the canonicalise + substitute_calls interaction (when both https://github.com/gadievron/raptor/pull/241 and this land), e.g. "strlen(input) is greater than 0" should parse end-to-end.

Will make a note for future! I think these will be very illustrative tests. Ready for @grokjc's eyes.

grokjc:
One thing left: this branch is rebased on 1c368b5 (the #239 squash-merge) but hasn't picked up #248 (timeout caps in real_tests.sh) yet. As a result the diff shows #248's timeout 10s wraps being removed, which would revert that PR if merged. A quick rebase onto current main head will resolve.                                                                                                                             
                                                                                                                                                                                                                   
Once rebased, ready to merge.

unprovable:
Rebased to main hash - a la #241's review comment from @grokjc .

grokjc:
thanks, I think I was wrong anyway 😄 (but it didn't hurt)

unprovable:
Aha! No worries - It's never goign to hurt . Cheers! M.

unprovable:
Wroking on that conflict.

grokjc:
```t = _substitute_calls(_canonicalise(text), vars_, profile=profile) ```

PR #246, opened 2026-04-27
fix(diagram): harden Mermaid sanitizer edge cases

Hinotoi-agent:
## Summary

This PR tightens RAPTOR's Mermaid diagram sanitization and regression coverage across diagram generators.

Changes:
- escape ampersands before angle brackets and normalize additional line separators in diagram labels;
- keep `sanitize_id()` collision-resistant while still falling back to `node` when an ID contains no safe characters;
- use sanitized IDs in context-map `class` assignments, matching sanitized node declarations;
- sanitize flow-trace step-derived IDs before node, edge, and class emission;
- sanitize label fields in flow traces, hypotheses, attack paths, attack-tree enrichment, and findings-summary pie labels;
- add regression tests for callback-shaped Mermaid payloads, quoted-label breakout attempts, subgraph IDs, class assignments, and pie labels.

## Why

Mermaid node IDs and some directive-like constructs are not quoted. RAPTOR already sanitized several node declarations, but a review pass found additional places where attacker-controlled report data could still be rendered into IDs, class assignments, subgraph declarations, or quoted labels without going through the shared sanitizer. These paths matter because RAPTOR diagrams are generated from analysis artifacts and may be viewed in Mermaid-capable renderers.

## Self-review loops

I ran three improvement passes over the PR:

1. Sanitizer helper pass: added ID fallback/collision checks and line-separator normalization coverage.
2. Context/flow pass: checked class/edge/node sinks and added flow-trace regressions.
3. Cross-generator pass: checked hypotheses, attack paths, attack tree, and findings-summary pie labels for directive breakout cases.

## Files changed

| File | Change |
| --- | --- |
| `packages/diagram/sanitize.py` | Escape `&`, normalize additional line separators, and preserve safe ID replacement positions while retaining empty-ID fallback. |
| `packages/diagram/context_map.py` | Reuse `sanitize_id()` for class assignment IDs. |
| `packages/diagram/flow_trace.py` | Sanitize step-derived node IDs and user-controlled label fields. |
| `packages/diagram/hypotheses.py` | Sanitize subgraph IDs and hypothesis/prediction label fields. |
| `packages/diagram/attack_paths.py` | Sanitize step type and location label fields. |
| `packages/diagram/attack_tree.py` | Sanitize status/enrichment labels. |
| `packages/diagram/findings_summary.py` | Sanitize pie-slice labels before Mermaid emission. |
| `packages/diagram/tests/test_diagram.py` | Add sanitizer and generator regression tests. |

## Test plan

Validated against `8106a0c61af556fdf86ae3fb23cd0e5445948872`:

- [x] `python -m pytest packages/diagram/tests/test_diagram.py -q` — 80 passed
- [x] `python -m compileall -q packages/diagram core/security`
- [x] `git diff --check`
- [x] `python -m ruff check --ignore F401,E741 packages/diagram/sanitize.py packages/diagram/context_map.py packages/diagram/flow_trace.py packages/diagram/hypotheses.py packages/diagram/attack_paths.py packages/diagram/attack_tree.py packages/diagram/findings_summary.py packages/diagram/tests/test_diagram.py`
- [x] added-line scan for obvious fake/real secrets and dangerous Python execution sinks

Note: `ruff format --check` would reformat the legacy diagram test file broadly, so I left formatting churn out of this focused PR.

### DIFF
--- a/packages/diagram/attack_paths.py
+++ b/packages/diagram/attack_paths.py
@@ -62,9 +62,9 @@ def generate_single(path_data: dict[str, Any], path_index: int) -> str:
         nid = f"P{path_index}S{i+1}"
         # Steps may be objects or strings
         if isinstance(step, dict):
-            step_type = step.get("type", "call")
+            step_type = _sanitize(step.get("type", "call"))
             desc = _sanitize(step.get("description", step.get("action", str(step))))
-            loc = step.get("call_site") or step.get("definition") or ""
+            loc = _sanitize(step.get("call_site") or step.get("definition") or "")
             tainted = _sanitize(step.get("tainted_var", ""))
             parts = [f"[{i+1}] {step_type.upper()}"]
             if loc:
--- a/packages/diagram/attack_paths.py
+++ b/packages/diagram/attack_paths.py
@@ -62,11 +62,11 @@ def generate_single(path_data: dict[str, Any], path_index: int) -> str:
         nid = f"P{path_index}S{i+1}"
         # Steps may be objects or strings
         if isinstance(step, dict):
-            step_type = _sanitize(step.get("type", "call"))
+            step_type = _sanitize(str(step.get("type", "call")).upper())
             desc = _sanitize(step.get("description", step.get("action", str(step))))
             loc = _sanitize(step.get("call_site") or step.get("definition") or "")
             tainted = _sanitize(step.get("tainted_var", ""))
-            parts = [f"[{i+1}] {step_type.upper()}"]
+            parts = [f"[{i+1}] {step_type}"]
             if loc:
                 parts.append(loc)
             if tainted:
--- a/packages/diagram/attack_tree.py
+++ b/packages/diagram/attack_tree.py
@@ -265,7 +265,7 @@ def generate(
     # Style classes
     status_groups: dict[str, list[str]] = {}
     for node in nodes:
-        s = node.get("status", "unexplored")
+        s = _sanitize(node.get("status", "unexplored"))
         status_groups.setdefault(s, []).append(node.get("id", "?"))
 
     lines.append("")
--- a/packages/diagram/attack_tree.py
+++ b/packages/diagram/attack_tree.py
@@ -69,7 +69,7 @@ def _build_hypothesis_index(hypotheses: list[dict]) -> dict[str, str]:
     index: dict[str, str] = {}
     for h in hypotheses:
         fid = h.get("finding") or h.get("finding_id", "")
-        status = h.get("status", "")
+        status = _sanitize(h.get("status", ""))
         claim = h.get("claim") or h.get("hypothesis", "")
         if fid and fid not in index:
             index[fid] = f"{status}: {_sanitize(claim[:60])}" if claim else status
@@ -80,9 +80,7 @@ def _node_label(node: dict, proximity_idx: dict, disproven_idx: dict) -> str:
     nid = node.get("id", "?")
     goal = _sanitize(node.get("goal", node.get("technique", nid)))
     technique = _sanitize(node.get("technique", ""))
-    status = node.get("status", "unexplored")
-    node_type = node.get("type", "")
-
+    status = _sanitize(node.get("status", "unexplored"))
     parts = [goal]
     if technique and technique != goal:
         parts.append(technique)
--- a/packages/diagram/context_map.py
+++ b/packages/diagram/context_map.py
@@ -127,13 +127,13 @@ def generate(data: dict[str, Any]) -> str:
     lines.append("    classDef sink fill:#fee2e2,stroke:#dc2626,color:#7f1d1d")
 
     if entry_points:
-        ep_ids = ",".join(ep.get("id", "") for ep in entry_points)
+        ep_ids = ",".join(_sid(ep.get("id", "")) for ep in entry_points)
         lines.append(f"    class {ep_ids} ep")
     if boundary_details:
-        tb_ids = ",".join(tb.get("id", "") for tb in boundary_details)
+        tb_ids = ",".join(_sid(tb.get("id", "")) for tb in boundary_details)
         lines.append(f"    class {tb_ids} tb")
     if sink_details:
-        sink_ids = ",".join(s.get("id", "") for s in sink_details)
+        sink_ids = ",".join(_sid(s.get("id", "")) for s in sink_details)
         lines.append(f"    class {sink_ids} sink")
 
     return "\n".join(lines)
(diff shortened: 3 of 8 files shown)


grokjc:
Mechanics are right: & → &amp; before < → &lt; (chain order matters); line-separator coverage is the correct set of four; sanitize_id correctly handles "all chars stripped" via .strip('_'). Per-module _sid() application closes the gap where IDs landed unquoted in class assignments and subgraph headers.                                                                                                                  
                                                                                                 
The 80+ regression tests with full directive-injection payloads exercise the actual escape route — that's the right shape for this class of bug.                                                                 
                                                                                                                                                                                                                   
Two non-blocking notes for a follow-up:                                                                                                                                                                          
  1. Truncation runs after &amp; escape, so a long label can be cut mid-entity ("Tom &am..."). Cosmetic, not a security issue, but worth a comment.                                                                
  2. | isn't sanitized. No callers currently flow user data into edge labels, but a docstring note in sanitize.py would lock the invariant in.

Hinotoi-agent:
Thanks for the careful review. I addressed both follow-up notes in f9327f1:

- documented that truncation is applied after HTML entity escaping, so a long escaped label can be cut mid-entity as a cosmetic-only artifact rather than a Mermaid injection boundary;
- documented that `sanitize()` is for quoted node labels / similarly quoted text and intentionally does not escape `|`, so callers must not pass user-controlled text into unquoted edge labels.

Validation run:
- `uv run pytest packages/diagram/tests/test_diagram.py -q` — 80 passed
- `uv run ruff check packages/diagram/sanitize.py` — passed
- `uv run pytest -q` — still blocked by pre-existing static-analysis test import collection errors unrelated to this docstring-only change
- `uv run ruff check .` — still blocked by pre-existing repo-wide lint debt unrelated to this docstring-only change

grokjc:
Additional stylistic/cosmetic issues:

  1. attack_tree.py:270-282 — status used in class lines is not sanitized in generate():
  s = node.get("status", "unexplored")       # line 270 — raw
  status_groups.setdefault(s, []).append(...)  # used as dict key
  cls = status if status in (...) else "unexplored"  # line 281 — whitelisted
  This is safe because cls is validated against a string literal whitelist, and ids are sanitized upfront (line 173). But it means an attacker-controlled status value becomes a dict key that's silently mapped to "unexplored" — no injection, just a style mismatch.

  2. flow_trace.py — .upper() after sanitize:
  level = _sanitize(attacker_control.get("level", ""))
  ac_label = f"Attacker control: {level.upper()}\\n{what}"
  If level contained <, sanitize produces &lt;, then .upper() produces &LT; — cosmetically wrong but not a security issue. In practice, level is always "high"/"medium"/"low".

grokjc:
Would also be interested in the pytest and ruff issues if you want to raise them separately

danielcuthbert:
I too also want to ensure it still results in a usable mermaid diagram too, ill run those tests

Hinotoi-agent:
Thanks — I added a focused regression on top of the sanitizer hardening to make that explicit.

Commit b907e5b adds Mermaid usability coverage that verifies sanitized context-map, flow-trace, and attack-path outputs still have recognizable/renderable flowchart structure after malicious label/id inputs are escaped, and also checks rendered markdown Mermaid blocks keep valid entrypoints/fences.

Validation:
- `uv run pytest packages/diagram/tests/test_diagram.py -q` — 84 passed
- `uv run ruff check packages/diagram/tests/test_diagram.py` — passed
- `git diff --check` — passed
- `uv run pytest -q` — still blocked by the existing `packages/static-analysis/tests/*` import-collection errors noted earlier
- `uv run ruff check .` — still blocked by pre-existing repo-wide lint debt outside this touched file
- GitHub Actions for b907e5b — all PR jobs passed (`deps`, `python-unit-tests`, and bash test matrix)

grokjc:
> Additional stylistic/cosmetic issues:
> 
> 1. attack_tree.py:270-282 — status used in class lines is not sanitized in generate():
>    s = node.get("status", "unexplored")       # line 270 — raw
>    status_groups.setdefault(s, []).append(...)  # used as dict key
>    cls = status if status in (...) else "unexplored"  # line 281 — whitelisted
>    This is safe because cls is validated against a string literal whitelist, and ids are sanitized upfront (line 173). But it means an attacker-controlled status value becomes a dict key that's silently mapped to "unexplored" — no injection, just a style mismatch.
> 2. flow_trace.py — .upper() after sanitize:
>    level = _sanitize(attacker_control.get("level", ""))
>    ac_label = f"Attacker control: {level.upper()}\n{what}"
>    If level contained <, sanitize produces <, then .upper() produces < — cosmetically wrong but not a security issue. In practice, level is always "high"/"medium"/"low".

@Hinotoi-agent can you address these specifics?

grokjc:
Suggestions:

1. attack_tree.py:270 — status fed into status_groups is raw from the node dict. The allowlist on line 281 prevents injection, but for consistency with the rest of this PR, sanitize at the source: ```s = _sanitize(node.get("status", "unexplored"))```
2. flow_trace.py:159 — .upper() is called after _sanitize(), so &amp;lt; becomes &amp;LT; (broken entity). Fix: sanitize after uppercasing: ```level = _sanitize(attacker_control.get("level", "").upper())```                                                                                                                                    

Both are one-line changes. Tests should still pass since the valid status strings contain no special characters.

Hinotoi-agent:
Addressed both style/cosmetic sanitizer follow-ups in fd993fc, then added extra regression coverage in 6c527b9:

- `attack_tree.py`: sanitize `status` before it is used as the `status_groups` key, so odd status text no longer survives as an internal class-group key before being mapped to `unexplored`.
- `flow_trace.py`: uppercase the raw attacker-control level before sanitizing it, and avoid a second `.upper()` on the escaped label, so escaped entities like `&lt;` are not converted into broken `&LT;` text.
- Added explicit tests for those two review points: sanitized attack-tree status-group keys and attacker-control entity casing.
- Added one extra adjacent hardening check found while reviewing the same pattern: step type labels in both `flow_trace.py` and `attack_paths.py` now uppercase before escaping, with regressions proving `<sink>` renders as `&lt;SINK&gt;` rather than broken `&LT;SINK&GT;`.

Validation run:
- `uv run pytest packages/diagram/tests/test_diagram.py -q` — 88 passed
- `uv run ruff check packages/diagram/flow_trace.py packages/diagram/attack_paths.py packages/diagram/tests/test_diagram.py` — passed
- `uv run python -m compileall -q packages/diagram core/security` — passed
- `git diff --check` — passed
- added-line secret/sink scan — no findings
- GitHub Actions for 6c527b9 — all current PR jobs passed (`deps`, `python-unit-tests`, and bash test matrix)

This addresses both https://github.com/gadievron/raptor/pull/246#issuecomment-4330187846 and https://github.com/gadievron/raptor/pull/246#issuecomment-4330412273, plus adds the requested extra checks around the same escaped-uppercase edge case.

grokjc:
Ready to merge unless @danielcuthbert has any concerns?

PR #247, opened 2026-04-27
SMT: bridge /understand flow traces into Stage E feasibility

grokjc:
When /understand --trace walks a dataflow, it already extracts the path conditions for that flow.  Stage E was hand-extracting them again from the source — duplicate LLM work, with the LLM sometimes picking a different (wrong) path.

Adds an opt-in passthrough so the trace's conditions feed Stage E directly:

- trace.md schema gains two optional fields, path_conditions (list of predicates the flow knows must hold, as strings or {text, negated} dicts) and path_profile (e.g. "uint32" matching the dominant C type).
- core/orchestration/understand_bridge.py validates and forwards both fields when importing flow traces into attack-paths.json.  Malformed shapes are dropped with a logged warning rather than propagated.
- libexec/raptor-smt-validate-path adds a --from-trace TRACE-ID mode (with --workdir / --attack-paths overrides) that reads attack-paths .json, finds the trace by id, and uses its path_conditions/profile. Profile precedence: --profile flag > trace's path_profile > uint64.
- stage-e-feasibility.md [E-3c] tells the LLM to prefer --from-trace when the finding has an imported flow trace, with explicit handling for missing / empty / non-empty path_conditions.

Tests: 101 in test_smt_path.py + test_understand_bridge.py covering round-trip preservation, malformed-input rejection, profile precedence, trace lookup, conflict-with-stdin error path, and an end-to-end negated-condition round-trip (bridge → file → libexec → solver). Full repo suite: 2427 passed (1 pre-existing skip).

### DIFF
--- a/.claude/skills/code-understanding/trace.md
+++ b/.claude/skills/code-understanding/trace.md
@@ -52,6 +52,21 @@ Identify all branches in the flow — not just the happy path:
 
 Trace each significant branch separately.
 
+**[TRACE-3b] Path Conditions (for SMT consumption)**
+
+Separately from `branches[]` (which describes *alternative* paths), record the **conjunction of guards on the dangerous path** — every condition that must hold for execution to actually reach the sink along the trace you've followed. These are downstream-consumed by the SMT path-feasibility helper (`raptor-smt-validate-path`), so emit them as simple bitvector predicates over named variables:
+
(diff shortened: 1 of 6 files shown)


grokjc:
@unprovable FYI

unprovable:
Ah, I get it! Ok, this is starting to feel a lot more like an LLM friendly workflow than just a parse-and-solve-and-done thing. I like it!

PR #250, opened 2026-04-27
SMT: named verbs for CWE-190 / OOB / null-deref, csem-direct

grokjc:
Three pattern-named SMT verbs that the LLM reaches for when the source matches a known shape, instead of reconstructing the encoding ad-hoc through validate_path():

- check_overflow(operands, op, profile, kind=unsigned_wrap|signed_wrap)
- check_oob(buffer_size, index, profile)
- check_null_deref(ptr, profile)

Each is a thin verb that builds its intrinsic Z3 predicate directly via core.smt_solver.csem (uadd_overflows, umul_overflows, usub_underflows, sadd_overflows, smul_overflows) — no string concat, no parser round-trip for the verb's own predicate.  The intrinsic is provably right by construction (csem is the C-semantics module).

Operand forms: identifier (count), numeric literal (16, 0x10), or balanced function-call shape (strlen(input), sizeof(record)) — function calls go through the same free-variable fallback as PR #244, allocating a fresh _anon_N Z3 variable.  The over-approximation caveat from #244 applies and is documented in the skill prompt.

Caller-supplied guards still go through the standard parser, inheriting PR #241 (English-phrase canonicalisation) and PR #244 (free-var fallback) transparently.  Compound non-call operands (a + b, *p, s->len) are rejected at the verb layer with a clear ValueError pointing at raptor-smt-validate-path for free-form predicates.

Multi-operand chains compose correctly: per-step overflow predicates ORed together, so wrap at any binary step in a chain like a*b*c is detected (not just the final result vs first operand).

Profile/kind compatibility checked at the verb layer: unsigned_wrap requires unsigned profile; signed_wrap requires signed. signed_wrap not supported with '-' (subtraction's signed-overflow shape needs opposite-sign operands, too narrow for the binary template).

smt_path_validator.py refactored:
- Factor _classify_text_condition + _solve_pending out of check_path_feasibility — shared with the new check_verb_feasibility
- New public make_var / make_val / make_anon_call_var / is_balanced_call helpers for verb operand construction
- _next_anon_index helper shared between parser-side _substitute_calls (#244) and verb-side make_anon_call_var so anon namespaces don't collide.  Existing 79 path-validator tests pass unchanged.

Stage E skill (stage-e-feasibility.md) gains:
- [E-3 SMT] decision tree: --from-trace first (PR-2), named verbs second (this PR), free-form validate-path third (#225)
- [E-3d] section listing the three verbs with usage and CLI invocations

Tests: 62 verb tests covering canonical bug shapes, profile/kind mismatch rejection, compound-operand rejection, function-call operand fallback, distinct-anon-vars-for-distinct-calls, JSON serialisation, libexec end-to-end. Full repo: 2596 passed, 1 pre-existing skip.

### DIFF
--- a/.claude/skills/exploitability-validation/stage-e-feasibility.md
+++ b/.claude/skills/exploitability-validation/stage-e-feasibility.md
@@ -147,6 +147,18 @@ arithmetic), CWE-193 (off-by-one), CWE-476 (null deref).
 
 The heuristic `note` field is always preserved alongside `smt_feasibility` — compare both when they conflict. When `smt_feasibility.feasible` is `false`, prefer the SMT verdict for the `chain_breaks` entry.
 
+### [E-3 SMT] Choosing an SMT entry point
+
+You have three SMT tools, listed in the order to reach for them:
+
+| Situation | Tool | Why first |
+|---|---|---|
+| The finding has an imported `/understand` flow trace with `path_conditions` (look in `attack-paths.json` for entries with `source: "understand:trace"` carrying that field) | `raptor-smt-validate-path --from-trace TRACE-ID --workdir "$OUTPUT_DIR"` ([E-3c]) | The trace's conditions were extracted alongside the dataflow walk with full source-code context — usually richer and more accurate than re-extracting from prose |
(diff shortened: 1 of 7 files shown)


grokjc:
@unprovable FYI, any thoughts before I merge?

unprovable:
Ah, nice idea! Adding the safeguards (like 'Don't call multiple tools for the same finding — pick one per the decision table') is essential. Having tools that check per-CWE from codeql output is very handy. I'd certainly merge to open up access to the Z3 capability!

PR #253, opened 2026-04-28
SMT path validator: C operator precedence + grouping parens  

unprovable:
Replace the strict left-to-right expression parser in `_parse_expr` with a precedence climber over a four-level C precedence table, and accept balanced grouping parentheses as part of the same change.                                                                                  

Two friction points hit real LLM output regularly:

1. Mixed-class expressions like `a + b * c <= max` rejected wholesale with `MIXED_PRECEDENCE`, even though the C reading is unambiguous.
2. Non-call grouping parentheses (`(a + b) * c`) rejected with `PARENS_NOT_SUPPORTED` because the LTR parser had no way to override precedence — the only escape hatch we offered was "flatten by hand."                                                           

Both classes are eliminated by this change. The cost is silent re-interpretation of inputs that previously rejected; the gain is that those inputs now contribute to feasibility analysis with C-correct semantics.                                                           

### Operator precedence 

Within an expression: `*` > `+ -` > `<< >>` > `|`, all left-associative. The notable C surprise: shifts bind **less tightly** than additive operators, so `a + b << 2` parses as `(a + b) << 2`, not `a + (b << 2)`. Documented in the module docstring; use parentheses to make the grouping explicit when the C reading isn't what was meant.                                                                                                   

### Backward compatibility

- Every single-operator-class input that parsed before still parses to the same Z3 expression (additive/multiplicative/shift/OR are all left-associative under both the old and new parser). 
- `RejectionKind.MIXED_PRECEDENCE` and `RejectionKind.PARENS_NOT_SUPPORTED` remain in the enum, marked deprecated. No encoder emits them any more. They are retained so external consumers that match on these values keep compiling.
- New `RejectionKind.UNBALANCED_PARENS` covers the structurally-broken paren cases (extra `(` / `)`, mismatched nesting, count not returning to zero).                                                       
                                                                                                                                    
### What Has Changed                                                                                                                  
                                                                                                                                    
**`core/smt_solver/rejection.py`**
- Added `UNBALANCED_PARENS`
- Updated docstrings on the two deprecated kinds                                                                                  
                                                                                                                                    
**`packages/codeql/smt_path_validator.py`**
- Tokeniser captures `(` and `)`
- `_parse_expr` rewritten as a precedence climber; atoms include parenthesised subexpressions, which recursively re-enter the climb with `min_prec=0`
- Trailing-token classification routes `(`/`)` to `UNBALANCED_PARENS`, relational tokens to `UNSUPPORTED_OPERATOR`, anything else to `TRAILING_TOKENS`
- `_parse_condition` replaces its early `'(' in t` rejection with a one-pass paren-balance scan that emits `UNBALANCED_PARENS` for the no-relational-op cases that wouldn't otherwise reach `_parse_expr`
- Module docstring updated to describe C precedence and the shift surprise                                                        
                                                                                                                                    
  **Tests**                                                                                                                         
  - Repurposed the `MIXED_PRECEDENCE` / `PARENS_NOT_SUPPORTED` assertions in `packages/codeql/tests/test_smt_path_validator.py`                                                                              
  - `test_unbalanced_parens_still_rejected` now asserts `UNBALANCED_PARENS`                                                         
  - Added `TestPrecedence` (6 cases): pin C precedence by asserting concrete values that distinguish the correct parse from any LTR misparse (e.g., `a + b * c == 64` is feasible iff parsed as `a + (b * c)`)                                                               
  - Added `TestParensGrouping` (10 cases): grouping override, nested parens, parens on both sides of a relational, bitmask LHS, all four imbalance modes (open-without-close, close-without-open, swapped `)(`, empty `()`)                                                        
  - Updated `packages/exploit_feasibility/tests/test_smt_path.py`; one downstream test used `(a + b) > (c + d)` as a "still-unparseable" example; replaced with `~mask == 0xFFFFFFFF` (NOT operator silently dropped, caught by the consumed-input check)

### DIFF
--- a/core/smt_solver/rejection.py
+++ b/core/smt_solver/rejection.py
@@ -52,13 +52,24 @@ class RejectionKind(str, Enum):
     """An operator outside the accepted set appeared in the expression."""
 
     PARENS_NOT_SUPPORTED = "parens_not_supported"
-    """Input contained ``(`` or ``)`` — function calls and grouping
-    aren't supported by the current grammar."""
+    """Deprecated — kept for backward compatibility with downstream
+    consumers that match on this value.  No encoder emits it any more:
+    the path validator's expression parser now supports grouping
+    parentheses via precedence climbing, and unbalanced cases emit
+    :data:`UNBALANCED_PARENS` instead."""
+
+    UNBALANCED_PARENS = "unbalanced_parens"
+    """Input had ``(`` without a matching ``)`` (or vice versa).  Fired
+    by the path validator's expression parser when the bracket structure
+    of a grouping subexpression doesn't close cleanly, or by the
+    condition-level balance check before dispatch."""
 
     MIXED_PRECEDENCE = "mixed_precedence"
-    """Expression mixed additive and multiplicative/bitwise operators.
-    The parser is strictly left-to-right with no precedence, so it
-    rejects mixed forms rather than risk silent mis-encoding."""
+    """Deprecated — kept for backward compatibility with downstream
+    consumers that match on this value.  No encoder emits it any more:
+    the path validator's expression parser now uses C operator precedence
+    (``*`` > ``+ -`` > ``<< >>`` > ``|``) and accepts mixed-operator
+    expressions directly.  Use parentheses to override precedence."""
 
     TRAILING_TOKENS = "trailing_tokens"
     """Tokens were left unconsumed after parsing (e.g. ``a b``)."""
--- a/packages/codeql/smt_path_validator.py
+++ b/packages/codeql/smt_path_validator.py
@@ -37,18 +37,17 @@ mis-encoded):
 
   - **Operators outside the supported set.**  Accepted: ``+ - * |``,
     relational ``< <= > >= == !=``, shifts ``<< >>``, bitmask
-    ``&`` (only in the ``flags & MASK == VAL`` form).  Rejected:
-    unary NOT (``~``), XOR (``^``), division (``/``), modulo (``%``),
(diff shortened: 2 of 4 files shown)


grokjc:
A few minor tweaks:
                                                          
Python-vs-C precedence is going to bite the LLM at some point. The docstring calls out that a + b << 2 parses as (a + b) << 2 (correctly, per C) — but LLMs trained on Python expect the opposite (Python's << binds tighter than +). With the old parser this got MIXED_PRECEDENCE and a clear retry signal; now it just produces a confidently wrong verdict with no rejection. Probably worth a one-paragraph warning in [E-3c] of the Stage E skill telling the LLM to parenthesise explicitly when it isn't sure.
             
Recursion depth is unbounded. Precedence climber is recursive, so ~500 nested parens would hit Python's default stack limit. Real risk is low (LLM-extracted conditions aren't deeply nested), but a depth cap rejecting at e.g. 100 would be defensive.
                                                                                                                                                                                            
Small UX one: (a == b) + 1 rejects with UNBALANCED_PARENS ("expected ')'") when the real issue is that == isn't a binary arithmetic op. The LLM might waste a refinement pass chasing parens. Easy fix in _atom's paren branch — if the next token is relational, prefer UNSUPPORTED_OPERATOR. And while you're in there, _RELATIONAL_TOKENS includes & (which is bitmask, not relational) — name is slightly misleading.

This is already mergeable but we've been polishing as we go, so I'm sure you don't mind addressing these too! 😄

grokjc:
I had a stab at those fixes myself, hopefully they work - let me know if not! 😄

Issue #73, opened 2026-03-10
[Feature Req] Code Coverage


Hamid-K:
Thank you for Raptor! 

I was wondering if you have any code-coverage backed reporting and enforcement planned as part of upcoming updates? 

Current pipelines, including Raptor, when tasked to audit binaries and repositories, even when relying on SAST tools which we assume would provide decent coverage, tend to ignore notable chunk of a project or binary. That seems to be partly assumptions and hallucinations by the LLM, partly the default on-demand review of code and partly poor SAST coverage (unless custom queries are tasked to be modeled and generated). 

The outcome would be missing potential vulnerabilities, albeit often less critical ones.
As I've been experimenting with my own half-manual pipeline, I consistently get better/more results when I enforce the LLM to actually map the entire codebase, create a coverage map and checklist and go through all of it. Interestingly, enforcing that also improves how CodeQL results are consumed, as it would reveal if standard build and queries for CodeQL are failing or partially failing.

grokjc:
#153 will add two important features: we'll be able to track what the LLM actually read + all tools will write coverage records.

grokjc:
#167 gives you the basics. There is a lot more to come, but this surfaces what we already collect and analyse, and enables conversations with the LLM about code coverage. See also 'raptor project coverage' command.

Hamid-K:
Thank you for your great work and the implementation!

grokjc:
There's more to do but #691 is the next step toward where we need to be.

grokjc:
This is where it starts to get interesting 😄  
 | PR | What it landed |
  |------|----------------|
  | #691 | Persistent verdict-bearing coverage **store** + Phase-1 inventory completion (interstitial/top_level/C-globals) + producer→consumer integration |
  | #699 | **Foundation hardening** — `lines` standardisation, extractor kinds (multi-declarator/typedefs), tolerant `schema.py`, bitmap fallback, `--keep 0` |
  | #708 | **Reporting unification** — one store-backed report (state + execution detail), `/understand` fold-in, function-level LLM marks, `--detailed` per-file table; retired `compute_summary` |
  | #710 | **Phase 4 import** — gcov/lcov/coverage.py parsers + `raptor-coverage-summary --import` |
  | #714 | **Phase 4 collection layer** — gcov-build/llvm-cov collectors + DWARF/drcov/sancov binary→source |

grokjc:
@Hamid-K — closing this out. Your framing shaped the architecture.

[...]

PR #257, opened 2026-05-02
feat(project): add coverage threshold gate

Hinotoi-agent:
## Summary
- Adds `raptor project coverage --fail-under <pct>` as a coverage gate for project coverage reports.
- Computes LLM item coverage from the existing coverage summary shape.
- Prints a clear PASS/FAIL threshold line and exits non-zero when the threshold is not met.
- Adds unit coverage for the threshold helpers.

Addresses part of #73.

## Why
Issue #73 asks for coverage-backed reporting and enforcement so operators can avoid assuming that an audit reviewed the full target. RAPTOR already records checklist/coverage data; this adds a small enforcement hook that can be used in local workflows and CI.

Example:

```bash
raptor project coverage my-project --fail-under 80
```

## Test plan
- `python3 -m pytest core/coverage/tests -q`
- `python3 -m compileall -q core/coverage core/project/cli.py`
- `git diff --check`

### DIFF
--- a/core/coverage/summary.py
+++ b/core/coverage/summary.py
@@ -211,6 +211,32 @@ def _pl(n: int, word: str) -> str:
     return f"{n} {word}" if n == 1 else f"{n} {word}s"
 
 
+def llm_item_coverage_percent(summary: Dict[str, Any]) -> float:
+    """Return percentage of inventory items reviewed by LLM coverage records."""
+    if not summary:
+        return 0.0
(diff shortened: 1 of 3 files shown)


danielcuthbert:
Thanks for this, looks like a good addition. will review

danielcuthbert (approved):
for those running raptor in a CI, this is helpful (I know I will find it useful), thanks

PR #269, opened 2026-05-02
feat(sandbox): --audit mode

grokjc:
Adds operator-facing --audit / --audit-verbose flags that compose with any non-'none' sandbox profile. Audited workflows run to completion AND record what enforcement WOULD have blocked, giving operators visibility without losing observability the way --no- sandbox would.

Three layers, each audit-mode capable:

- b1 (egress proxy): hostname allowlist gate switches to log-and- allow when --audit acquires the singleton's audit ref-count. CONNECT to non-allowlisted hosts emits a 'would_deny_host' event AND records to sandbox-summary.json. Resolved-IP block (DNS- rebinding defense, gate 2) stays enforcing — pure attack signal. Ref-counted acquire/release prevents concurrent mixed-profile sandboxes from racing on the singleton.

- b2 (syscall): seccomp filter swaps deny action from SCMP_ACT_ERRNO to SCMP_ACT_TRACE. Kernel pauses on each blocked syscall and notifies the attached ptrace tracer (core/sandbox/tracer.py) which logs and resumes. Fail-closed install: any seccomp_load failure -> os._exit(126), matching Landlock's posture.

- b3 (filesystem + network): tracer follows open / openat / openat2 / connect via process_vm_readv path-resolution; multi- process via PTRACE_O_TRACEFORK/VFORK/CLONE; multi-arch (x86_64
  + aarch64) syscall tables; PTRACE_O_EXITKILL cascades SIGKILL to tracees if tracer dies. Filtered mode logs only would-be- blocked events; --audit-verbose logs every traced syscall.

Audit-target routing (audit_run_dir kwarg):

The new audit_run_dir= kwarg decouples 'where audit JSONL lands' from 'what Landlock restricts writes to'. Callers like codeql analyze pass audit_run_dir without output= so they get audit signal without taking a writable_paths restriction that would break the workflow.

Graceful degradation:

When audit was requested but couldn't engage (mount-ns blocked, cmd[0] outside bind tree, no target/output for tracer attach, speculative-retry fall-through, etc.), record_audit_degraded writes sandbox-audit-degraded.json with a precise reason + operator instructions. Operator distinguishes 'audit ran, no events' from 'audit didn't run' at a glance.

CLI propagation:

raptor.py passes --audit through to subprocesses (raptor_agentic forwards to scanner.py + codeql/agent.py via sandbox_passthrough). codeql/agent.py registers add_cli_args/apply_cli_args so the flag is recognised at subprocess parse time. scanner.py + codeql/agent .py call set_active_run_dir(out_dir) at start-of-main + summarize _and_write(out_dir) at end-of-main so audit JSONL written by the tracer aggregates into sandbox-summary.json within their own subprocess (parent agentic's lifecycle covers $RUN/ but not subdirs).

Built on top of PR #265's mount-ns auto-fallback + tool_paths infrastructure: when audit + tool_paths supplied but mount-ns fails at exec (typical Python tool with native deps outside the bind set), speculative-C retry routes via Landlock-only and the audit-degraded marker fires.

Tests: 598 sandbox tests + 54 audit-specific tests + comprehensive E2E covering 8 of the 16 known coverage gaps (multi-process TRACEFORK, --audit-verbose, openat2 trace, io_uring_setup with visibility-gap note, --sandbox debug --audit composition, --sandbox network-only --audit composition, egress proxy audit_log_only with would_deny_host, concurrent record_audit_degraded race-safety).

### DIFF
--- a/core/sandbox/_spawn.py
+++ b/core/sandbox/_spawn.py
@@ -566,7 +566,19 @@ def run_sandboxed(
             os.set_inheritable(t_ready_w, True)
             _parent_fds.update({t_ready_r, t_ready_w})
 
-        child_pid = os.fork()
+        # Suppress Python 3.12+ DeprecationWarning about multi-threaded
+        # fork(). Our post-fork code does namespace setup via ctypes
+        # syscalls + Landlock + seccomp + execvp — no Python objects,
+        # no GIL acquisition, no malloc-arena access. posix_spawn()
+        # can't do the bespoke namespace setup, so we need raw fork.
+        # See module docstring for the fork-safety contract.
+        import warnings
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore", category=DeprecationWarning,
+                message=r".*fork.*may lead to deadlocks.*",
+            )
+            child_pid = os.fork()
     except BaseException:
         # Any failure before fork returns: close opened pipes AND the
         # mkdtemp stub. Without this, a pipe-exhaustion OSError or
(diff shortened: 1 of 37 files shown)


grokjc:
I have an OSX sandbox working that relies on this being merged first 😄

danielcuthbert:
@grokjc keen to see your macos one, i have a partial one too but will get yours in over mine

grokjc:
This audit functionality is incremental and doesn't touch much of the existing workflow, so tempted to get this merged and in the hands of testers. Meanwhile I'm just finishing off the equivalent audit layer for OSX and we'll have parity (almost) across Linux and Mac. 🥳

danielcuthbert (approved):
LGTM

PR #273, opened 2026-05-02
feat: anti-prompt-injection defense envelope with per-model profiles

grokjc:
Wraps every LLM callsite in a layered defense envelope — per-model profiles, nonce-tagged boundaries, input preflight, output sanitisation, cross-family schema validation retry, and Rule of Two enforcement for agentic passes.


danielcuthbert (commented):
The test \`test_instruction_split_across_lines\` explicitly documents that \`"ignore all\nprevious instructions"\` doesn't trigger preflight, and the comment calls it a "known design choice." Fair enough for a non-blocking signal, but it's worth being explicit in the PR that this is accepted scope rather than a gap to revisit. Any attacker with write access to the scanned repo can trivially split any injection phrase across two lines and walk straight past the regex corpus. The preflight layer is advisory regardless, so this doesn't break anything, but I'd rather we document the assumption than leave it implicit.

Suggestion: add a \`re.MULTILINE\` variant of the highest-risk patterns (the "ignore previous instructions" family) as a separate corpus file, e.g. \`english_multiline.txt\`, compiled with \`re.IGNORECASE | re.MULTILINE | re.DOTALL\`. Keep the single-line corpus as-is so the false-positive rate stays low for the bulk of content. The multiline file would be opt-in per call site, used only where the input is short and structured (SCA install scripts, \`/understand\` verdicts) rather than bulk source code where the false-positive risk is higher.

danielcuthbert (commented):
When \`probe_envelope_compatibility\` fails, the profile silently degrades to \`PASSTHROUGH\`, which strips out envelope XML, datamarking, base64, and slot discipline, leaving only markdown stripping. That's logged at WARNING to \`raptor.security\` but nothing surfaces to the operator during the actual scan. If a model intermittently flunks the probe (flaky structured-output day, rate-limited, whatever), the run proceeds with most defences off and the operator has no idea. I'd want to see the degraded state reflected in the scan summary or terminal output at minimum, not just buried in the logger.

Suggestion: \`probe_envelope_compatibility\` already returns a \`ProbeResult\` with \`.compatible\`. The caller (the orchestrator) should check this and, when \`False\`, print a visible warning to stdout before proceeding, something like \`"DEFENSE WARNING: envelope probe failed for {model_id}, running with reduced defences. See defense-telemetry.json for details."\` The telemetry JSON already captures it; this just makes sure the operator sees it without having to go looking. Optionally, add a \`--require-envelope-probe\` flag that makes a failed probe a hard stop rather than a degraded run.

danielcuthbert (commented):
These preflight patterns will fire on legitimate RAPTOR output:

\`\`\`
"is_exploitable": false
"ruling": "false_positive"
"safe": true
\`\`\`

Those are standard fields in ruled-out findings. If \`preflight()\` ever gets called on RAPTOR's own JSON in a re-analysis pass, a validate stage reading a prior findings file, or anything similar, the haircut fires on your own data. Needs a comment in the file explicitly stating these patterns are only safe to run against incoming untrusted content, not RAPTOR's own output, and ideally a test or guard in the call sites that enforces this.

Suggestion: two changes. First, add a module-level docstring note to \`structural_injection.txt\` stating explicitly that the JSON field patterns assume the input is attacker-controlled source content, not RAPTOR output. Second, add a thin wrapper around \`preflight()\` in the call sites that only applies the structural corpus when the content origin is \`"source-code"\` or \`"scanner-message"\`, and skips it for \`"agent-output"\` or \`"raptor-findings"\`. The \`UntrustedBlock.kind\` field is already in the right place to carry this distinction.

grokjc:
> Suggestion: add a `re.MULTILINE` variant of the highest-risk patterns (the "ignore previous instructions" family) as a separate corpus file, e.g. `english_multiline.txt`, compiled with `re.IGNORECASE | re.MULTILINE | re.DOTALL`. Keep the single-line corpus as-is so the false-positive rate stays low for the bulk of content. The multiline file would be opt-in per call site, used only where the input is short and structured (SCA install scripts, `/understand` verdicts) rather than bulk source code where the false-positive risk is higher.

added english_multiline.txt corpus compiled with re.MULTILINE|re.DOTALL for high-risk phrases. preflight() now accepts corpora=() so callers opt in to multiline matching for short inputs only.  Single-line corpus unchanged.

grokjc:
> Suggestion: `probe_envelope_compatibility` already returns a `ProbeResult` with `.compatible`. The caller (the orchestrator) should check this and, when `False`, print a visible warning to stdout before proceeding, something like `"DEFENSE WARNING: envelope probe failed for {model_id}, running with reduced defences. See defense-telemetry.json for details."` The telemetry JSON already captures it; this just makes sure the operator sees it without having to go looking. Optionally, add a `--require-envelope-probe` flag that makes a failed probe a hard stop rather than a degraded run.

Probe failure already hard-stops unless --accept-weakened-defenses is passed — so the default is strict. Made the passthrough-mode banner more prominent with the failure reason and a description of the model-independent floor that still applies.

grokjc:
> Suggestion: two changes. First, add a module-level docstring note to `structural_injection.txt` stating explicitly that the JSON field patterns assume the input is attacker-controlled source content, not RAPTOR output. Second, add a thin wrapper around `preflight()` in the call sites that only applies the structural corpus when the content origin is `"source-code"` or `"scanner-message"`, and skips it for `"agent-output"` or `"raptor-findings"`. The `UntrustedBlock.kind` field is already in the right place to carry this distinction.

Right — the JSON override patterns ("ruling": "false_positive" etc.) will fire on our own findings. These patterns are designed to detect adversarial content in attacker-controlled input (target repo source, commit messages, issue text) where an attacker tries to override RAPTOR's output schema; they're not meant to scan RAPTOR's own structured data. Dispatch preflight now excludes structural_injection via corpora=. Added a scope header to the corpus file documenting this assumption.

grokjc:
Thanks for review, this one has been in the works for a while and can't see the wood for the trees 😄

danielcuthbert:
I love this stuff, it's bloody clever and I like where we are going with this, well done

PR #286, opened 2026-05-03
core/llm: explicit preference + Claude Code as a fully usable provider

grokjc:
Two layers of substrate work that together make Claude Code a viable tool-use backend for consumers like cve-diff Phase 2 — including the keyless case (no API key, just ``claude`` CLI on PATH).

RESOLVER LAYER
==============

* ``get_client(prefer="anthropic")`` — explicit consumer preference. Tries the named providers via env var first; falls through to operator's thinking-model config; falls through to default-order autodetect. Lenient: unknown / unavailable preferred providers are silently skipped. Consumers express their preference in code rather than depending on coincidence with the substrate's default order (which would silently regress consumer behaviour if re-tuned).

* Claude Code subprocess fallthrough — when no API key is set and no Ollama is running but ``claude`` CLI is on PATH, the resolver returns ``ModelConfig(provider="claudecode")``. cve-diff (and any future tool-using consumer) works for keyless users. Used only as last resort: any API key or Ollama beats it.

* ``LLMClient.primary_provider`` property — exposes the underlying ``LLMProvider`` for tool-use consumers driving a ``ToolUseLoop`` directly. Cached via the existing ``_get_provider`` machinery; raises ``RuntimeError`` if no primary_model is configured.

Resolution order, pinned by tests:
  1. preferred providers via env var (consumer's signal)
  2. operator's thinking-model config (file-based; covers non-env-var setups like Vertex auth for Gemini)
  3. default-order autodetect: Anthropic > OpenAI > Gemini > Mistral >
     Ollama > Claude Code (subprocess, last resort)

PROVIDER LAYER
==============

* ``ClaudeCodeLLMProvider.turn()`` via ``--json-schema`` structured output. The ABC's JSON-in-prompt synthesis (``_tool_use_fallback``) does not work for CC: anti-prompt-injection training refuses to roleplay as a different agent system when a system prompt says "you have these tools, emit JSON to call them" — that framing is indistinguishable from an attacker injecting a fake tool schema, and CC correctly refuses (the same hardening that makes CC safe to use against attacker-influenced content per #273).

  The ``--json-schema`` path reframes the task as form-filling rather than roleplay and bypasses the guard. We give CC a discriminated-union schema (either ``tool_call`` or ``complete``) plus the tool catalog as reference material, and CC fills in the form. Verified empirically: real CC subprocess drives multi-turn cve-diff-shaped tool catalogs, including cross-turn data dependencies (osv_raw output → gh_commit_detail input) and terminal-tool detection (``terminal_tool="submit_result"`` returns the structured payload via ``ToolLoopResult.terminal_tool_input``).

* ``generate_structured`` now returns ``StructuredResponse`` with per-call cost / tokens populated. Fixes a concurrent-loop race where two threads on the same provider could see each other's cost in the delta computation. Existing ``result, raw = generate_structured(...)`` tuple-unpack callers keep working via ``StructuredResponse.__iter__``.

* Anti-value-guessing rules in the substrate's CC system prompt: values placed into ``tool_input`` must come from conversation history or the user's request, not guessed from training data. Substrate-level mitigation; consumer-level gates (cve-diff's verified-SHA check) remain the second line — model behaviour is not deterministic, so prompt + downstream gate are belt-and-braces.

* CC's per-turn timeout bumped from 120s to 300s. Calibrated against real ``--json-schema`` runs over a 16-tool catalog — simple turns are 5-15s, rich-catalog turns can hit 60-180s. 300s gives 2-3x headroom without letting a single turn consume a whole ``ToolUseLoop.max_seconds`` budget.

Closes the gap that PR #282 promised but didn't deliver: a unified tool-use loop that works regardless of which LLM is configured, including the keyless Claude-Code-only case.

### DIFF
--- a/core/llm/client.py
+++ b/core/llm/client.py
@@ -290,6 +290,29 @@ class LLMClient:
 
         return self.providers[key]
 
+    @property
+    def primary_provider(self) -> LLMProvider:
+        """The :class:`LLMProvider` for the configured ``primary_model``.
+
+        Exposed publicly so consumers that need direct provider access
+        — typically for tool-use loops via :class:`core.llm.tool_use.ToolUseLoop` —
+        can reach it without going through :meth:`generate`. Cached;
+        the same instance is returned across calls.
+
+        Raises ``RuntimeError`` if no primary model is configured (the
+        client should normally not have been constructed in that
+        case — :func:`packages.llm_analysis.get_client` returns
+        ``None`` instead).
+        """
+        if self.config.primary_model is None:
+            raise RuntimeError(
+                "LLMClient has no primary_model configured; cannot "
+                "expose primary_provider. Use packages.llm_analysis."
+                "get_client() which returns None when no provider is "
+                "available, instead of constructing LLMClient directly."
+            )
+        return self._get_provider(self.config.primary_model)
+
     def _get_cache_key(self, prompt: str, system_prompt: Optional[str], model: str) -> str:
         """Generate cache key for prompt."""
         content = f"{model}:{system_prompt or ''}:{prompt}"
--- a/core/llm/config.py
+++ b/core/llm/config.py
@@ -189,92 +189,201 @@ def _get_best_thinking_model() -> Optional['ModelConfig']:
     return best_model
 
 
-def _get_default_primary_model() -> Optional['ModelConfig']:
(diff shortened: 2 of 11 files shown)


grokjc:
Still working on this

PR #302, opened 2026-05-04
feat: multi-model analysis and correlation

grokjc:
Thanks to Nicolas Krassas for the input

--model is repeatable: N models each independently analyse every finding via Cartesian product dispatch, then results are correlated. Includes judge role, SCA exploit/patch dispatch, /project correlate, and cross-run finding correlation.


grokjc:
See also #304

PR #305, opened 2026-05-04
feat: OpenAnt integration — source-code LLM vulnerability scanner

gadievron:
## Summary

- Adds `packages/openant/` bridge package: path discovery, subprocess-isolated scanner (PYTHONPATH=core_path, never sys.path), schema translator, SARIF deduplication
- Adds `raptor_openant.py` standalone workflow launcher with full run lifecycle
- Adds `libexec/raptor-openant` bash wrapper and `/openant` slash command
- Adds `--openant`/`--openant-only`/`--openant-core`/`--openant-model`/`--openant-level` flags to `raptor_agentic.py` Phase 1b
- Adds `openant` mode to `raptor.py` unified launcher
- CLEANUP-C1: `Path.resolve(strict=True)` FileNotFoundError re-raised as RuntimeError (unified exception type at boundary)
- CLEANUP-B1: TOCTOU window in merge block documented and pinned with tests

## Bug fixes

| Bug | File | Fix |
|-----|------|-----|
| BUG-R-004 | `packages/llm_analysis/agent.py` | Move Feasibility import out of `__init__` to module level |
| BUG-R-009/010 | `core/run/metadata.py` | Add `"openant": "openant"` to `_PREFIX_MAP` + test |
| BUG-R-011 | `raptor_agentic.py` | OpenAnt findings merged into `validation/findings.json` before Phase 3 |
| BUG-R-012 | `raptor_openant.py` | Remove dead `--no-analyze` flag (declared, never checked) |
| BUG-R-013 | `packages/openant/scanner.py` | PYTHONPATH validated with `resolve(strict=True)` + sentinel file check |
| BUG-R-015 | `packages/openant/scanner.py` | Full stderr persisted to disk; capped at 1 MiB (disk fill prevention) |
| BUG-R-016 | `raptor_agentic.py` | Merge block now handles `validation/findings.json` as dict (Raptor's actual format) |
| BUG-R-016-VARIANT | `raptor_agentic.py` | `--openant-only` else branch writes dict format, not plain list (Phase 3 AttributeError crash) |
| BUG-R-017 | `packages/openant/scanner.py` | Use OpenAnt's `.venv/bin/python3` (has tree-sitter bindings), not `sys.executable` |
| BUG-R-018 | `packages/openant/scanner.py` | Unrecognized `--language` values (e.g., `zig`) fall back to `auto` |
| os.access fix | `packages/openant/scanner.py` | `_find_venv_python` uses `os.access(X_OK)` not `.exists()` (executability check) |

## Tests added (86 total, all pass)

| File | Count | What |
|------|-------|------|
| `packages/openant/tests/test_translator.py` | 26 | Schema translation, dedup, verdict mapping |
| `packages/openant/tests/test_scanner.py` | 22 | PYTHONPATH validation, stderr persistence, venv Python selection, language fallback |
| `packages/openant/tests/test_phase1b_integration.py` | 21 | Phase 1b merge block (BUG-R-016/016-VARIANT/011), dict/list format handling |
| `packages/openant/tests/test_config.py` | 9 | Path discovery, env resolution, validation |
| `packages/openant/tests/test_raptor_dispatch.py` | 8 | Dispatch integration coverage |

## Test plan

- [x] `python3 -m pytest packages/openant/tests/ -q` — 86 passed, 5 subtests passed
- [x] Multi-language smoke test (Python, C, Ruby, PHP, Go, JS, Zig) via raptor_openant.py → OpenAnt scan
- [x] Judge agent review: all fixes PASS (verified files read against claims)
- [x] Repo confirmed private via `gh api repos/gadievron/raptor --jq '.private'` → true

---

> [!NOTE]
> **Medium Risk**
> Adds a new security-scanner execution path and merges its results into the main `agentic` findings flow, which could affect validation/analysis behavior and output schema handling. Subprocess/env/path handling changes (PYTHONPATH/cwd/venv selection) may be brittle across environments despite added regression tests.
> 
> **Overview**
> Adds a new `/openant` workflow end-to-end: `raptor.py openant` + `raptor_openant.py` runner + `libexec/raptor-openant` wrapper, with Claude command docs/allowlist updates.
> 
> Introduces `packages/openant` to locate `openant-core`, run OpenAnt in an isolated subprocess (validated `PYTHONPATH`, `cwd` isolation, venv Python selection, stderr persistence), translate `pipeline_output.json` into Raptor finding schema, and deduplicate against SARIF.
> 
> Extends `raptor_agentic.py` with `--openant`/`--openant-only` flags to run OpenAnt as Phase 1b and **merge translated OpenAnt findings into `validation/findings.json`** (handling both dict-wrapped and list formats, and fixing the `--openant-only` dict-vs-list crash). Also updates run-metadata inference and moves `Feasibility` import in `packages/llm_analysis/agent.py` to module scope, with a substantial new regression test suite for the integration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 716a2dcb232a483eba889b7a3ab8551cf140d943. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.claude/commands/openant.md
+++ b/.claude/commands/openant.md
@@ -1,0 +1,99 @@
+# /openant — OpenAnt LLM-powered source-code vulnerability scan
+
+Run OpenAnt against a repository to find vulnerabilities using AST analysis and
+per-function LLM reasoning. Unlike Semgrep/CodeQL (pattern matching), OpenAnt
+reads and understands each function in context — catching business-logic flaws,
+authentication bypasses, and subtle injection patterns that static tools miss.
+
+**Execution:** `libexec/raptor-openant --repo <path> [options]`
+
+---
+
+## Prerequisites
+
+Set `OPENANT_CORE` to the `openant-core` directory:
+
+```bash
+export OPENANT_CORE=/path/to/OpenAnt/libs/openant-core
+```
+
+Or pass `--openant-core <path>` directly.
+
+---
+
+## Usage
+
+```
+/openant --repo /path/to/code [options]
+```
+
+### Flags
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--repo <path>` | `$RAPTOR_CALLER_DIR` | Repository to scan (required) |
+| `--model <name>` | `sonnet` | LLM model: `sonnet` or `opus` |
+| `--level <name>` | `reachable` | Depth: `all`, `reachable`, `codeql`, `exploitable` |
+| `--language <lang>` | `auto` | Override language detection |
+| `--no-enhance` | off | Skip OpenAnt enhance phase (faster, less accurate) |
+| `--verify` | off | Enable stage-2 LLM verification pass |
+| `--workers <n>` | `4` | Parallel analysis workers |
+| `--max-findings <n>` | `50` | Cap findings in report |
+| `--openant-core <path>` | `$OPENANT_CORE` | Path to openant-core |
+| `--no-analyze` | off | Skip Raptor LLM analysis of findings |
+
+### Analysis levels
+
+- `all` — every function, regardless of reachability (thorough, expensive)
+- `reachable` — functions reachable from entry points (balanced, recommended)
+- `codeql` — functions flagged by CodeQL dataflow (targeted, cheapest)
+- `exploitable` — only functions already marked exploitable
+
+---
+
+## Supported languages
+
+Python, JavaScript/TypeScript, PHP, Ruby, C/C++, Java, Go, Zig
+
+---
+
+## Output files
+
+| File | Description |
+|------|-------------|
+| `openant_findings.json` | Translated findings in Raptor schema |
+| `openant-report.md` | Human-readable markdown report |
+| `raptor_openant_report.json` | Machine-readable run summary |
+| `openant_scan/pipeline_output.json` | Raw OpenAnt output |
+
+---
+
+## Examples
+
+```bash
+# Quick scan with default settings
+/openant --repo /path/to/myapp
+
+# Deep scan — all functions, verification pass, opus model
+/openant --repo /path/to/myapp --level all --verify --model opus
+
+# Fast scan — skip enhance, limit to 20 findings
+/openant --repo /path/to/myapp --no-enhance --max-findings 20
+
+# Target Python only
+/openant --repo /path/to/myapp --language python
+
+# Use OpenAnt alongside agentic (Semgrep+CodeQL+LLM) workflow
+/agentic --repo /path/to/myapp --openant
+```
+
+---
+
+## Notes
+
+- OpenAnt findings have no line numbers — deduplication with SARIF tools uses
+  `(file, CWE)` as the key.
+- `vulnerable` + `confirmed` findings translate to Raptor `error` level.
+- `vulnerable` without stage-2 confirmation translates to `warning`.
+- `safe` findings are suppressed.
+- Use `--verify` to get stage-2 confirmation (costs ~2× tokens).
--- a/.claude/commands/openant.md
+++ b/.claude/commands/openant.md
@@ -40,7 +40,6 @@ Or pass `--openant-core <path>` directly.
 | `--workers <n>` | `4` | Parallel analysis workers |
 | `--max-findings <n>` | `50` | Cap findings in report |
 | `--openant-core <path>` | `$OPENANT_CORE` | Path to openant-core |
-| `--no-analyze` | off | Skip Raptor LLM analysis of findings |
 
 ### Analysis levels
 
--- a/libexec/raptor-openant
+++ b/libexec/raptor-openant
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+#
+# raptor-openant — thin libexec wrapper around `python3 raptor.py openant`.
+#
+# Lets .claude/commands/openant.md invoke the OpenAnt pipeline via a stable
+# libexec command path that is allowlisted in .claude/settings.json, so the
+# slash command does not trigger a permission prompt for `python3 raptor.py`
+# on every run.
+#
+# Usage:
+#   libexec/raptor-openant --repo <path> [--model sonnet] [--level reachable] [...]
+#
+# All arguments pass through to raptor_openant.py via raptor.py.
+
+set -euo pipefail
+
+# Resolve symlinks so RAPTOR_DIR is correct even when this wrapper is
+# symlinked into ~/bin or similar.
+SCRIPT="$0"
+while [ -L "$SCRIPT" ]; do
+    DIR="$(cd "$(dirname "$SCRIPT")" && pwd)"
+    SCRIPT="$(readlink "$SCRIPT")"
+    [[ "$SCRIPT" != /* ]] && SCRIPT="$DIR/$SCRIPT"
+done
+RAPTOR_DIR="$(cd "$(dirname "$SCRIPT")/.." && pwd)"
+
+if [ ! -f "$RAPTOR_DIR/raptor.py" ]; then
+    echo "raptor-openant: cannot find raptor.py under $RAPTOR_DIR" >&2
+    echo "raptor-openant: (resolved from $SCRIPT)" >&2
+    exit 1
+fi
+
+exec python3 "$RAPTOR_DIR/raptor.py" openant "$@"
--- a/packages/openant/config.py
+++ b/packages/openant/config.py
@@ -1,0 +1,91 @@
+"""OpenAnt integration — path discovery and run configuration.
+
+Discovers the OpenAnt core library via:
+  1. OPENANT_CORE environment variable (explicit path)
+  2. $RAPTOR_DIR/../libs/openant-core  (sibling heuristic)
+  3. raptor_dir/../libs/openant-core   (caller-supplied raptor_dir)
+  4. RuntimeError with clear diagnostic
+
+No sys.path manipulation here; that happens in scanner.py.
+"""
+
+from __future__ import annotations
+
+import os
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Optional
+
+OPENANT_CORE_ENV = "OPENANT_CORE"
+OPENANT_MODEL_ENV = "OPENANT_MODEL"
+OPENANT_LEVEL_ENV = "OPENANT_LEVEL"
+
+_SENTINEL = Path("/does/not/exist")
+_CORE_MARKER = "core/scanner.py"
+
+
+@dataclass
+class OpenAntConfig:
+    core_path: Path
+    model: str = "sonnet"
+    level: str = "reachable"
+    enhance: bool = True
+    verify: bool = False
+    workers: int = 4
+    timeout_seconds: int = 1800
+    language: str = "auto"
+
+    def validate(self) -> None:
+        marker = self.core_path / _CORE_MARKER
+        if not marker.exists():
+            raise RuntimeError(
+                f"OpenAnt core not found at {self.core_path!r}: "
+                f"expected {_CORE_MARKER} to exist. "
+                f"Set OPENANT_CORE to the libs/openant-core directory."
+            )
+
+    @classmethod
+    def from_env(cls, raptor_dir: Optional[Path] = None) -> "OpenAntConfig":
+        core_path = _discover_core(raptor_dir)
+        model = os.environ.get(OPENANT_MODEL_ENV, "sonnet")
+        level = os.environ.get(OPENANT_LEVEL_ENV, "reachable")
+        config = cls(core_path=core_path, model=model, level=level)
+        config.validate()
+        return config
+
+
+def _discover_core(raptor_dir: Optional[Path]) -> Path:
+    explicit = os.environ.get(OPENANT_CORE_ENV)
+    if explicit:
+        return Path(explicit)
+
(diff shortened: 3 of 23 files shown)


raptor_agentic.py
```diff
@@ -735,10 +655,67 @@ def _cleanup_git_temp(p=temp_dir):
...
+                print(f"✓ OpenAnt: {len(openant_extra_findings)} unique finding(s)")
+                # NOTE: do NOT add to total_findings here. total_findings feeds
+                # run_validation_phase which uses it to compute duplicates_removed
+                # = total_findings - unique_sarif_count. Adding OpenAnt findings
+                # before that call inflates the baseline and makes the dedup report
+                # show OpenAnt findings as "duplicates" (they're not SARIF findings).
+                # The count is added to validated_findings in the merge block below.
+        except RuntimeError as e:
+            logger.warning(f"OpenAnt not configured (continuing without it): {e}")
+        except Exception as e:
```
grokjc, line 716: --openant-only swallows scanner failure (blocking)
  ```
  except Exception as e:
      logger.warning(f"OpenAnt scan failed (continuing): {e}")
  ```
  In --openant (additive) mode, continuing is the right call. In --openant-only, the user explicitly opted out of Semgrep/CodeQL, so a silent failure produces an empty report instead of an error. Could we branch on args.openant_only and abort there? Probably worth a regression test mirroring the dict-format ones.

libexec/raptor-openant
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
```
grokjc, line 1: Missing trust-marker block in libexec/raptor-openant (blocking)
  
libexec/raptor-agentic on main (lines 17-24) gained a trust-marker gate after this PR branched:
```  
  if [ -z "${CLAUDECODE:-}" ] && [ -z "${_RAPTOR_TRUSTED:-}" ]; then
      echo "$0: internal dispatch script." >&2
      echo "  Run via 'bin/raptor' instead." >&2
      echo "  Tests / power users: set _RAPTOR_TRUSTED=1 to bypass." >&2
      exit 2
  fi
  ```

Could you add the same block to libexec/raptor-openant before the symlink-walk? Otherwise raptor-openant will be the only orchestrator wrapper openly callable after this lands. Pure copy-paste, no behaviour change for legit callers.

File: .claude/commands/openant.md [hunk 3]
```diff
@@ -0,0 +1,98 @@
...
+Run OpenAnt against a repository to find vulnerabilities using AST analysis and
+per-function LLM reasoning. Unlike Semgrep/CodeQL (pattern matching), OpenAnt
+reads and understands each function in context — catching business-logic flaws,
+authentication bypasses, and subtle injection patterns that static tools miss.
+
+**Execution:** `libexec/raptor-openant --repo <path> [options]`
+
+---
+
+## Prerequisites
```
grokjc, line 12: No OpenAnt install story (blocking)
  
  The integration discovers OpenAnt at $RAPTOR_DIR/../libs/openant-core with its own .venv (tree-sitter bindings live there per scanner.py:140-150), but the PR ships no submodule, no bootstrap script, and the user-facing error is just "Set OPENANT_CORE to a valid libs/openant-core directory". Could .claude/commands/openant.md get a "Prerequisites — installing OpenAnt" block (clone URL + python -m venv + pip install)? Otherwise a fresh user has no way to know what OpenAnt is or where to get it.

packages/openant/translator.py
```diff
@@ -0,0 +1,184 @@
...
+
+def _normalize_path(file_path: str) -> str:
+    return os.path.normpath(file_path).lstrip(os.sep)
+
+
+def deduplicate_with_sarif(
+    openant_findings: list[dict],
+    sarif_findings: list[dict],
+) -> tuple[list[dict], int]:
+    """Remove OpenAnt findings that duplicate SARIF findings.
```
grokjc, line 144: Stale dedup-key docstring + coarse key (blocking)
  
packages/openant/translator.py:144 documents the dedup key as (normalized_file, line_bucket_of_5, cwe_id_str), but _sarif_key/_openant_key (lines 170 and 179) actually compute (file, CWE). Two requests:
  
  1. Either fix the docstring to say "function-blind by design", or
  2. Include metadata["route_key"] in the key — it's already populated unconditionally at line 77.
  
  As is, a single SARIF CWE-78 hit in app.py drops every OpenAnt CWE-78 finding in app.py regardless of function, which works against OpenAnt's main value (per-function semantic catches). The route_key inclusion is two lines.

.claude/commands/openant.md
```diff
@@ -0,0 +1,98 @@
...
+```
+/openant --repo /path/to/code [options]
+```
+
+### Flags
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--repo <path>` | `$RAPTOR_CALLER_DIR` | Repository to scan (required) |
+| `--model <name>` | `sonnet` | LLM model: `sonnet` or `opus` |
```
grokjc, line 35: Provider lock-in is invisible
  
  The scanner hardcodes model to {opus, sonnet}), independent of models.json. A user whose primary is Gemini will silently bill Anthropic on /agentic --openant. Could you add a one-line note in .claude/commands/openant.md flagging that OpenAnt always uses Anthropic?

Or can we do this with any model?

packages/openant/scanner.py
```diff
@@ -0,0 +1,239 @@
...
+    lang = config.language if config.language in _OPENANT_CLI_LANGUAGES else "auto"
+    cmd: list[str] = [
+        python_exe, "-m", "openant",
+        "scan", str(repo_path),
+        "--output", str(out_dir),
+        "--model", config.model,
+        "--level", config.level,
+        "--language", lang,
+        "--workers", str(config.workers),
+        "--no-report",
```
grokjc, line 171: _extract_usage with --no-report
  
  scanner.py:215-229 parses proc.stdout as JSON to read token usage, but the command line passes --no-report (scanner.py:171). Have you confirmed OpenAnt still emits the JSON usage envelope on stdout under --no-report? If not, the path falls back silently to pipeline_stats.costs, which means token usage may be under-reported without any warning.

packages/openant/config.py
```diff
@@ -0,0 +1,91 @@
...
+        core_path = _discover_core(raptor_dir)
+        model = os.environ.get(OPENANT_MODEL_ENV, "sonnet")
+        level = os.environ.get(OPENANT_LEVEL_ENV, "reachable")
+        config = cls(core_path=core_path, model=model, level=level)
+        config.validate()
+        return config
+
+
+def _discover_core(raptor_dir: Optional[Path]) -> Path:
+    explicit = os.environ.get(OPENANT_CORE_ENV)
```
grokjc, line 58: Asymmetric path validation in OpenAntConfig
  
  packages/openant/config.py:58-60:
```  
  explicit = os.environ.get(OPENANT_CORE_ENV)
  if explicit:
      return Path(explicit)
```  
  The two heuristic branches that follow check (candidate / _CORE_MARKER).exists() (lines 65, 70). Could the explicit-env branch get the same marker check? OpenAntConfig.validate() catches it later, but matching all three branches up-front would make a misconfigured OPENANT_CORE fail at the same place as a misconfigured filesystem layout.

packages/openant/translator.py
```diff
@@ -0,0 +1,184 @@
...
+) -> Optional[dict]:
+    # OpenAnt uses "stage1_verdict" in pipeline_output.json
+    verdict = (finding.get("stage1_verdict") or "").lower()
+    level = _compute_level(verdict, finding)
+    if level is None:
+        return None
+
+    location = finding.get("location") or {}
+    cwe_id_raw = finding.get("cwe_id")
+    cwe_str = f"CWE-{cwe_id_raw}" if cwe_id_raw else None
```
grokjc, line 74: cwe_id == 0 dropped by truthy check
  
  translator.py:74, :85, and :131 (_make_finding_id) all use if cwe_id_raw else .... Integer 0 falls through to None/"unknown". CWE-0 isn't a real id but the truthy check is a hygiene smell — could we use is not None?

packages/openant/config.py
```diff
@@ -0,0 +1,91 @@
...
+
+@dataclass
+class OpenAntConfig:
+    core_path: Path
+    model: str = "sonnet"
+    level: str = "reachable"
+    enhance: bool = True
+    verify: bool = False
+    workers: int = 4
+    timeout_seconds: int = 1800
```
grokjc, line 35: No CLI timeout knob
  
  OpenAntConfig.timeout_seconds = 1800 (config.py:35) is hardcoded. With --level all --verify on a large repo, 30 min may not be enough. Could raptor_openant.py and the --openant-* flag family in raptor_agentic.py expose --openant-timeout?

raptor_openant.py
```diff
@@ -0,0 +1,302 @@
...
+                if f.get("message"):
+                    lines.append(f.get("message", ""))
+                    lines.append("")
+                if f.get("snippet"):
+                    lines.append("```")
+                    lines.append(f.get("snippet", ""))
+                    lines.append("```")
+                    lines.append("")
+
+    out_dir.joinpath("openant-report.md").write_text("\n".join(lines))
```
grokjc, line 289: write_text(...) call should pass encoding="utf-8" to avoid locale-dependent output.

raptor_agentic.py
```diff
@@ -757,6 +734,88 @@ def _cleanup_git_temp(p=temp_dir):
...
+                        f"validation/findings.json has unrecognized format "
+                        f"(got {type(existing).__name__}); skipping merge"
+                    )
+        else:
+            # No prior validation/findings.json (--openant-only or SARIF produced no file).
+            # BUG-R-016-VARIANT: must write dict format, NOT a plain list.
+            # Phase 3 (packages/llm_analysis/agent.py:_load_validated_findings) calls
+            # data.get("findings", []) which raises AttributeError if data is a plain
+            # list. All other Raptor validation paths write {"stage":..,"findings":[...]}.
+            (out_dir / "validation").mkdir(exist_ok=True)
```
grokjc, line 806: (out_dir / "validation").mkdir(exist_ok=True) should add parents=True for defensiveness; not strictly needed under the current call chain, but cheap insurance.

grokjc (changes requested):
Repo-root pollution (blocking)
  
The PR adds four scratch .md files at the repo root totalling 1,229 lines: AGENT-LOG.md, FIXES-JUSTIFICATION.md, audit-bug-A-R016-new.md, collisions-23738.md. These look like agent/session process artefacts rather than durable repo docs — anyone landing on the repo cold will be confused by what "BUG-R-016-VARIANT" or an "Agent Coordination Log" mean. Could you drop them and fold any durable substance into the commit messages or a packages/openant/README.md?
  
PR-body privacy claim (blocking)

The PR body says "Repo confirmed private via gh api repos/gadievron/raptor --jq '.private' → true". The repo is currently public ("private": false, "visibility": "public"). Could you correct or drop that line so the description is accurate?

Test gaps

  The 86 new tests are solid coverage of what's there, but a few gaps from the issues above:
  - --openant-only abort-on-failure (issue 3) is untested because the behaviour doesn't exist yet — adding a test alongside the fix would lock it in.
  - cwe_id == 0 boundary (issue 10) — no test exercises the 0 case.
  - vulnerable_code 2000-char truncation (translator.py:89) — no test covers the truncation point.
  - End-to-end: none of the 22 scanner tests drive the real subprocess.run path against a stub python -m openant exiting 0/1/2/timeout. The validation-helper tests cover env/cwd/PYTHONPATH but not exit-code handling. Would a single subprocess-stub test add meaningful confidence here?

Symlinks etc

  A few of the new write sites use the established mkdir(exist_ok=True) + predictable-filename pattern that doesn't check for symlinks (scanner.py:56,107; raptor_openant.py:98,155; raptor_agentic.py:806). The PR isn't introducing this pattern — semgrep and codeql do it too — but two new attack surfaces are worth knowing about:

  1. pipeline_output.json is read back from the predictable path at scanner.py:118-119. A co-resident attacker who can pre-position that file before OpenAnt runs can inject false findings into Phase 3 LLM
  analysis. Worth at least an O_NOFOLLOW-equivalent stat on the parent dir, or moving the OpenAnt subprocess output into a tempfile.mkdtemp() and only copying validated content into out_dir afterwards.
  2. openant.stderr.log (scanner.py:107) is the one entirely new write site. Stderr can leak target-repo paths and command lines if the path is redirected by a pre-positioned symlink. The try: ... except  OSError: pass also masks any failure that would indicate redirection.

  Threat model is local-attacker-with-FS-write under the predicted out_dir parent — i.e., shared analysis boxes and CI environments, not single-user laptops.

grokjc:
Defense-in-depth follow-up after #321 (run-dir safe_io) merged: please replace the mkdir(parents=True, exist_ok=True) calls at packages/openant/scanner.py:56, raptor_openant.py:98, raptor_openant.py:155, and  
  raptor_agentic.py:806 with core.run.safe_io.safe_run_mkdir, matching the pattern now used in raptor.py, raptor_agentic.py, packages/codeql/agent.py, etc. Same 2-line shape:                                     
```                                                                                                                                                                                                                   
  out_dir.parent.mkdir(parents=True, exist_ok=True)
  safe_run_mkdir(out_dir) 
```

gadievron:
I'm going to take this one back to the drawing board for a while.

gadievron:
I'm going to take this one back to the drawing board for a while.

PR #312, opened 2026-05-05
feat: enforce trusted-caller marker on libexec/ scripts

grokjc:
libexec/ scripts are internal dispatch entry points; users invoke bin/raptor or bin/cve-diff, which strip dangerous env vars before dispatching. Direct invocation of libexec/raptor-* bypassed that strip layer. Each libexec script now refuses to run without one of CLAUDECODE, _RAPTOR_TRUSTED, or RAPTOR_DIR set; tests and power users bypass with _RAPTOR_TRUSTED=1.

This is API hygiene, not a security boundary — an attacker with env control can set the marker themselves. The benefit is forcing callers through the documented entry points so accidental invocation fails loudly. The check is inlined in every script rather than factored into a shared helper, both because it runs before sys.path mutation (a PYTHONPATH-shadowed helper could no-op it) and because requiring 24 diffs to weaken the check is harder to slip past review than one. A new test (libexec/tests/test_marker_coverage.py) enforces presence, sentinel comment, and placement.

### DIFF
--- a/bin/cve-diff
+++ b/bin/cve-diff
@@ -39,4 +39,7 @@ fi
 
 export RAPTOR_DIR
 export PYTHONPATH="$RAPTOR_DIR:$RAPTOR_DIR/packages/cve_diff"
+# Trust marker — libexec/ scripts refuse to run without one of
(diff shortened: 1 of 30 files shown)


grokjc:
Needs followup - on it!

PR #313, opened 2026-05-05
Typed plan layer design

unprovable:
Four-piece typed substrate on top of `hypothesis_validation`. All additions are backward-compatible — no existing API was modified, no existing test was touched.
                                                                                                                                  
  ### New modules                                           

  | File | LOC | Purpose |
  | --- | ---: | --- |
  | `packages/hypothesis_validation/verdict.py` | 93 | `verdict_from(evidence, llm_claim)` — the runner's three downgrade rules in one function. `aggregate(...)` for multi-adapter combine. |                                                                    
  | `packages/hypothesis_validation/provenance.py` | 102 | `hash_hypothesis(h)` — stable SHA-256 with sort-keys + recursive whitespace normalisation. `ensure_same_provenance(items)` + `ProvenanceMismatch`. |                                             
  | `packages/hypothesis_validation/iteration.py` | 102 | `IterationStep`, `uncertainty(step)`, `must_progress(prev, curr)`, `IterationStalled` — standalone progress guard, **not wired into the runner**. |                                                
                                                            
  ### Modified modules                                                                                                            
                                                            
  | File | Δ | Change |
  | --- | --- | --- |
  | `packages/hypothesis_validation/hypothesis.py` | +147 | Adds optional structured fields: `source`, `sink`, `flow_steps`, `sanitizers`, `smt_constraints` (plus `SourceLocation`/`SinkLocation`/`FlowStep` dataclasses). `to_dict` only emits new keys when set. |
  | `packages/hypothesis_validation/result.py` | +13 | Adds `Evidence.refers_to: str = ""` for provenance. `to_dict` only emits when set. |                                                                                                                     
  | `packages/hypothesis_validation/runner.py` | +26 | `_evaluate` refactored to call `verdict_from` instead of inlining the three downgrade rules. Behaviour-preserving. |                                                                                       
  | `packages/hypothesis_validation/__init__.py` | +29 | Re-exports the new surface alongside the existing one. |
                                                                                                                                  
  ### Tests                                                 
                                                                                                                                  
  | File | LOC | Change |                                   
  | --- | ---: | --- |
  | `packages/hypothesis_validation/tests/test_typed_layer.py` (new) | 422 | 42 tests across the four areas, including 4 runner-behaviour-preservation cases that lock the refactor to the pre-existing semantics. |

All 135 pre-existing tests (`test_security.py`, `test_runner.py`, `test_models.py`, `test_adapters.py`, `test_adapters_codeql_smt.py`) pass untouched. **Total: 177 / 177 passing.**

### Architectural invariants preserved                                                                                          
                                                            
  | Invariant | Where it lives now |
  | --- | --- |
  | Tool failure → inconclusive | `verdict_from` |                                                                                
  | `claim=confirmed` with no matches → refuted | `verdict_from` |
  | `claim=refuted` with matches → inconclusive | `verdict_from` |                                                                
  | Forged-tag neutralisation | `runner._neutralize_forged_tags` (unchanged) |                                                    
  | Sandbox + safe-env defaults | adapters (unchanged) |
                                                                                                                                  
  The runner's behaviour against `test_security.py` is byte-identical to the pre-refactor version.


unprovable:
An aside - the `if x: raise IterationStalled(...)` pattern in `must_progress` is a Hoare-style postconditional. He died recently, so a note to say RIP [Prof. Hoare](https://en.wikipedia.org/wiki/Tony_Hoare)

grokjc:
Suggestion: wire provenance into runner.validate
  
  Evidence.refers_to is added and hash_hypothesis() exists, but runner.validate doesn't compute the hash and pass it when constructing the Evidence object. So every Evidence today has empty refers_to and the feature is dormant.
  
In runner.validate, just after the ```tool_evidence = adapter.run(...)``` call:
```  
  from .provenance import hash_hypothesis
  
  evidence_record = Evidence(
      tool=tool_evidence.tool,
      rule=tool_evidence.rule,
      summary=tool_evidence.summary,
      matches=tool_evidence.matches,
      success=tool_evidence.success,
      error=tool_evidence.error,
      refers_to=hash_hypothesis(hypothesis),
  )
```  
  Plus one test asserting result.evidence[0].refers_to == hash_hypothesis(h). Keeps the feature live rather than waiting on iteration.

grokjc:
Pin hash semantics for None vs empty default
  
to_dict() skips structured fields when None but emits them with default values when set, so Hypothesis(source=None) and Hypothesis(source=SourceLocation()) hash differently. Probably correct (explicit empty is meaningfully different from omitted), but it's a sharp edge.
  
Add a test in test_typed_layer.py that pins this behaviour both ways:
```
  def test_hash_distinguishes_none_from_empty_source():
      h1 = Hypothesis(claim="x", target=Path("/x"))                       # source=None
      h2 = Hypothesis(claim="x", target=Path("/x"), source=SourceLocation())  # source=empty
      assert hash_hypothesis(h1) != hash_hypothesis(h2)

  def test_hash_stable_across_field_addition_when_omitted():
      # Adding optional structured fields without populating them must
      # NOT change the hash of an existing hypothesis.
      h = Hypothesis(claim="x", target=Path("/x"))
      expected = hashlib.sha256(
          json.dumps(
              {"claim": "x", "target": "/x", "target_function": "",
               "cwe": "", "suggested_tools": [], "context": ""},
              sort_keys=True, separators=(",", ":"),
          ).encode("utf-8")
      ).hexdigest()
      assert hash_hypothesis(h) == expected
```

The second test catches accidental "always emit" changes to to_dict() that would invalidate every existing stored hash.

grokjc:
The whole point of hash_hypothesis is process-independence. The current tests probably check that two calls in the same process produce equal hashes; what we want is a hash that's stable across Python versions and machines. One way to pin this:
  
```
  def test_hash_is_byte_stable():
      """Pin the hash for a fixed hypothesis so changes to canonicalisation
      are caught. If this fails, every stored Evidence.refers_to in the
      wild becomes invalid — bump the version explicitly rather than
      silently changing the hash."""
      h = Hypothesis(claim="foo bar", target=Path("/x"), cwe="CWE-129")
      assert hash_hypothesis(h) == "<actual hash from one run, copied in here>"
```  
  
Treat that hex string as a contract. If anyone changes to_dict or _normalise, this test fails loudly and the PR has to address the migration story.

unprovable:
This commit adds the suggestion to wire provenance into runner.validate - runner.py validate() now stamps refers_to=hash_hypothesis(hypothesis) on every Evidence it constructs:
- The main path after adapter.run(...) (the case in your suggestion)
- The empty-rule early return as well, so the audit trail is consistently provenanced

Workig on the other two!

unprovable:
ok these two should shape the code to do what you suggested @grokjc  - I especially like hashing specific workflows. :)

unprovable:
Ugh - just realised; one portability caveat: os.path.normpath("/x") returns "/x" on POSIX but "\\x" on Windows, so this exact hex would mismatch if the test ran on Windows. Is this a blocker @grokjc ?

grokjc:
Much as I would question the sanity of trying to run raptor on Windows, Claude is right ...  os.path.normpath is platform-dependent. On Windows it converts forward slashes to backslashes:                                                                                                                   
  - Linux: os.path.normpath("src/foo.c") → "src/foo.c"
  - Windows: os.path.normpath("src/foo.c") → "src\\foo.c"                                                                                                                                                          
                                                            
  So a hypothesis hashed on a Linux dev machine and the same hypothesis hashed on a Windows CI runner will produce different hashes — which breaks the "stable across machines" contract that's the whole point of the hash.                                                                                                                                                                                                        
                                                                                                                                                                                                                   
  Fix is simple:                                                                                                                                                                                                 
  ```                                                          
  import posixpath                                                                                                                                                                                                 
  \# ...                                                     
  out = posixpath.normpath(out)
```

  posixpath.normpath always uses / as the separator regardless of host OS. Same path-collapse semantics, but platform-independent.                                                                                 
   
  Worth a regression test that pins the expected behaviour:                                                                                                                                                        
  ```                                                          
  def test_hash_path_normalisation_is_platform_independent():                                                                                                                                                      
      """The hash of a forward-slash path must not depend on the host OS.                                                                                                                                          
      
      os.path.normpath converts to backslashes on Windows; posixpath.normpath                                                                                                                                      
      does not. Hashes are stored cross-machine, so platform-dependent
      normalisation breaks Evidence.refers_to lookups."""                                                                                                                                                          
      h = Hypothesis(claim="x", target=Path("src/./foo.c"))                                                                                                                                                        
      # The hash must contain forward slashes in its canonical form,                                                                                                                                               
      # regardless of where this test runs.                                                                                                                                                                        
      expected_hash = hash_hypothesis(                                                                                                                                                                             
          Hypothesis(claim="x", target=Path("src/foo.c"))                                                                                                                                                          
      )                                                     
      assert hash_hypothesis(h) == expected_hash                                                                                                                                                                   
```

unprovable:
Got my Tin Frenchman to do it - here's the summary: 

`provenance.py` — replaced os.path.normpath with posixpath.normpath. The module docstring now explains why (Windows would rewrite / to \\, breaking cross-machine hash stability). The _normalise_string docstring carries the same note inline so anyone touching that function sees the constraint.                                                                                             

`tests/test_typed_layer.py` — new test_hash_path_normalisation_is_platform_independent:                                           
  - Asserts Hypothesis(target=Path("src/./foo.c")) and Hypothesis(target=Path("src/foo.c")) hash the same.
  - Re-derives the expected digest with an explicit forward-slash target in the payload — if the implementation ever swaps in \\, the hash will diverge from this hand-computed expected value, catching the regression directly rather than only via the byte-stable pin.                                                                                                                

The byte-stable hash pin (066b6bcb…) was unaffected because the dev machine is POSIX. If anyone ever adds Windows CI, that pin and the new platform-independence test together form the contract: identical hash on every host.

unprovable:
Ok this was in the todo list for a while and actually fits in very nicely, if my bot does say so itself:  

Tl;dr - Bayesian posterior checking. No, not the kind in Soho on a Saturday night; Beta-Bernoulli verdict aggregation.                                                                                                                
                                                                                                                                 
packages/hypothesis_validation/posterior.py (new, ~140 lines, stdlib only):
  - Posterior(alpha, beta) — frozen dataclass, .mean and .strength properties.                                                   
  - UNIFORM_PRIOR — Beta(1, 1), the no-prior-knowledge baseline.                                                                 
  - update(p, *, confirms, weight=1.0) — single conjugate update step.
  - posterior_from(evidence_list, prior=UNIFORM_PRIOR) — aggregate over an Evidence/ToolEvidence list. Mechanical reading: matches → confirm, no matches → refute, failure → no-op.                                                                       
  - verdict_from_posterior(p, *, confirm_threshold=0.8, refute_threshold=0.2) — threshold projection back to the 3-valued ladder for callers that still want a label.                                                                                           

Re-exported from __init__.py alongside the other typed-layer pieces. (update is exposed as posterior_update to avoid colliding with stdlib namespace if anyone does from packages... import update.)                                                          
                                                                                                                                 
Tests (22 new, four classes):                                                                                                  
  - TestPosterior — arithmetic + invariants (frozen, weight scaling, degenerate strength=0 guard).                               
  - TestPosteriorFrom — empty/all-confirm/all-refute/mixed/failure-noop/strength-monotone/custom prior/Evidence dataclass compat.
  - TestVerdictFromPosterior — threshold semantics + tunability + invalid-threshold validation.                                  
  - TestPosteriorPreservesSignal — the headline property: a 3-confirm-1-refute distribution and a 1-each balanced distribution both map to "inconclusive" under aggregate, but the posterior cleanly distinguishes them by mean and strength.

Waht is not here:                                                                                                  

  - No runner integration. `validate()` still returns a 3-valued verdict. The standalone module is callable from outside; wire it into `ValidationResult` only after a real caller asks for it (matches the must_progress pattern).                                
  - No per-adapter weight calibration. Default weight = 1.0; the docstring flags weight as a tuning knob. Calibration against ground-truth corpora is the natural next step but needs corpora.                                                               
  - No information-theoretic stopping yet. Step 2 from the math plan, depends on this. The entropy of Beta(α, β) is closed-form (`scipy.stats.beta.entropy`, or the math-by-hand version with lgamma) — but that's the next conversation.                        
                                                                      
  The headline test (`test_aggregate_collapses_where_posterior_does_not`) is the one to read first if you want to see the win in one assertion. It pins the exact case where the legacy lattice loses signal that the posterior preserves.

grokjc:
Ohh, I love bonus maths :smile: LGTM but noting that we need to calibrate the likelihoods for the Bayesian aggregation at some point in the not-too-distant future.

unprovable:
Yeah I'm looking at some codebases now to see how best to calibrate it, but it's such an obvious shoe in for hypothesis testing like this, and should help reduce time and improve accuracy to boot. 👨🏻‍🏫

unprovable:
Final note for anyone reading this in the future - here's a reasonable writeup on beta-Bernoulli Bayesian inference: https://gregorygundersen.com/blog/2020/08/19/bernoulli-beta/

PR #324, opened 2026-05-06
feat(packages/code_understanding): real LLM dispatch for hunt and trace

grokjc:
Wires up the multi-model substrate consumers from PR2a with actual LLM dispatch via core.llm.tool_use.ToolUseLoop. Adds the libexec entry point, the /understand skill update so multi-model mode is opt-in via --model, and operator-facing observability (cost reporting, verbose tool-call tracing).

New components:
- packages/code_understanding/dispatch/tools.py Sandboxed Read/Grep/Glob handlers with path-traversal defense, output capping, deterministic walk order, DoS-resistant per-line reads.
- packages/code_understanding/dispatch/_tool_specs.py Shared ToolDef factory: hunt and trace expose identical Read/Grep/Glob descriptions to the model.
- packages/code_understanding/dispatch/hunt_dispatch.py default_hunt_dispatch — runs one ToolUseLoop per model with submit_variants terminal tool.
- packages/code_understanding/dispatch/trace_dispatch.py default_trace_dispatch — same shape, submit_verdicts terminal tool.
- packages/code_understanding/prompts/{hunt,trace}_system.py System prompts. CRITICAL: Tool-first behaviour preamble (E2E-required).
- libexec/raptor-understand CLI entry point with --verbose for per-turn tool-call tracing. Invoked from .claude/commands/understand.md when the user passes --model.

PR2a protocol fix: TraceDispatchFn extended (model, traces) → (model, traces, repo_path). Original 2-arg form was based on wrong assumption that traces don't need codebase access.

Hardening (across multiple adversarial review rounds):
- Path traversal: rejects absolute paths, resolves+checks symlinks
- DoS: bounded file reads, per-line caps, file-size skip in grep
- Determinism: sorted walk so cap-truncation produces stable result sets
- LLM type drift: filters malformed variants/verdicts at dispatch boundary
- Provider failures: caught with diagnostic context preserved
- Wall-clock cap: 600s per model
- Argparse validation: positive_int / positive_float / nonempty_str types

E2E-discovered fixes (real LLM smoke test):
- System prompts require tool-first behaviour. Without this, Gemini's JSON-fallback path could narrate intent in plain text and terminate the loop without any tool calls. Reproduced reliably; fixed.
- libexec output surfaces per-model error text and per-model cost, removing the need to dig into substrate.per_model_raw for debugging.
- New --verbose flag prints turn-by-turn tool calls to stderr so operators can self-debug a model that's burning budget or going off-script.

Real LLM smoke test (Gemini 2.5 Pro):
- Hunt: 2/2 strcpy variants found, correctly skipped bounded strncpy and out-of-scope format-string bug. Cost: $0.0163.
- Trace: correct verdicts including refuting a bad hypothesis. $0.083.

178 PR2 tests; full repo suite (6314) green.

### DIFF
--- a/.claude/commands/understand.md
+++ b/.claude/commands/understand.md
@@ -11,12 +11,50 @@ It is a work in progress, remember that.
 ## Usage
 
 ```
-/understand <target> [--map] [--trace <entry>] [--hunt <pattern>] [--teach <subject>] [--out <dir>]
+/understand <target> [--map] [--trace <entry>] [--hunt <pattern>] [--teach <subject>]
+                     [--out <dir>] [--model <name> ...]
 ```
 
 If no mode flag is given, default to `--map`.
 
+### Multi-model mode (opt-in)
+
+For `--hunt` and `--trace`, you can pass one or more `--model` flags to
+run independent analyses across multiple LLMs and correlate the results.
+Each model produces its own findings; the substrate identifies items
+where models agree (high confidence) vs. disagree (worth a closer look).
+
+```
+/understand <target> --hunt "<pattern>" --model claude-opus-4-7 --model gpt-5
+/understand <target> --trace traces.json --model claude-opus-4-7 --model gpt-5
+```
+
+**When to dispatch to libexec instead of running in-session:** if the
+user passes `--model` AND the mode is `--hunt` or `--trace`, you MUST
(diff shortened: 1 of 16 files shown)


grokjc:
Need to fix tests (unrelated)

PR #325, opened 2026-05-06
feat(exploit_feasibility): attack tree probability propagation

unprovable:
Replaces flat additive confidence adjustments with proper AND/OR probability composition through the primitive dependency graph. The correct propagation for a tree of primitives should be:

```
  AND-node:  P(success) = Π P(childᵢ)
  OR-node:   P(success) = 1 − Π (1 − P(childᵢ))
  BLOCK:     P(success) = P(node) × (1 − P(mitigation_active))
```

Key design decision fixed: vulnerability-type nodes return P=0 unless explicitly in starting_primitives. They represent scanner findings, not theoretical possibilities — without that guard, the tree was computing P≈1 with no input at all.

- `_MITIGATION_FACTORS`: multiplicative attenuation per active complication (derived from ConfidenceAdjustment)
- `_atp_node` / `_atp_capability`: recursive attack tree evaluation. Vulnerability-type nodes require explicit confirmation in `starting_primitives` (P=0 otherwise — they are findings, not hypotheticals). Cycle guard via frozenset call-stack tracking.
- `attack_tree_probability(goal, starting_primitives)`: public method computing P(goal) over the full dependency graph.
- `ExploitPath.p_success`: path-specific probability populated by `_build_exploit_path` using the same multiplicative model. Paths sorted by `p_success` instead of raw `total_reliability`.

By files changed: 

`graph.py`:                                                                                                                            
- `_MITIGATION_FACTORS` — maps each MitigationID to a multiplicative factor (STACK_CANARY → ×0.90, POINTER_MANGLING → ×0.85, etc.), derived from the existing ConfidenceAdjustment constants                                                                            
- `_atp_capability()` — resolves a requirement name to its probability, handling both primitive names and named capabilities via the provider map                                                                                                                        
- `_atp_node()` — recursive core; AND over requires, OR over requires_any, multiplicative mitigation attenuation at each leaf; cycle guard via immutable frozenset call-stack                                                                                            
- `attack_tree_probability(goal, starting_primitives)` — public API: "given these confirmed vulns, what's P(code_execution)?"

`primitives.py`:                                                                                                                       
- `ExploitPath.p_success: Optional[float]` — correctly computed path probability, shown in `summary()` in place of the old raw reliability percentage                                                                                                              
                                                                                                                                      
Tests were also added but these might need expanding for coverage or more scope checking. 17 new tests covering unit-interval invariants, AND/OR composition, mitigation attenuation, cycle safety, and `p_success` on paths are in.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -2,7 +2,8 @@
   "env": {
     "CLAUDE_ENV_FILE": ".claude/raptor.env",
     "CLAUDE_BASH_MAINTAIN_PROJECT_WORKING_DIR": "1",
-    "BASH_DEFAULT_TIMEOUT_MS": "300000"
+    "BASH_DEFAULT_TIMEOUT_MS": "300000",
+    "CLAUDECODE": "1"
   },
   "permissions": {
     "allow": [
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -2,8 +2,7 @@
   "env": {
     "CLAUDE_ENV_FILE": ".claude/raptor.env",
     "CLAUDE_BASH_MAINTAIN_PROJECT_WORKING_DIR": "1",
-    "BASH_DEFAULT_TIMEOUT_MS": "300000",
-    "CLAUDECODE": "1"
+    "BASH_DEFAULT_TIMEOUT_MS": "300000"
   },
   "permissions": {
     "allow": [
--- a/packages/exploit_feasibility/graph.py
+++ b/packages/exploit_feasibility/graph.py
@@ -25,6 +25,10 @@ from .primitives import (
 # Multiplicative success-rate factors per active mitigation, applied at each
 # primitive whose `complicated_by` list includes that mitigation.
 # Derived from ConfidenceAdjustment constants: factor = 1 + delta/100.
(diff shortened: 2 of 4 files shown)


grokjc:
Nice, solid addition.

A few things need tweaking, in priority order:

  1. Drop the .claude/settings.json hunk - how did that get in there? 😃 
  2. Commit message says "replaces" additive model but it's actually parallel (compute_confidence() still runs). 
  3. find_paths_to_goal sort fallback is dead (p_success always set) — simplify to key=lambda p: p.p_success or 0.0.
  4. _MITIGATION_FACTORS covers 8 of ~20 mitigations referenced in complicated_by; rest fall through to silent 0.95 default — document or make explicit.                                                           
  5. Two different p_success semantics share a name (path = chain product, tree = recursive AND/OR); add one-line docstring distinguishing them.                                                                   
  6. Branch is far behind origin/main — rebase recommended.

On that last point I'm just about to land a huge amount of churn on main - you might want to wait ~15 mins 😄

unprovable:
Oh wuaw you were busy! Ok, I think most of the points are covered? Let me know if something else is needed :)

grokjc:
All done I think! Thanks - LGTM

PR #336, opened 2026-05-06
[security] fix(web): redact crawler URL artifacts

Hinotoi-agent:
## Summary
- Redact URL-embedded secrets from crawler results returned by `WebCrawler.get_results()` before they are persisted/shared
- Redact crawler log messages that include target URLs, including URL text embedded in exception messages
- Keep internal crawl state raw so scope checks and discovery behavior are unchanged
- Preserve exact target URLs when the operator explicitly uses `reveal_secrets=True`

## Why
Crawler results can be written to artifacts such as `crawl_results.json`. If target URLs contain sensitive query or fragment values, those values may otherwise leak into shareable scan outputs and logs. This keeps the default artifacts safer while preserving a deliberate debug path for operators who need exact URLs.

## Notes
- This is limited to target/web artifact and log redaction.
- Non-sensitive URL context such as host/path and ordinary query parameters remains visible for evidence quality.
- No broad scanner behavior or crawl scope semantics are changed.
- `BeautifulSoup` is imported lazily in the HTML parsing path so crawler redaction tests do not require the optional HTML parser dependency during collection.

## Validation
- `uv run ruff check packages/web/crawler.py packages/web/tests/test_secret_redaction.py`
- `uv run python -m pytest core/security/tests/test_redaction.py packages/web/tests/test_secret_redaction.py packages/web/tests/test_client_scope.py packages/web/tests -q` — 55 passed
- `git diff --check`
- Added-line scans for obvious committed secret literals and dangerous execution sinks


grokjc:
PR #336 Review — [security] fix(web): redact crawler URL artifacts                                                                                                                                               
   
  What it does: Wraps URL-bearing fields in WebCrawler.get_results() and crawler log messages with redact_url_secrets_only (already used by WebClient/WebFuzzer). Honors WebClient.reveal_secrets for opt-out. Adds
   4 tests covering default redaction, reveal mode, log paths, and exception messages.
                                                                                                                                                                                                                   
  Solid points:   
  - Reuses existing helper consistently — same pattern as the rest of packages/web/. Coherent.
  - Internal state (self.discovered_urls, etc.) stays raw, so urlparse().netloc scope checks aren't broken. Good separation.                                                                                       
  - Tests cover OAuth-fragment tokens (relies on the fragment fix in 5f6b2ef, already on main).                             
  - 12/12 pass; no regressions in test_secret_redaction.py.                                                                                                                                                        
                                                                                                                                                                                                                   
  Gap to flag:                                                                                                                                                                                                     
                                                                                                                                                                                                                   
  1. _redacted_form skips inputs. Only action and page_url are redacted. inputs is shaped like {"csrf_token": {"type": "hidden", "value": "xxx"}, "api_key": {"type": "hidden", "value": "yyy"}} — pre-filled      
  hidden inputs leak verbatim. CSRF tokens, API keys, OAuth state in hidden fields all bypass this PR. Either redact values when the input name is in _SECRET_QUERY_KEYS (or name.lower().endswith(("_token",      
  "_secret", "_key"))), or document explicitly that input values are out of scope and file a follow-up.                                                                                                            
                  
  Nits:                                                                                                                                                                                                            
   
  2. Behavioral change unannounced: discovered_parameters switched from list(...) to sorted(...), and URL lists are now sorted via _redacted_url_list. Deterministic output is good, but it's not in the commit    
  message — worth one line.
  3. Lazy BeautifulSoup import is fine but unrelated to the security fix; mention in the commit message ("incidentally moves bs4 import to lazy") so reviewers don't grep for the rationale.                       
  4. _redact_message_for_log(e) redacts str(e) — works for the test cases. If a library exception carries the URL in a non-__str__ attribute (response.url, e.args[1] as a tuple), only the default rendering gets 
  cleaned. Worth a one-line docstring caveat.

Hinotoi-agent:
Updated this PR with the follow-up fixes from all review comments and the CodeQL log-sink follow-up.

Current head: `05904bc`

Addressed comments:

- **Hidden/pre-filled form inputs**: `3695517` redacts sensitive pre-filled form input values in crawler artifacts, including hidden `csrf_token`, `api_key`, and OAuth `state`/`nonce`/`csrf` style fields, while preserving them under explicit `reveal_secrets=True`.
- **Deterministic artifact ordering**: URL artifact lists and `discovered_parameters` are emitted deterministically, so generated crawl output is stable across runs.
- **Lazy BeautifulSoup import**: the crawler keeps the `bs4` import lazy inside HTML processing so test collection does not require BeautifulSoup unless the HTML parser path is exercised.
- **Exception-message caveat**: crawler exception logs no longer include exception message text because CodeQL still treats URL-bearing exception strings as sensitive logging sinks. The current behavior intentionally logs the exception type plus a safe page label; persisted artifacts keep redacted URL context for operator review.
- **Log-information loss from host-only labels**: `75fa63b` / `05904bc` replaced host-only logging with a per-page label: the non-secret target origin plus `page_id=page-000N`. That makes `Max depth`, `Crawling`, non-200, parser, and JS endpoint messages correlatable per page without logging URL path/query/fragment text.
- **Concatenated secret field names**: `ecea44d` extends explicit secret-field matching for no-separator names such as `accessToken`, `clientSecret`, and `refreshToken`, using the safer explicit-list approach suggested in review.
- **CodeQL follow-up**: after redacted URL logs were still flagged, `05904bc` avoids both URL logging and hashing sensitive URL material by assigning per-crawl page IDs instead of SHA-256-derived labels.

Local validation:

```bash
uv run pytest core/security/tests/test_redaction.py packages/web/tests/test_secret_redaction.py packages/web/tests/test_scanner_none_llm.py packages/web -q
# 60 passed

uv run ruff check core/security/redaction.py packages/web/crawler.py packages/web/tests/test_secret_redaction.py
uv run ruff format --check core/security/redaction.py packages/web/crawler.py packages/web/tests/test_secret_redaction.py
uv run python -m compileall -q core/security packages/web *.py
git diff --check
```

Also reran added-line scans over the touched files for obvious secret literals and dangerous execution sinks; no findings.

GitHub checks are green on `05904bc`, including CodeQL, and the PR is currently mergeable (`mergeStateStatus: CLEAN`).

grokjc:
A few followup comments:

**Log-information loss from _target_log_label()**                                                                                                                                                                 
                                                                                                                                                                                                                   
  Every crawl log line now collapses to the bare scheme://host of client.base_url, regardless of which page is actually being processed. After this PR an operator tailing logs sees:                              
                                                                                                                                                                                                                   
  Crawling: https://example.com (depth=2, pages=17)                                                                                                                                                                
  Non-200 response for https://example.com: 404                                                                                                                                                                    
  Max depth reached for https://example.com
                                                                                                                                                                                                                   
  The "Max depth reached for ..." line is now actively misleading — it reports the base URL, not the deep URL that tripped the limit. Same for non-200 responses: you can no longer tell which path 404'd.         
   
  Suggested lighter-touch alternative:                                                                                                                                                                             
 ```                                                                                                                                                                                                                
  logger.info(f"Crawling: {redact_url_secrets_only(url)} (depth={depth}, pages={len(self.visited_urls)})")
  ```                                                                                                                                                                                                                 
  Keeps path + non-secret query params, scrubs only the secret params. The redactor was built for exactly this.                                                                                                    
                                                                                                                                                                                                                   
  **Exception messages reduced to type(e).__name__**                                                                                                                                                                
```                                                                                                                                                                                                                 
  logger.warning(f"Error crawling {self._target_log_label()}: {type(e).__name__}")                                                                                                                                 
```                                                                                                                                                                                                                   
  This drops everything from e.args — the network failure reason, the BS4 parse complaint, the SSL handshake detail, etc. Two log sites where this hurts debugging:                                                
                                                                                                                                                                                                                   
  - _crawl_recursive — error type only, no cause                                                                                                                                                                   
  - _process_html_response — error type only, no parser context                                                                                                                                                  
  - _process_json_response — same                                                                                                                                                                                  
                  
  Same redactor fix works:                                                                                                                                                                                         
 ```                 
  logger.warning(f"Error crawling {redact_url_secrets_only(url)}: {redact_url_secrets_only(str(e))}")                                                                                                              
  ```                                                                                                                                                                                                                 
  **is_secret_field_name separator-only matching**
```                                                                                                                                                                                                                   
  _SECRET_FIELD_SUFFIXES = ("_token", "-token", "_secret", "-secret", "_key", "-key")                                                                                                                              
 ```                                                                                                                                                                                                                  
  Catches api_key, api-key, apiKey (after lowercasing → apikey — no match). Misses concatenated forms: apikey, accesstoken, clientsecret, refreshtoken. The explicit _SECRET_QUERY_KEYS set covers the common      
  literal names, so this is a long-tail issue, but worth either:                                                                                                                                                   
                                                                                                                                                                                                                   
  - adding the no-separator forms to _SECRET_QUERY_KEYS, or
  - changing the suffix tuple to also match without the separator ("token", "secret", "key") — though that has its own false-positive risk (monkey, donkey, ...).
                                                                                                                                                                                                                   
  Probably the explicit-list approach is safer.

**Comment by Hinotoi-agent** [comment]
@grokjc thanks for the follow-up — I addressed the items you flagged in the latest PR head `05904bc`.

What changed:

- Hidden/pre-filled form values are now redacted when their field names indicate secrets, including `csrf_token`, `api_key`, OAuth `state`/`nonce`/`csrf`, and concatenated/camelCase names such as `accessToken`, `clientSecret`, and `refreshToken`.
- Crawler logs no longer collapse every event to only the base URL. They now use the safe target origin plus a stable per-crawl page label like `page_id=page-000N`, so `Crawling`, non-200, parser, JS endpoint, and max-depth messages remain correlatable without logging URL path/query/fragment text.
- Exception logs intentionally keep the exception type plus the safe page label rather than full `str(e)`, because CodeQL still treats URL-bearing exception strings as clear-text logging sinks. Redacted URL context is still preserved in generated artifacts for operator review.
- Artifact output ordering remains deterministic, and the lazy BeautifulSoup import is kept scoped to the HTML parser path.

Validation:

```bash
uv run pytest core/security/tests/test_redaction.py packages/web/tests/test_secret_redaction.py packages/web/tests/test_scanner_none_llm.py packages/web -q
# 60 passed

uv run ruff check core/security/redaction.py packages/web/crawler.py packages/web/tests/test_secret_redaction.py
uv run ruff format --check core/security/redaction.py packages/web/crawler.py packages/web/tests/test_secret_redaction.py
uv run python -m compileall -q core/security packages/web *.py
git diff --check
```

GitHub checks are green on `05904bc`, including CodeQL, and the PR is currently mergeable.

grokjc:
Thanks! LGTM

PR #281, opened 2026-05-03
feat: hardware-aware resource tuning via tuning.json

grokjc:
- New `core/tuning/` package: loads `tuning.json`, resolves `"auto"` values via hardware detection (RAM, CPUs), validates per-key with fallback to defaults, stat-based cache invalidation                       
  - `libexec/raptor-tune` CLI: `show` / `max` / `balanced` / `default` profiles that rewrite tuning.json with hardware-aware values
  - `RaptorConfig` constants (`CODEQL_RAM_MB`, `CODEQL_THREADS`, `MAX_SEMGREP_WORKERS`, `MAX_CODEQL_WORKERS`) now backed by classproperties that defer to `get_tuning()`                                           
  - `--max-parallel` in `/agentic` defaults from tuning.json; fuzzing planner escalation capped by `max_fuzz_parallel`                                                                                             
  - JSON comment stripping upgraded: string-aware `//` and `#` support (inline + full-line), replacing regex that couldn't handle URLs or trailing comments                                                        
  - `/tune` skill for Claude Code

Issue #276, opened 2026-05-02
Compute Preferences - Thoughts ?


Splinters-io:
# Possible Approaches

## The Gap

RAPTOR's parallelism is hardcoded. Thread counts, worker pools, batch sizes, and concurrency limits are scattered across individual modules as magic numbers. Each tool integration — r2 batch decompilation, AFL++ fuzzing, Semgrep scanning, CodeQL analysis — manages its own parallelism independently with no awareness of what the host can actually provide.

This creates two problems:

**1. Underutilisation by default.** When a module hardcodes a conservative worker count and the host has far more capacity available, the remaining compute sits idle. The operator has to notice low CPU utilisation and manually intervene — adjusting parameters mid-session, restarting with different arguments, or editing module code. The gap between what the hardware can do and what RAPTOR asks it to do is entirely artificial.

**2. No way to throttle down.** When the host is shared or running concurrent workloads, there's no mechanism to reduce RAPTOR's footprint. The choice is the hardcoded defaults or nothing. There's no "use half" option.

The root cause is architectural. `RaptorConfig` defines paths, timeouts, and tool dependencies, but says nothing about how much parallelism to use. There's no single place where compute allocation is defined, read, or enforced.

**Modules with hardcoded parallelism today:**

| Module | What's Hardcoded |
|--------|-----------------|
| r2/ghidra batch decompilation | `ProcessPoolExecutor(max_workers=N)` |
| AFL++ launcher | Thread count per instance, number of parallel instances |
| ~~GPU-fuzz mutator~~ | ~~Worker count, batch size for seed generation~~ |
| Semgrep runner | Relies on Semgrep's own parallelism defaults |
| CodeQL runner | Relies on CodeQL's `--threads` default |

---

## Hypothesis

If RAPTOR had a way to express compute preferences centrally — a single configuration that all modules read from — then:

1. **Default utilisation would match available hardware.** Modules would inherit a resource budget from the host rather than guessing independently.

2. **Resource sharing would be intentional.** An operator could express "use everything" or "leave headroom" as a single decision rather than hunting through per-module constants.

3. **Operational tuning would persist.** When experience shows that decompilation benefits from more workers, or that CodeQL needs RAM headroom, that knowledge would survive across sessions instead of being rediscovered each time.

4. **One-off adjustments would be trivial.** A single flag to override for one invocation, without touching permanent configuration.

[...]

PR #338, opened 2026-05-06
feat(tuning): auto-detect worker limits

Hinotoi-agent:
## Summary

Follow-up to #276 / #281 that lets RAPTOR's CPU-backed worker limits participate in the existing `"auto"` tuning path.

This PR adds auto detection for:

- `max_semgrep_workers`
- `max_codeql_workers`
- `max_fuzz_parallel`

Each resolves conservatively to half of detected CPUs, with a minimum of 1 and a safe fallback of 2 when CPU detection is unavailable.

## Why

#281 added the lightweight tuning layer and `"auto"` resolution, while #276 called out hardcoded worker counts as the practical compute-preference gap. Today, `"auto"` only works for CodeQL RAM/threads; setting worker limits such as `max_semgrep_workers` to `"auto"` falls back to shipped defaults.

This keeps the design lightweight: no env vars, no scheduler, no default behavior change. Existing defaults remain unchanged unless an operator explicitly sets one of these worker limits to `"auto"`.

## Notes

- Existing shipped defaults are unchanged.
- CodeQL native `codeql_threads = "auto"` behavior is unchanged (`0`).
- `max_agentic_parallel` is left explicit because Claude Code agent concurrency is not purely CPU-bound and can involve provider/session limits.
- `raptor-tune` comments now document which worker limits support `"auto"`.

## Test plan

- `uv run python -m pytest core/tuning/tests/test_tuning.py -q`
- `uv run python -m compileall -q core/tuning libexec/raptor-tune`
- `uv run ruff check core/tuning/__init__.py core/tuning/tests/test_tuning.py libexec/raptor-tune`
- `git diff --check`
- added-line scan for obvious secret literals and dangerous execution sinks
- `_RAPTOR_TRUSTED=1 uv run python libexec/raptor-tune`

### DIFF
--- a/core/tuning/__init__.py
+++ b/core/tuning/__init__.py
@@ -11,6 +11,7 @@ never blocks a session.
 from __future__ import annotations
 
 import logging
+import math
 import os
 from dataclasses import dataclass
 from pathlib import Path
@@ -87,8 +88,37 @@ def _detect_fuzz_parallel() -> int:
     return _detect_half_cpu_parallelism()
 
 
+def _detect_cgroup_cpu_quota() -> int | None:
+    """Return an integer CPU quota from Linux cgroups, if configured."""
+    cpu_max = Path("/sys/fs/cgroup/cpu.max")
+    try:
+        quota_text = cpu_max.read_text(encoding="utf-8").strip().split()
+    except OSError:
+        quota_text = []
+    if len(quota_text) >= 2 and quota_text[0] != "max":
+        try:
+            quota = int(quota_text[0])
+            period = int(quota_text[1])
+        except ValueError:
+            quota = period = 0
+        if quota > 0 and period > 0:
+            return max(1, math.ceil(quota / period))
+
+    quota_path = Path("/sys/fs/cgroup/cpu/cpu.cfs_quota_us")
+    period_path = Path("/sys/fs/cgroup/cpu/cpu.cfs_period_us")
+    try:
+        quota = int(quota_path.read_text(encoding="utf-8").strip())
+        period = int(period_path.read_text(encoding="utf-8").strip())
+    except (OSError, ValueError):
+        return None
+    if quota > 0 and period > 0:
+        return max(1, math.ceil(quota / period))
+    return None
+
+
 def _detect_available_cpus() -> int:
-    """Return CPUs available to this process, respecting Linux affinity."""
+    """Return CPUs available to this process, respecting affinity/cgroups."""
+    candidates: list[int] = []
     sched_getaffinity = getattr(os, "sched_getaffinity", None)
     if sched_getaffinity is not None:
         try:
@@ -96,11 +126,19 @@ def _detect_available_cpus() -> int:
         except OSError:
             affinity_cpus = 0
         if affinity_cpus > 0:
-            return affinity_cpus
-    cpus = os.cpu_count()
-    if cpus is None:
+            candidates.append(affinity_cpus)
+
+    cpu_count = os.cpu_count()
+    if cpu_count is not None and cpu_count > 0:
+        candidates.append(cpu_count)
+
+    cgroup_cpus = _detect_cgroup_cpu_quota()
+    if cgroup_cpus is not None:
+        candidates.append(cgroup_cpus)
+
+    if not candidates:
         return 4
-    return cpus
+    return min(candidates)
 
 
 def _detect_half_cpu_parallelism(max_workers: int | None = None) -> int:
--- a/core/tuning/__init__.py
+++ b/core/tuning/__init__.py
@@ -60,9 +60,37 @@ def _detect_threads() -> int:
     return 0
 
 
+def _detect_semgrep_workers() -> int:
+    """Resolve a conservative CPU-based Semgrep worker count.
+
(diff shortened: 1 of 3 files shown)


grokjc:
PR #338 Review — feat(tuning): auto-detect worker limits

  What it does: Lets max_semgrep_workers, max_codeql_workers, and max_fuzz_parallel accept "auto" in tuning.json, resolved to max(1, os.cpu_count() // 2) with a 2-worker fallback when cpu_count() returns None.
  Defaults are unchanged. max_agentic_parallel is intentionally excluded.

  Solid:
  - Opt-in only — existing integer defaults (4/2/4) untouched, so no regression for anyone not editing tuning.json.
  - Plugs cleanly into the existing _AUTO_RESOLVERS mechanism.
  - Test coverage is thorough: per-resolver unit tests + integration tests via load_tuning, including the unknown-CPU fallback path.
  - User-facing _create_default_file comments and libexec/raptor-tune comments both updated to note "(auto = half CPUs)".

  Concerns:

  1. os.cpu_count() ignores cgroup/affinity limits — RAPTOR runs inside a sandbox (core/sandbox/) and inside Docker/devcontainers; os.cpu_count() reports host CPUs, not what the process can actually use. Prefer
  len(os.sched_getaffinity(0)) on Linux with a cpu_count() fallback for macOS. On a 64-core host with the process pinned to 8 cores, current code returns 32 — wrong by 4×.
  2. No upper bound — memory blowout risk for CodeQL. On a 256-core machine, max_codeql_workers resolves to 128. Each CodeQL DB build wants codeql_ram_mb (auto-clamped to ≤16 GB) — 128 × 16 GB = 2 TB. Recommend
  capping, e.g. max_codeql_workers ≤ total_ram_mb // codeql_ram_mb, or a hard ceiling (4–8). Less critical for Semgrep/AFL, but still worth a sane max.
  3. Three identical wrappers — _detect_semgrep_workers, _detect_codeql_workers, _detect_fuzz_parallel all just call _detect_half_cpu_parallelism(). Defensible per the per-tool docstrings (room to diverge
  later), but if you don't expect them to diverge, plug _detect_half_cpu_parallelism directly into _AUTO_RESOLVERS and drop the wrappers.

  Nit: os.cpu_count() or 4 is opaque; if cpus is None: return 2 reads more directly.

Hinotoi-agent:
Thanks for the review — addressed in `798c9eeccb8abc23a40fecd505aae2b15c25f4bc`.

Changes made:
- `"auto"` worker resolution now prefers `len(os.sched_getaffinity(0))` when available, falling back to `os.cpu_count()` and then the existing 2-worker unknown-CPU behavior.
- `max_codeql_workers` is capped by both total RAM vs. resolved `codeql_ram_mb` and a hard ceiling of 8 to avoid very large CodeQL fan-out on high-core hosts.
- Updated `raptor-tune` and generated `tuning.json` comments to describe available-CPU detection and the CodeQL cap.
- Added regression coverage for affinity-aware CPU detection and CodeQL RAM/hard-ceiling behavior.

Validation:
- `uv run ruff check core/tuning/__init__.py core/tuning/tests/test_tuning.py libexec/raptor-tune` ✅
- `uv run pytest core/tuning/tests/test_tuning.py -q` ✅ (`34 passed`)
- `git diff --check` ✅
- `python3 -m compileall -q core/tuning libexec/raptor-tune` ✅
- `uv run python libexec/raptor-tune` under `_RAPTOR_TRUSTED=1` ✅
- GitHub Actions for the pushed head are green ✅

I also ran the configured broad commands. The repo-wide `uv run ruff check .` is still blocked by pre-existing lint debt outside these touched files, and `uv run pytest -q` is still blocked during collection by pre-existing issues (`core/sandbox/tests/test_tracer.py` f-string syntax and `packages/cve_diff` test package import errors).

Hinotoi-agent:
Thanks again — I pushed a follow-up in `e27c50ae923feedd0296f917341aded3210bc5de` to tighten the CPU detection further.

Additional changes:
- Auto worker sizing now also honors Linux cgroup CPU quotas, not just `sched_getaffinity(0)` / `os.cpu_count()`.
- Supports both cgroup v2 `cpu.max` and cgroup v1 `cpu.cfs_quota_us` / `cpu.cfs_period_us`.
- Uses the most restrictive available CPU signal across affinity, host CPU count, and cgroup quota, so container/devcontainer-limited runs do not over-size workers on large hosts.
- Kept the existing unknown-CPU fallback behavior at 2 auto workers after the half-CPU calculation.
- Added regression coverage for cgroup v2 quota, cgroup v1 quota, and the explicit unknown-CPU fallback path.

Validation:
- `uv run ruff check core/tuning/__init__.py core/tuning/tests/test_tuning.py libexec/raptor-tune` ✅
- `uv run pytest core/tuning/tests/test_tuning.py -q` ✅ (`37 passed`)
- `uv run python -m compileall -q core/tuning libexec/raptor-tune` ✅
- `git diff --check` ✅
- `_RAPTOR_TRUSTED=1 uv run python libexec/raptor-tune` ✅
- Added-line secret/sink scans on the touched tuning files ✅
- GitHub Actions on the new head are green ✅

grokjc:
LGTM. Thanks!

PR #372, opened 2026-05-07
feat(security): wrap tool-result content + opt-in refuse on injection

grokjc:
The ToolUseLoop now wraps every ToolResult (success AND error) in an <untrusted-{nonce}> envelope before threading it back into the conversation, so the LLM consistently treats tool-result content as data rather than instructions. Defends against prompt-injection payloads embedded in attacker-controlled content the model fetches via tool calls (target source files, web pages, command stdout, AND handler exception messages — the last being a content-laundering vector via raise ValueError(target_content)).

Wrapping uses the same envelope shape and per-call random nonce as core/security/prompt_envelope.py:_render_nonce_only — existing neutralize_tag_forgery defangs forged </untrusted-FAKE> in the body.

Two-layer defence:

  Layer 1 (always-on): envelope wrapping. The LLM sees a consistent
  "this is data" framing for every tool-result. Real-world test
  against Gemini 2.5 Pro: wrapping alone successfully defended an
  injection asking the model to exfil a marker string.

  Layer 2 (consumer opt-in): refuse_on_indicators=("english", ...).
  When listed corpus names fire on the RAW content, the result is
  replaced with a synthetic is_error=True placeholder before the LLM
  sees it. Validated at construction time against the loaded corpora
  so a typo can't silently disable enforcement. Advisory event fires
  either way so operators see what was filtered.

ToolResultPreflight event surfaces preflight indicators on raw content regardless of whether refuse fires.

Order in the dispatch hook is critical: x-source value discovery extracts strings from RAW content before wrap-replacement, otherwise envelope tokens (untrusted-NONCE, tool-result, tool name) would pollute the discovered-values set and falsely whitelist tool inputs on subsequent calls.

16 new tests covering wrapping defaults, error wrapping (the laundering vector), nonce randomness, tag-forgery neutralisation, advisory event emission, no-blocking semantics, opt-in refuse enforcement (5 sub-cases), persistence in conversation history, and the x-source-on-raw invariant.

E2E verified against real Gemini 2.5 Pro with three scenarios: advisory-only with injection (wrapping held, marker not echoed), refuse-on-english with injection (synthetic error returned, marker filtered before LLM saw it), clean content (no false positives).

### DIFF
--- a/core/llm/tool_use/loop.py
+++ b/core/llm/tool_use/loop.py
@@ -50,6 +50,9 @@ from typing import TYPE_CHECKING, Any
 if TYPE_CHECKING:
     from core.llm.providers import LLMProvider
 
+from core.security.prompt_envelope import wrap_tool_result
+from core.security.prompt_input_preflight import preflight
+
 from .types import (
     CacheControl,
     ContextOverflow,
@@ -68,6 +71,7 @@ from .types import (
     ToolHandlerTimeout,
     ToolLoopResult,
     ToolResult,
+    ToolResultPreflight,
     TurnCompleted,
     TurnResponse,
     TurnStarted,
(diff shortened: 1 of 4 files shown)


grokjc:
Fixing ...

PR #467, opened 2026-05-11
Add Actions workflow to update dependencies inside devcontainer

some-natalie:
This PR adds a Python script and GitHub Actions workflow to update the hard-coded dependencies inside of `.devcontainer/Dockerfile` by opening a PR like a janky Dependabot.  I use scripts like this to reduce toil on maintaining pinned deps.

⚠️ In order for this to work, you'll need to allow Actions to open PRs under "settings -> actions -> general" and scroll down to the checkbox toggle for it.  This can increase the blast radius of a supply chain attack w/i Actions, but it seems like everything's squared away on pinning these too. :)

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -66,7 +66,8 @@ RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
 # SEMGREP (static-analysis, scan, raptor-scan, agentic commands)
 # ============================================================================
 
-RUN pip install --no-cache-dir semgrep==1.161.0
+ARG SEMGREP_VERSION=1.162.0
+RUN pip install --no-cache-dir semgrep==${SEMGREP_VERSION}
 
 # ============================================================================
 # RR DEBUGGER (rr-debugger skill, crash-analysis agents)
@@ -133,12 +134,13 @@ RUN apt-get update \
 # Capture-then-execute the nodesource setup script so a curl failure isn't
 # silently swallowed by the pipeline (curl | bash ignores curl's exit status).
 # Node 22 is the current LTS; Node 20 went EOL April 2026.
+ARG CLAUDE_CODE_VERSION=2.1.138
 RUN curl -fsSL https://deb.nodesource.com/setup_22.x -o /tmp/nodesource_setup.sh \
     && bash /tmp/nodesource_setup.sh \
     && rm /tmp/nodesource_setup.sh \
     && apt-get install -y --no-install-recommends nodejs \
     && apt-get clean && rm -rf /var/lib/apt/lists/* \
-    && npm install -g @anthropic-ai/claude-code@2.1.119
+    && npm install -g @anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}
 
 # ============================================================================
 # JS PACKAGE MANAGERS (yarn replaced apt source removed above)
--- a/.github/scripts/update_devcontainer.py
+++ b/.github/scripts/update_devcontainer.py
@@ -1,0 +1,141 @@
+#!/usr/bin/env python3
+
+"""
+This script updates the dependencies inside the container files.
+
+Sources:
+  SEMGREP_VERSION     - https://github.com/semgrep/semgrep/releases (GitHub releases, pip install)
+  CODEQL_VERSION      - https://github.com/github/codeql-cli-binaries/releases (GitHub releases)
+  CLAUDE_CODE_VERSION - https://github.com/anthropics/claude-code/tags (GitHub tags, npm install)
+"""
+
+import json
+import re
+import urllib.request
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parent.parent.parent
+
+DOCKERFILES = [
+    ".devcontainer/Dockerfile",
+]
+
+# Matches a stable semver tag, optionally v-prefixed (e.g. "v1.2.3" or "1.2.3").
+# Pre-releases like "v1.2.3-rc.1" deliberately don't match — never auto-bump to one.
+SEMVER_RE = re.compile(r"^v?(\d+)\.(\d+)\.(\d+)$")
+
+
+def _fetch(url: str) -> bytes:
+    req = urllib.request.Request(
+        url,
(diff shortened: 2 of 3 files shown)


grokjc:
Can we hold off until SCA lands please. 'latest' isn't necessarily the best/safest/compatible version and I'm 30+ commits deep into that journey at the moment 😄

some-natalie:
> Can we hold off until SCA lands please. 'latest' isn't necessarily the best/safest/compatible version and I'm 30+ commits deep into that journey at the moment 😄

Of course!  I just wanted to bump CodeQL forward myself and figured I'd contribute back.

LMK if you need help or a rubber duck on SCA stuff.

grokjc:
Hey, appreciate the contribution and offer of support. I already have a CI component in the works but have added your devcontainer idea specifically. When the commit lands you will have a co-author credit on that front.  

But you've also led me to ponder the question of can a GH action actually do a 'super-dependabot' upgrade whilst being safe vs potential supply chain attacks? And the answer is sort of, but some of the defences need an LLM to reason (e.g. 'does this thing look like it was backdoored?') - so hard to do in that environment.  To that end I'm working on an advisory model rather than a fully-automated one (with fully auto being a toggle should it be within user risk appetite).

I'd definitely like some assistance once I get my dev branch off my local machine and onto GH. It is a huge area and I imagine there are a lot of things that will need tweaking/improving/iterating.  In the meantime don't be surprised if a few random 'Co-Authored: by ' credits appear in your feed as I implement some of the precursors to your suggestion and they will likely land on main before the rest of the SCA branch.

some-natalie:
> can a GH action actually do a 'super-dependabot' upgrade whilst being safe vs potential supply chain attacks?

There's a lot to unpack there on safety, partly because there's not a fantastic limit on how deep Actions will resolve dependencies (or if there's vendored dependencies) and each different types of Actions all have their quirks.  My approach is usually to try and let Actions agent do the resolution, then look at the delta with binary capability diffs ... which has some limitations too.

I took a quick stab at this a while ago with [malcontent](https://github.com/chainguard-dev/malcontent) from @tstromberg while we were both at Chainguard.  Here's [a workflow](https://github.com/some-natalie/fedora-acs-override/blob/main/.github/workflows/malcontent-actions.yml) and [PR it reviewed](https://github.com/some-natalie/fedora-acs-override/pull/314).  My implementation of it is brittle, but I'm hoping to carve out some time to improve it soon. 😅

I'd imagine that a "super" type upgrade would be equally complex as trying to just tackle the Actions ecosystem.  More food for thought!

edit - thank you, and I'm happy to help, just lmk how

grokjc:
Making progress. Just landed #545 to help with this. Full SCA module waiting in the wings ... not long now 😄

grokjc:
Hi @some-natalie - now that #595 (the SCA behemoth) has landed, I want to close this out properly, because your PR turned into a lot more than one workflow's worth of value and I want to be specific about what survived and where it lives.

 Your update_devcontainer.yml + update_dockerfile() was the proof-of-shape for a general "safe auto-bump" loop. We took the cron → branch → resolve-latest → rewrite → PR-open pattern and generalised it across eight version-pin surfaces instead of just the devcontainer:

  - Python / Debian apt / npm manifests (raptor-sca fix --harden)
  - Dockerfile ARG <NAME>_VERSION= pins
  - Dockerfile FROM image refs
  - inline RUN pip install pkg==N
  - GHA uses: foo@<ref> (tag-pinned and SHA-pinned-with-# was vX-comment)
  - k8s / compose / gitlab-ci image: refs
  - Helm Chart.yaml deps
  - git submodules

Every surface runs the same loop your Dockerfile code prototyped: walk → find pin → upstream-latest lookup → verdict → idempotent in-place rewrite.

The difference from vanilla dependabot/renovate is the verdict gate. Before a bump is applied it's checked against the full SCA signal set with no LLM in the CI path: OSV / KEV / EPSS, yanked-release, license policy, install-hook delta, typosquat / compromised-package feed, maintainer-change, version-diff size, binary-capability-delta (more on that below), and function-level reachability. Only Clean verdicts get written; Review/Block leave the finding for a human.

  **The workflow itself**

  .github/workflows/sca-self-bump.yml — directly descended from your update_devcontainer.yml:

  - Mondays 12:00 UTC, single self-updating PR (--force-with-lease so weekly runs don't stack)
  - --self-test re-runs SCA on the patched tree to confirm findings actually decrease before the patch is emitted
  - No auto-merge — branch protection + human review stays enforced. The Megalodon PPE attack (May 2026, ~5,500 repos) is cited inline as the reason we don't zero-touch this.

  **Where your work is in the tree**

  You're mentinoed via Co-Authored-By: on 17 commits on main. The load-bearing ones:

  - core/upstream_latest/github_releases.py: your _fetch() + release/tag fetchers + the stable-semver pre-release filter + v-prefix-strip (commit body literally reads "Adapted from Natalie Somersall's PR #467")
  - packages/sca/rewriters/dockerfile_arg.py: direct adaptation of your update_dockerfile(): the ^(ARG <NAME>=)(\S+) regex, the idempotent rewrite, the skip-if-unchanged + change-tuple shape
  - the SCA-side ARG-pin scanner: your workflow needed something to know which ARGs to bump; that's the scan-side complement
  - the sca-self-bump.yml workflow shape itself

  And separately, your malcontent / capability-diff suggestion from the thread became a whole feature line, not just a footnote. Rather than bundle malcontent we reused our existing radare2-based binary-understanding substrate to diff the syscall/network/file-IO capability surface of current vs target; a bump that suddenly grows capabilities (network where there was none, new execve-family calls, etc.) escalates the verdict. That's the binary-capability-delta detector + the fingerprint CLI + the CycloneDX capability embedding + the --fail-on-capability-drift gate — all attributed to you, since the idea was yours.

  **Closing this PR**

  I'm closing #467 as superseded by #595 because the equivalents are already wired into the broader bumper and merging the original on top would conflict and undo the wider feature. But the design and the idea both shipped, at considerably larger scope than the original diff. Thank you for the idea, and the inspiration. As for your offer of help, I'm sure there are many bugs lurking here, it was a significant piece of work.

PR #470, opened 2026-05-11
[Feature] Adding Zero Knowledge Proof of Exploit (ZKPoX) 

unprovable:
# Overview

Much of the hype around AI Agents finding bugs stems from the fact that scaled vulnerability disclosure and verification is hard. Part of this is the necessity to share deep technical details in a way that might risk exposure of a critical flaw - something seen as 'too hard' and 'risky'. 

Adding Zero-Knowledge Proofs of eXploit (ZKPoX) allows RAPTOR to find an exploit and prove publicly that the exploit is valid _without_ disclosing the exploit except to those who may be charged with the fix. Allowing such functionality permits the risk to be checked and validated, without a working and potentially dangerous exploit being made public without a fix being in place and available. 

Whilst this may be controversial, it is certainly experimental and is included here as a proof of concept for reviewer deliberation (@grokjc @danielcuthbert @thomasdullien ) over providing a full assured zero knowledge disclosure pipeline. Google recently improved quantum factoring and hid the algorithm in a ZKP in a similar way to show that it was a) done work, and b) not ready for disclosure: https://research.google/blog/safeguarding-cryptocurrency-by-disclosing-quantum-vulnerabilities-responsibly/

Responsible disclosure is changing, and agents that speak the math are probably the right place to trial these methods. Especially those that might, like raptor, find issues at scale. Enclosed is an example, from CVE-2017-9047 (see below) that gives a demonstration of the workflow proposed. 

## What this is

Lands the **ZKPoX MVP** — zero-knowledge proof-of-exploit. Convert a working exploit witness into a publicly-verifiable disclosure bundle: the witness stays private, the proof is public, the vendor gets full details immediately via an age-encrypted envelope, and the public-witness-decrypt time is enforced cryptographically by a Drand time-lock. Anchored in Sigstore Rekor for tamper-evident timestamping.

Design proposal lives in this PR (`docs/proposals/raptor-zkpox-design.md`), spec → implementation has been a single-author build-out over a few weeks; commits track the phase plan from the proposal's §13.

## What's in here

**Rust workspace under `core/zkpox/`:**
  - SP1 guest (RISC-V) — links three freestanding C targets through a shared redzone+pattern primitive
    - Target 01: classic stack BOF (no bound check at all)
    - Target 02: textbook off-by-one (`i <= buf_size`)
    - Target 03: **CVE-2017-9047** — libxml2 `xmlSnprintfElementContent` stale-`len` (real, public, low-CVSS bug, fixed in commit `932cc989`)
  - Host prover (`zkpox-prove`) — `--execute` / `--prove` with `--wrap={core,groth16}`, target dispatch, JSON bench records
  - Verifier (`zkpox-verify`) — standalone CBOR-bundle verification
  - Two violation gadgets: `crash-only` (uniform canary) + `memory-safety::oob-write` (position-varying pattern with structured count/offset outputs)

  **Python orchestration under `packages/zkpox/`:**
  - `prove.run()` wrapper around the host prover
  - CBOR bundle assembly (`bundle.py`), age vendor envelope + Drand tlock (`envelope.py`), Rekor anchoring (`anchor.py`)
  - Pytest coverage (light fast tier + slow integration sweep gated on `RAPTOR_SLOW_TESTS=1`)

  **Command surface:**
  - `/prove-exploit` — produce a disclosure bundle
  - `/verify-exploit-proof` — verify one
  - `.claude/skills/zkpox/SKILL.md` + violation-gadget skill files
  - `tiers/personas/disclosure_engineer.md`

  **Tests:**
  - 40-witness corpus (T01: 13, T02: 12, T03: 15) covering benign / crash / canarymatch-false-negative cases
  - `run-tests.sh` regression harness — 40/40 pass execute-mode on Mac
  - `packages/zkpox/tests/` Python wrapper + slow integration test
  - **New CI job `zkpox-regression`** — path-filtered on `core/zkpox/**` / `packages/zkpox/**`; builds the SP1 cross-compile pipeline and runs the execute-mode sweep on every relevant PR

## Benchmarks (Apple Silicon, single witness, target #1)

  | Mode | Wall | Proof size | Verified |
  |---|---|---|---|
  | `--execute` | 0.06 s | — | n/a |
  | `--prove --wrap=core` | 15–18 s (warm pk) | 2.65 MB | ✓ |
  | `--prove --wrap=groth16` | 17 m 11 s | **1,704 B** | ✓ (4.8 ms) |

  The Groth16 wrap pays a 60× wall-clock penalty for a 1,560× proof-size reduction — the right shape for a CBOR disclosure bundle. Full breakdown in `docs/research/zkpox-phase0-findings.md`.

## Test plan

  - [ ] `python3 -m pytest packages/zkpox/tests/ -v` — fast unit tests pass, regression test correctly skips without `RAPTOR_SLOW_TESTS=1`
  - [ ] `cargo build --release --manifest-path core/zkpox/Cargo.toml` (needs SP1 toolchain: `curl -L https://sp1up.succinct.xyz | bash`)
  - [ ] `bash core/zkpox/test/run-tests.sh` — 40/40 pass
  - [ ] CI job `zkpox-regression` runs on this PR and passes
  - [ ] Skim `docs/proposals/raptor-zkpox-design.md` (§1–3 for the pitch, §7 for the gadget catalogue, §8 for the bundle format)
  - [ ] Skim `docs/research/zkpox-phase0-findings.md` for empirical numbers and recorded architectural decisions

## Scope: what's NOT in this PR

  Explicitly deferred per the design doc's §4 roadmap:
  - **LLVM-IR mode** (the 50–100× cheaper path). Phase 0 evaluated this as multi-week scope; current fidelity model is C-via-FFI with freestanding C cross-compiled into the SP1 RISC-V guest.
  - **EVM gadgets** (v0.2 in the roadmap)
  - **Embedded / Cortex-M** (v0.3)
  - **x86-64 binary-only** (v0.4, gated on GPU pipeline maturity)
  - **Prove-mode in CI.** Execute-mode is the PR gate; `--wrap=core` would add ~11 min/run and `--wrap=groth16` would add hours + 6.2 GB artifact download. Both belong on a nightly / dispatch tier.
  - **Vendor-key registry.** Producers currently hand in an age public key; the curated-mapping question (CISA? FIRST? community PR-driven?) is design proposal §14, open.

## Caveats

⚠️  Phase 1.5: placeholder hashes — not suitable for real disclosure

Bundles produced by this branch contain placeholder verifier_key_hash and harness.hash fields — they hash the literal strings "placeholder-vk-1.5" and "harness-1.5", not the real SP1 verifying key or harness binary.
- python3 raptor.py prove-exploit refuses to write a bundle without --allow-placeholder-hashes. Verify-time prints a loud warning.

**NB** - See `docs/zkpox-scope.md` for the full statement of what this MVP proves and doesn't prove.

- Phase 1.5.x replaces both fields with their real digests (sp1-sdk verifying key + harness binary hash).
- Do not use bundles produced by this merge for real CVE disclosure until 1.5.x lands. This is fine for demos, internal testing, and end-to-end pipeline validation. 

## Open questions worth reviewer input

  1. **Single PR vs. split.** Squash-and-merge means the commit history inside the branch isn't load-bearing, but if you'd rather merge per-phase, I can split — the phases are reasonably independent from 1.4 onward.
  2. **CVE-2017-9047 selection.** Picked deliberately at the small end of the severity distribution (CVSS 5.9, single stale variable across one strcat). If you'd rather see the gadget against a bigger CVE before merging, propose one.
  3. **Prove-mode nightly job.** Not in this PR. Worth adding before merge, or follow-up?

## References
  
  - Design proposal: `docs/proposals/raptor-zkpox-design.md`
  - CHEESECLOTH paper: https://eprint.iacr.org/2022/1223
  - Phase 0 findings + benchmarks: `docs/research/zkpox-phase0-findings.md`
  - CVE-2017-9047 advisory: https://www.cvedetails.com/cve/CVE-2017-9047/
  - libxml2 fix commit `932cc989`: https://gitlab.gnome.org/GNOME/libxml2/-/commit/932cc9896ab41475d4aa429c27d9afd175959d74


unprovable:
### Code-side fixes (`645f323`)

  - **Placeholder hashes refuse to ship silently.** `raptor_zkpox.py` now refuses to write a bundle without
  `--allow-placeholder-hashes`. When the opt-in is on, a loud warning fires on every prove. Phase 1.5.x replaces both hashes with real SP1-derived values; until then the bundle can't masquerade as production-ready.
  - **`gadget_hash` → `gadget_id_hash`.** The field hashes the identifier string, not the gadget source. Renamed to be honest about what it commits to. Phase 1.5.x adds a separate `gadget_code_hash` over the implementation files.
  - **`zkpox-verify --strict`.** Exits non-zero on any DEFERRED check (currently STARK proof verification and Rekor Merkle inclusion). Default off in 1.5 so operators can inspect bundles produced today; becomes the default in 1.5.x. Non-strict, non-JSON runs now print a loud "MODE: structural-only" banner.
  - **Anchor fail-closed on missing `integratedTime`.** Previously substituted `time.time()` on parse failure, which would persist a local-clock value as the canonical "Rekor integrated me at" timestamp. Now raises `AnchorError`.
  - **Envelope/version cross-check in the verifier.** `vendor_envelope.scheme` must start with the supported v1 prefix (or be the no-envelope sentinel). Without this a future v2 envelope could be replayed under v1 framing — AAD mismatch would otherwise only surface on decrypt.

  ### Scope doc (`311bf53`)

  `docs/zkpox-scope.md` covers, in plain English:

  - What this MVP proves (memory-safety violation predicate; the five SP1 public values; what a verifier learns).
  - What this MVP does NOT prove (control-flow hijack, RCE, info-leak with attacker-chosen target, exploit reliability, vulnerability class label — the producer's `vulnerability.class` field is metadata, not committed).
  - What you trust when you trust a bundle (SP1, the compiled guest, age, Drand, Sigstore Rekor, your local clock at anchor time), each with current-MVP caveats spelled out.
  - Failure modes (vendor loses age sk, Drand unavailable, vendor patches early, vendor refuses to patch, sp1-sdk soundness bug, producer lies about `target_id` in plaintext metadata).
  - The redzone gadget's FP/FN profile — `2^-8` for 1-byte overflows, `~2^-112` for 14-byte overflows under uniform-random witnesses, with the adversarial-witness caveat explicit.
  - Roadmap for the next vulnerability classes by name.

danielcuthbert (commented):
Three items worth addressing before merge, just keeping in with how the safety angle of RAPTOR works:

**1. `sys.path` safety rule**

`raptor_zkpox.py` does:
```python
sys.path.insert(0, str(Path(__file__).resolve().parent))
```
The established RAPTOR pattern (enforced in #312) is:
```python
sys.path.insert(0, os.environ["RAPTOR_DIR"])
```
The hard `KeyError` on an unset `RAPTOR_DIR` is intentional, it fails loudly rather than silently falling back to a cwd that might not be the repo root. In practice this script lives at the repo root so the behaviour is equivalent, but it's inconsistent with every other entry point and will silently do the wrong thing if the script is ever symlinked or invoked from a different working directory. Easy one-line fix. shows we care and take this seriously

**2. No explicit `get_safe_env()` in nested subprocess calls**

`packages/zkpox/prove.py` (`subprocess.run(cmd, ...)`) and the `age`/`tle` calls in `packages/zkpox/envelope.py` don't pass `env=RaptorConfig.get_safe_env()`. Functionally this is fine `raptor.py` launches `raptor_zkpox.py` via `_run_with_lifecycle` which already calls `get_safe_env()` at line 485, so the nested processes inherit a clean environment. Buuuut the RAPTOR convention is to be explicit at each subprocess boundary so a future caller that invokes these functions directly (tests, integration scripts) can't accidentally pass a dirty env through. A comment at each `subprocess.run()` call noting the inherited-sanitised-env assumption would also serve as documentation. 

**3. Placeholder hashes worth a CHANGELOG note**

The `verifier_key_hash` and `harness.hash` fields in Phase 1.5 bundles are explicitly placeholders (they hash the string `"placeholder-vk-1.5"` rather than the real SP1 verifying key). This is correctly gated behind `--allow-placeholder-hashes` with loud warnings, and documented in `docs/zkpox-scope.md`. No code change needed here, just flagging it so reviewers are aware that bundles produced with this merge are not suitable for real disclosure until Phase 1.5.x. Worth a note in the CHANGELOG or PR description to make that explicit for anyone who merges and tries to use it immediately.

None of these block the merge, which is solid.  the isolation is clean and existing functionality is untouched. Items 1 and 2 are worth a follow-up commit before the branch lands if Mark has bandwidth. Thank you @unprovable for this, very sexy time

danielcuthbert:
I'm leaning more towards a Single PR if im honest

unprovable:
Ok, to be explicit here in the PR:

⚠️  Phase 1.5: placeholder hashes — not suitable for real disclosure

Bundles produced by this branch contain placeholder verifier_key_hash and harness.hash fields — they hash the literal strings "placeholder-vk-1.5" and "harness-1.5", not the real SP1 verifying key or harness binary.
- python3 raptor.py prove-exploit refuses to write a bundle without --allow-placeholder-hashes. Verify-time prints a loud warning.

**NB** - See `docs/zkpox-scope.md` for the full statement of what this MVP proves and doesn't prove.

- Phase 1.5.x replaces both fields with their real digests (sp1-sdk verifying key + harness binary hash).
- Do not use bundles produced by this merge for real CVE disclosure until 1.5.x lands. This is fine for demos, internal testing, and end-to-end pipeline validation. 

Do we want to add a changelog file formally? Else, see edit above.

unprovable:
OK, I think this starts to answer your review - thank you so much, @danielcuthbert . Also incoming in the hour should be a warning banner that says "this is still experimental". Experimental but this is the bleeding edge of disclosure, I think. Thanks for the engagement, too! :D

unprovable:
ok, also added a fix to make the tests faster as ~16"30s per run is way way way too long!

unprovable:
ok there's a bug, chasing. EDIT: Ok I think it's resolved. the ZKPoX tess shouldn't run on non-ZKPOX related changes. Ready to kick back to review, @danielcuthbert -vMany thanks!

grokjc:
This is cool.

But if this is experimental then I think it needs to be optional 😄 ... so requirements.txt needs to reflect that and we need to handle missing modules cleanly.

I'm working elsewhere on tagging slow tests so they can be gated appropriately, perhaps this is also a consumer...

grokjc:
To save you scratching your head re the failing tests: they are failing because cryptography and cbor2 are commented out in requirements.txt (which is what we want!) but packages/zkpox/__init__.py eager-imports them via .anchor / .bundle / .envelope. The zkpox-regression Rust tier you added is correctly gated, but the Python tests still collect into python-unit-tests-fast and blow up at import time before any path filter can save them.

  Keeping the deps optional, three changes:

  1. packages/zkpox/__init__.py — wrap the re-exports, expose AVAILABLE and require():
```
  AVAILABLE = False
  _IMPORT_ERROR = None

  try:
      from .anchor import (...)
      from .bundle import (...)
      from .envelope import (...)
      AVAILABLE = True
  except ImportError as exc:
      _IMPORT_ERROR = exc


  def require():
      if not AVAILABLE:
          raise RuntimeError(
              "zkpox needs cryptography and cbor2. "
              "pip install cryptography==45.0.4 cbor2==6.0.1 "
              f"(or uncomment them in requirements.txt). {_IMPORT_ERROR}"
          )
```
  2. raptor_zkpox.py — only one production import site (line 37). Call zkpox.require() at the top of main() so a missing dep surfaces as a one-line install hint instead of a traceback.

  3. packages/zkpox/tests/{test_anchor,test_bundle,test_envelope,test_prove}.py — top of each:
```
  import pytest
  pytest.importorskip("cryptography")
  pytest.importorskip("cbor2")
```
  Collection-safe without the deps (skip cleanly), unchanged when they're installed. No workflow change needed — your zkpox-regression tier stays as-is. HTH 😄

grokjc:
Sandboxing: Six raw subprocess.run sites in this PR need to route through core.sandbox. Same pattern as core/git/clone.py:175-185 and the recent ffuf PR (#489). get_safe_env() handles env hygiene but doesn't isolate filesystem or network.

  The sites:

  - envelope.py (×2) — age-keygen, then age/tlock encrypt/decrypt. The encrypt/decrypt one is the most attacker-facing — it processes crafted ciphertext and keys. run_untrusted, plus use_egress_proxy=True, proxy_hosts=["api.drand.sh"] for the tlock decrypt path (Drand fetch).
  - prove.py — zkpox-prove on the witness. run_untrusted, no network.
  - raptor_zkpox.py (×2) — zkpox-prove and zkpox-verify orchestration. Add use_egress_proxy=True, proxy_hosts=["rekor.sigstore.dev"] if Rekor anchoring is in-band here.

  Sketch for the most attacker-facing case (age/tlock decrypt):
```
  from core.sandbox import run_untrusted

  completed = run_untrusted(
      cmd,
      target=str(input_path.parent),
      output=str(self.out_dir),
      readable_paths=[str(input_path.parent)],
      use_egress_proxy=True,
      proxy_hosts=["api.drand.sh"],
      tool_paths=[str(Path(shutil.which("age")).parent)],
      caller_label="zkpox-envelope",
      timeout=...,
      input=stdin_bytes,
      capture_output=True,
  )
```
  Three things to know (same as the ffuf PR):                                                                                                                                                                      
   
  - run_untrusted always blocks network at the namespace level. The egress proxy works alongside, not against — child sees only HTTPS_PROXY=127.0.0.1:&lt;port&gt;.
  - proxy_hosts is hostname-only — port doesn't matter.         
  - The Rust binaries live in core/zkpox/target/release/, so tool_paths=[str(Path(prover).parent)] is mandatory — they're not on the default safe-bin path.                                                        
                                                                                                                                                                                                                   
  While I'm here: if you're shelling out to rekor-cli, consider doing the HTTPS call directly via core.http.HttpClient instead — keeps the egress accounting and proxy plumbing inside RAPTOR's existing surface rather than spawning another sandboxed binary.

unprovable:
Hmm! This is interesting - whilst working on this, it's turned out that the CodeQL checks bypassed the lines indicated in https://github.com/gadievron/raptor/pull/470/checks?check_run_id=76939588240 <-- mostly because of 'large diff', I think. Here are the offending lines: 

  ### `core/sandbox/context.py` (13)
  - [ ] `:555` — clear-text logging
  - [ ] `:637` — clear-text logging
  - [ ] `:712` — clear-text logging
  - [ ] `:714` — clear-text logging
  - [ ] `:804` — clear-text logging
  - [ ] `:1309` — clear-text logging (×2 sources tainting same sink)
  - [ ] `:1329` — clear-text logging
  - [ ] `:1393` — clear-text logging (×2 sources)
  - [ ] `:1630` — clear-text logging
  - [ ] `:1645` — clear-text logging
  - [ ] `:1652` — clear-text logging

  ### `core/sandbox/observe.py` (11)
  - [ ] `:138` — clear-text logging
  - [ ] `:147` — clear-text logging
  - [ ] `:154` — clear-text logging
  - [ ] `:157` — clear-text logging
  - [ ] `:195` — clear-text logging (×2 sources)
  - [ ] `:199` — clear-text logging (×2 sources)
  - [ ] `:205` — clear-text logging (×2 sources)
  - [ ] `:209` — clear-text logging (×2 sources)
  - [ ] `:312` — clear-text logging (×4 sources)
  - [ ] `:319` — clear-text logging (×4 sources)
  - [ ] `:335` — clear-text logging (×4 sources)

Is this worth a separate issue, @grokjc ?

grokjc:
On it, stand by 😄

grokjc:
All FPs. d6ab5ae7377c83935e2299c56a05efdb9cf97fbf should help. Next commit will trigger a re-scan.

unprovable:
Hmm still flagging - maybe the fix is that in `core/sandbox/context.py` on lines 1246, 1820, 1821, rename `proxy_token` to something like `proxy_registration` or `proxy_handle` or `proxy_reg_id`. This rename removes the only heuristic-named sensitive source in production sandbox code. That said, there's probably other code changes required. Any thoughts, @grokjc ?

grokjc:
Probably needs you to rebase to pick up the config change too ... nope you did that. Investigating further 😄

grokjc:
ok, try another rebase ...?

grokjc:
The `vendor_secret_key: Path` parameter pins CodeQL's heuristic source list (any identifier containing `secret_key` is tagged), even though the parameter is a *path* to a key file, not key material. The taint then flows through `cmd` lists in `age_decrypt_with` → `_run` → `run_untrusted` and surfaces as `py/clear-text-logging-sensitive-data` on every sandbox audit-log sink downstream.

  The cleanest fix is to align the name with the type — it's a path, name it as one. That defangs the heuristic at the root without suppressing real `*_secret_key` alerts elsewhere.

  ```suggestion
  def open_via_vendor(envelope: Envelope, vendor_key_path: Path) -> bytes:
      """Open an envelope by the vendor path: age decrypt K, then AES."""
      K = age_decrypt_with(envelope.ct_K_age, vendor_key_path)
      return aes_decrypt(envelope.aes_blob, K)
  ```

  `identity_file` (the inner parameter on `age_decrypt_with`) is already fine — no `secret` token. 

  For sandbox-side defence-in-depth, main now stamps these as triaged FPs in the SARIF on upload (see `KNOWN_FP_RULES` in `.github/scripts/sarif_known_fp_suppressions.py`), so even if a future caller introduces a similarly-named source variable the alert will land as Dismissed with the justification visible. The rename here is still worth doing.

unprovable:
its-alive

grokjc:
It prompted some useful engineering work around codeql, sorry you were the innocent victim here 😄

unprovable:
Not a problem! Always better to get them in the open and resolved :P Thanks for your help! :-D

grokjc:
I have something in the works that will open the door for ZKPoX for our users ... bear with - this exact PR will have a clear home and a way forward very shortly if all goes to plan! Just hold off a little longer 😄

grokjc:
See #625 and #632 for some foundational work opening this up to production

grokjc:
Suggest we adopt this convention please:
```
  /zkpox eligible      # the free pre-flight signal               (which witnesses qualify)
  /zkpox bundle <w>    # Tier 0/1 — assemble prover-ready bundle  (shipped)
  /zkpox reproduce <b> # Tier 1.5 — N× reproduction               (shipped)
  /zkpox prove <b>     # Tier 2/3 — the heavy proof               (#470)
  /zkpox verify <p>    # check a proof                            (#470)
```
(saves us from introducing many top-level commands/skills)

grokjc:
A few more things to (hopefully) help:

Tiers 0/1 + 1.5 (packages/zkpox/) are shipped and in use; they must keep importing on a box with no proving stack. So the SP1/RISC-V/age/tle imports need to be lazy, inside the prove/verify bodies, not at module top level. For juggling the deps/not-deps situation I've just landed #642 which should help:
```
  from packages.zkpox import require_proving_stack   # raises ProvingStackUnavailable, actionable msg
  def prove(...):
      require_proving_stack()
      import sp1   # lazy — only here
      ...
```

For slow tests you can now double-gate, too:
```
  from packages.zkpox import proving_stack_available
  
  @pytest.mark.slow
  @pytest.mark.skipif(not proving_stack_available(),
                      reason="SP1/RISC-V proving stack not installed")
  def test_real_proof_roundtrip(...): ...
```
Keep the deterministic stuff (CLI arg handling, the require_proving_stack error path, bundle validation, mocked-prover dispatch) unmarked so it runs every PR — that's the bulk of the coverage and it needs no deps.

The rest is up to you @unprovable 😄

unprovable:
@grokjc 
## Note: piggybacks an SCA fix

This PR was originally just the `/zkpox` convention switch, but the last CI run surfaced two failing SCA typosquat test groups that turned out to be a latent bug in `main` — not something this PR introduced. Rather than open a parallel PR and block on it, the fix is folded in here as commit `5a48d640`.

### What broke

The auto-refresh job (#686) lands top-popular npm packages into `packages/sca/data/popular/npm.json` once a week. Popularity feeds rank by raw download / depended-upon count, so a real-world typosquat that gets enough mistaken-install traffic can bubble into the top-N. `loadash` did exactly that. 😅

The typosquat detector treats anything in the popular list as legitimate by definition — so once `loadash` landed there, the detector silently disarmed for that name. The detector's tests use `loadash` as the canonical distance-1 typosquat of `lodash`, so they started failing.

### Why CI didn't catch it on main

`compute_filters.py:_read_changed_files()` falls back to forcing every filter `true` when the diff base is unavailable — explicitly the fork-PR partial-auth case. Subsequent push-to-main runs after #686 didn't touch SCA paths, so the `sca` filter skipped them and the breakage went unnoticed. The next fork PR (this one) forced all filters on and exposed it.

### Fix shape

- `packages/sca/data/known_typosquats/<eco>.json` — per-ecosystem hand-curated denylist; seeded with `loadash` on npm. Other ecosystems unseeded (no behaviour change without a list).
- `refresh_typosquat_lists.py` filters out denylisted names after each fetch, before persistence. Empty-after-filter is surfaced as `failed: empty result after denylist` so it can't silently disarm the detector wholesale.
- One-time scrub of the current `popular/npm.json` removes `loadash` so the present state matches what the next refresh would write (4995 → 4994 entries; `lodash` retained).
- Two tests pin the invariants.

grokjc:
Already working on it, you just got unlucky!

grokjc:
Ironically I had to fix a LOT of bugs to make these tests break 😃

unprovable:
I think we found one of the issues, seeing if this push fixes anything

grokjc:
Fixed with [27dbecd](https://github.com/gadievron/raptor/pull/688/commits/27dbecd473c8396f6e78f846b779da711edc8872)

grokjc:
cherry-picked that commit onto main, thanks!

unprovable:
I think it's settled down, now :P LEt me know if any changes need to be made further :)

grokjc:
Looks like the CBOR schema/module is missing somehow? Also, can you drop the SCA commits from this PR - that is now behaving itself and doesn't need these specific bits now.

grokjc:
Tier 2/3's input is ZKPoXBundle.as_dict() - the manifest bundle + reproduce already build and enrich. But cmd_prove (raptor_zkpox.py:160) never reads it: it takes raw flags (--witness, --target, --gadget-id, …), re-hashes --target, and constructs a fresh CBOR Bundle from scratch.
  
  Three consequences:
  1. Tier 1.5 evidence is lost - the reproduction block in the manifest never reaches the disclosure bundle.
  2. Unchecked re-keying - --target isn't reconciled against the manifest's target_binary_hash, so the proof can silently bind to a different artifact than the one triaged/reproduced.
  3. Name collision - ZKPoXBundle (manifest) vs the new CBOR Bundle are unrelated dataclasses with no conversion path.
  
  To be clear, the extra flags prove genuinely needs (--vendor-pubkey, --gadget-id, --harness-*, --wrap) are fine. The problem is it reads none of the bundle and rebuilds everything from flags.

Suggestion:
  - Make prove take the bundle dir as its primary arg; load manifest.json → ZKPoXBundle and derive witness/target-hash/outcome/claim from it. Keep the Tier 2/3 flags as additions.
  - Add one ZKPoXBundle → CBOR projection function, and rename the CBOR type (e.g. DisclosureBundle) to kill the collision.

unprovable:
Ok I think this fixes the concerns raised. I'm still looking for major improvements, but getting it inline with the overarching machinery looks to be done. Superb work on the backend magic @grokjc !!

grokjc:
cbor2/cryptography are commented out in requirements.txt (correct — keeps the runtime dependency-free) but they're also absent from requirements-dev.txt. So with the pytest.importorskip guards, test_disclosure / test_anchor / test_envelope / test_prove all skip in the Python CI tier, and the zkpox-regression job only builds + sweeps the Rust side. Net effect: the entire restored disclosure/anchor/envelope/prove layer — and the new manifest-loading path — is unverified by CI. A future rebase that re-drops the module or unwires __init__.py wouldn't be caught.
  
This is exactly the case requirements-dev.txt already handles for beautifulsoup4 ("Optional runtime deps that gate test collection. Listed here, not in requirements.txt, so CI runs the full suite"). Same pattern here — the importorskip guards stay as dev defense-in-depth.

Suggestion:

```suggestion  
  beautifulsoup4==4.14.3
  
  # ZKPoX Tier 2/3 (packages/zkpox/{disclosure,anchor,envelope,prove}).
  # Optional at runtime — kept out of requirements.txt to preserve the
  # dependency-free Tier 0/1 import contract — but required here so the
  # disclosure / anchor / envelope / prove tests actually run in CI instead
  # of importorskip-ing. Tests keep their importorskip guards as
  # defense-in-depth for devs running pytest without these installed.
  cryptography==45.0.4
  cbor2==6.0.1
  ```

**Comment by grokjc** [comment]
`test_parse_full_record_via_fake_binary` is flaky: it's a pure parser test but routes through `run_untrusted`, so it fails on kernels that drop the sandbox to Landlock-only (`pyvenv.cfg` PermissionError, unrelated to the code). Fix: extract a pure `_record_to_result()` and test it directly — the real prove path is covered by `core/zkpox/test/run-tests.sh`.
  
  **`packages/zkpox/prove.py` — lines 150–167:**
  
  ```suggestion
      return _record_to_result(record)
  
  
  def _record_to_result(record: dict) -> ProveResult:
      """Pure schema-mapping half of :func:`run`, split out so it can be
      unit-tested without spinning up the prover or the sandbox."""
      v = record.get("verdicts") or {}
      return ProveResult(
          tag=record["tag"],
          witness=Path(record["witness"]),
          witness_bytes=int(record["witness_bytes"]),
          mode=record["mode"],
          verdicts=Verdicts(
              target_id=int(v.get("target_id", 0x01)),  # default 01 for pre-1.6 records
              crash_only_crashed=bool(v["crash_only_crashed"]),
              oob_detected=bool(v["oob_detected"]),
              oob_count=int(v["oob_count"]),
              oob_first_offset=int(v["oob_first_offset"]),
          ),
          cycles=record.get("cycles"),
          wall_secs=float(record["wall_secs"]),
          proof_bytes=record.get("proof_bytes"),
          verified=record.get("verified"),
      )
  ```
  
  **`packages/zkpox/tests/test_prove.py` — lines 38–76:**

  ```suggestion
  def test_parse_full_record_preserves_precision(tmp_path):
      """Full-schema record → ProveResult, no precision loss. Pure parser,
      no subprocess; the real prove path is covered by run-tests.sh."""
      record = {
          "tag": "smoke", "witness": str(tmp_path / "w.bin"), "witness_bytes": 32,
          "mode": "execute",
          "verdicts": {"target_id": 1, "crash_only_crashed": True, "oob_detected": True,
                       "oob_count": 16, "oob_first_offset": 16},
          "cycles": 9807, "wall_secs": 0.014, "proof_bytes": None, "verified": None,
      }
      result = prove._record_to_result(record)
      assert (result.tag, result.witness_bytes, result.mode) == ("smoke", 32, "execute")
      assert result.verdicts.target_id == 1
      assert result.verdicts.oob_count == result.verdicts.oob_first_offset == 16
      assert result.cycles == 9807 and result.wall_secs == 0.014
      assert result.proof_bytes is None and result.verified is None
  
  
  def test_parse_record_defaults_target_id_for_pre_1_6(tmp_path):
      """Pre-1.6 record with no target_id defaults to 0x01."""
      record = {
          "tag": "old", "witness": str(tmp_path / "w.bin"), "witness_bytes": 8,
          "mode": "execute",
          "verdicts": {"crash_only_crashed": False, "oob_detected": False,
                       "oob_count": 0, "oob_first_offset": 0},
          "wall_secs": 0.001,
      }
      assert prove._record_to_result(record).verdicts.target_id == 0x01
  ```

**Comment by grokjc** [comment]
I think we're at the point where the crypto may be experimental but the code is production grade 😀 
Do you have a view as to whether the crypto side of things is likely to evolve any time soon?

**Comment by unprovable** [comment]
Ok, this might be a wordy response, but here goes...

The core thing to realise is that there is a stack here that has lots of moving parts cryptographically and beyond: there are essentially three tiers of volatility:
  
**Bedrock**; this won't move. AES-256-GCM, SHA-256, and age (X25519 + ChaCha20-Poly1305). These are settled. 256-bit symmetric crypto is even fine against a quantum computer (Grover's search only halves the security margin, and most people consider it a non-starter anyway). Nothing to track here, possibly ever.

**Mature, but with a known ceiling** The Groth16/BN254 SNARK wrap and Drand tlock (BLS + identity-based
  encryption on BLS12-381). The maths is stable and independently audited (Kudelski audited tlock). The evolution pressure on this layer is not "is it correct" — it's two structural things:
  - Trusted setup: Groth16 needs per-circuit setup; the ecosystem keeps drifting toward universal-setup systems (PLONK, etc.) — and notably SP1 already supports a PLONK wrap as well as Groth16.
  - Post-quantum: both the wrap and the time-lock are pairing-based and fall to a cryptographically-relevant quantum computer. drand themselves are blunt about it — "neither BLS nor the IBE scheme are quantum resistant, and there are no quantum-resistant schemes currently available that could replace the current design" (PQ signatures don't "thresholdize" well into IBE yet). Their own framing is the 5–10-year horizon. So this layer moves on a multi-year, research-gated clock... it evolves when someone has a concrete reason, and a working system (of which raptor now is!!) is part of what creates the reason.
  
**The Fast Frontier** The zkVM/proving layer (SP1) and the transparency log (Rekor). These two will
  keep evolving whether or not ZKPoX exists, because they're driven by the rollup and supply-chain ecosystems with enormous resources behind them:
  - SP1 has already had a major architectural redesign since the v6.1.0 bundles pin — the current generation ("Hypercube") does real-time Ethereum block proving on GPUs, and the Ethereum Foundation has committed to ZK as core infrastructure for 2026+. The proof-system field in this bundle (sp1-groth16/v6.1.0) is going to keep incrementing.
  - Rekor went v1→v2 (GA'd): tile-based backend, and crucially a new witnessing/cosigning model baked in for stronger append-only guarantees, plus annual log shards (log2025, log2026). The deferred "full Merkle inclusion proof verification" (the Phase 1.5.x item) lands right on top of this moving target when we can get it in (working on it!)

Will all this crypto evolve soon?? **Yes**, but mostly in the layers ZKPoX consumes rather than owns (SP1, Rekor), and that's continuous maintenance (version pins, verification paths), not a redesign. The layers that are load-bearing for the threat model (pairing-based wrap + time-lock, and the absence of true witness encryption) move on a years-long, research-gated horizon.
  
The single most interesting "evolves when people see it in play" candidate is witness encryption.... currently flagged as 'non-deployable' (Garg–Gentry–Sahai–Waters relied on broken multilinear maps). True witness encryption that is "decryptable by anyone who holds a valid exploit" would change the disclosure model entirely: no vendor key, no time-lock escrow... There's been LWE / evasive-LWE progress but it's firmly research-stage. That's where the epithet "people research it when they see the use case in a system like this" is most literally true. (Well, I hope...)

**Comment by grokjc** [comment]
One more tiny thing before this lands (unrelated to ZKPoX, just caught in passing.)
  
  The reformat of `_HELP_EPILOG` in `raptor.py` dropped the `sca` mode line. It's on `main` (and the merge-base) but not here, so the PR removes it — and since `main` hasn't touched that block, a rebase would silently keep the deletion rather than flag it, so it needs fixing here. Harmless (the `sca` command still works), it just won't show in `raptor.py --help`.

  One line, restoring the exact wording from main:

  **`raptor.py` — anchor on the `scan` epilog line (~818):**
  
  ```suggestion
    scan                  - Static code analysis with Semgrep
    sca                   - Software Composition Analysis (deps + advisories + SBOM)
  ```

With that, everything from the review is resolved on my end - conventions, manifest contract, CI coverage, and the de-flaked parser test all look good. 👍 (and I'm off to read your short crypto essay next 😆)

**Comment by grokjc** [comment]
Can't beat Friday afternoon PQC. That makes a lot of sense and it is clear which bits are evolving and which bits are relatively stable. Debatable if X25519 is a problem vs CRQC, depends on disclosure/escrow timelines I suppose 😄 

So ... with that said, when do you think your next bit (1.5.x) might land? Because that is the difference between "run this on main as a research project for a while" or "wait until next month and it is production ready". No pressure!

**Comment by grokjc** [comment]
Sorry, just tightened up the rules on the coverage file.  Run the python script as it suggests 😄 
(`python .github/tests/test_filter_coverage.py --update`)

**Comment by unprovable** [comment]
Running the script came back 'nothing to be done' but Monsieur derived that the issue was the order. Changes: 
- zkpox — reordered to canonical: own package first (packages/zkpox/**), then packages/* (packages/llm_analysis/exploit_verify.py), then core/* alpha — with core/zkpox/** sorting into the core block rather than pinned to the top.
- llm_analysis — dropped a duplicate core/hash/** + core/witness/** pair (both already appear in the core/* block).

**Comment by unprovable** [comment]
mmmmm not sure that's the cleanest fix >.<

**Comment by unprovable** [comment]
So, the prover already links sp1-sdk 6.0.1 and uses its proving/verifying-key machinery (ProvingKey, the gnark Groth16 path, etc.). The verifier deliberately doesn't link it yet — as doing so may effectively double the verifier's compile time. So the hard cryptographic capability exists in the workspace; 1.5.x is wiring it into the verify path and killing placeholders, not inventing anything.

Suppose I work it fully out - do you have some cycles coming up @grokjc to maybe help manage the time to compile? I think I can have a commit ready soon for testing and inspection. If it's too heavy and takes too long, we can bypass it for now.

**Comment by grokjc** [comment]
We've come this far, why not? 😆 I'd be happy to give it a go - no promises!

**Comment by unprovable** [comment]
Ok, here's half... rest will be incoming soon!

**Comment by unprovable** [comment]
ok so... what's up here is essentially a production MVP for ZKPoX with the full verifier. I'm thinking this can be validated and will spend some time today/tomorrow to do so. Testing with a CVE such as CVE-2017-9047 is my goto default as there's already a witness in `ore/zkpox/test/witnesses/03-overflow1-crash.bin`... Here's the test script I'm working with, in case @grokjc has any cycles coming up to test with? :-) [test-zkpox.sh](https://github.com/user-attachments/files/28423228/test-zkpox.sh)

**Comment by grokjc** [comment]
Spent some time looking at the workspace shape to see where build time goes — a few ideas that spring to mind, sharing in case any of them land easily:
  
  **Profile tuning in the top-level Cargo.toml** — confirmed no `[profile.*]` overrides anywhere in the workspace (checked all four Cargo.toml files), so deps build at `opt-level=0` in dev. Adding:

  ```toml
  [profile.dev.package."*"]
  opt-level = 3

  [profile.dev]
  debug = "line-tables-only"
  split-debuginfo = "unpacked"
```
  should net 20-40% on cold builds for sp1-sized workspaces, and deps stay cached across workspace-source iteration. Wouldn't change anything visible at runtime.

  Faster host linker — no .cargo/config.toml in the repo today, and no -fuse-ld / RUSTFLAGS override anywhere. lld is already installed in the zkpox-regression CI job (for guest RISC-V cross-compile) but isn't wired as the host linker. A new .cargo/config.toml like:
```  
  [target.x86_64-unknown-linux-gnu]
  linker = "clang"
  rustflags = ["-C", "link-arg=-fuse-ld=lld"]
```
  would route the prover/verifier link through lld without changing CI install steps. Typically 5-10× faster than the default linker on dep-heavy workspaces. (mold's slightly faster but adds a package; lld's free given it's already there.)

  sccache — not currently wired anywhere in the repo (no .cargo/config.toml, no mention in any CI workflow). System-wide compile cache, survives cargo clean; could be a local-dev win without touching CI yet.

  native-gnark on the verifier — sp1-sdk is pulled with ["blocking", "native-gnark"] for both prover and verifier. The zkpox-regression CI comment already notes sp1-recursion-gnark-ffi pulls golang-go in even on execute-only flows. The verifier's full_verify calls client.verify(&proof, pk.verifying_key(), None) — whether that path internally needs gnark depends on which wrap mode the proof was made with. If proofs intended for the standalone verifier are wrap=core only, dropping native-gnark from the verifier's sp1-sdk features would cut the verifier build substantially. You'd know better than I do whether shippable disclosure bundles need groth16 verification on the standalone path.
  
  The CI cache strategy for core/zkpox/target + ~/.sp1 keyed on Cargo.lock is already solid — survives most PR-tier runs. The full-verify feature gate is also a really clean split: contributors who aren't touching the crypto layer can build with --no-default-features --features structural and avoid the SP1 cost entirely. Neither packages/zkpox/README.md nor core/zkpox/README.md documents that escape hatch today - adding a "fast local build" note to one of them would help new contributors find it.

unprovable:
Thank you!! So, applied the two zero-risk ones: dev-profile tuning (release left alone for proof reproducibility) and a README "fast local build" note for the structural escape hatch, plus an opt-in lld/sccache config snippet. Held off committing a global `.cargo/config.toml` since the linker override would break any Linux dev box without clang/lld. On native-gnark: the standalone verifier has to verify shipped disclosure bundles, which are groth16-wrapped (core is too big for CBOR), so client.verify() needs gnark there, os keeping it if that makes sense?

unprovable:
[This doc](https://github.com/gadievron/raptor/pull/470/commits/1b54a14bdde08fe1ea3ec94bf0907d86da8d860b)  details a plan that includes using a differential oracle to encode a full-fat real CVE (basically, the grown up version of the pared down one I've been using for a lot of testing). This might be worth an attempt at some point for some wonder of a mortal soul...

grokjc:
Another rebase should fix your test failure.

unprovable:
sorry was at a conference - bit more tweaking...

unprovable:
ok assuming the last two tests go without hitch, I think this is ready @grokjc !

grokjc:
Another rebase will fix the CI error (unrelated to your work). From my POV this is looking good, I'm just holding off until I finish messing with /exploit so there is a consumer and we can do E2E testing. The Rust is solid, the python is all debugged, just a matter of timing now. Unless you are planning on evolving this further?

unprovable:
> Another rebase will fix the CI error (unrelated to your work). From my POV this is looking good, I'm just holding off until I finish messing with /exploit so there is a consumer and we can do E2E testing. The Rust is solid, the python is all debugged, just a matter of timing now. Unless you are planning on evolving this further?

Not planning on making more changes without a real pipeline to run it off :P so sounds like it's good to keep alive for when `/exploit` evolves!

unprovable:
This bug is hilarious... given we use pycrypto in this PR, updated accordingly: https://github.com/advisories/GHSA-537c-gmf6-5ccf

grokjc:
Issues seem to be vulnerable quinn-proto and p3-challenger packages.

grokjc:
Have dug into this further - the SCA gate is failing on one fixable high-severity finding:

  - **`quinn-proto@0.11.14`** - RUSTSEC-2026-0185 (remote memory exhaustion from unbounded out-of-order stream reassembly). Fixed in `0.11.15`. A `cargo update -p quinn-proto` in `core/zkpox/` should resolve it.

There are also 12 new medium-severity Cargo advisories introduced by this PR. Most are structural - transitive deps of the SP1 prover SDK (`p3-challenger`, `p3-symmetric`, `ansi_term`, `derivative`, `paste`, etc.) that can't be bumped independently. Two have patch-level fixes available:

  - `memmap2@0.9.10` → `0.9.11` (`cargo update -p memmap2`)
  - `lru@0.12.5` → `0.16.3` (major version bump - may need API changes upstream in SP1)

  The rest have no patched version and are waiting on SP1 to update their deps. None of these block the gate (threshold is high severity).

grokjc:
7ca15d787271f3d7917626c76508f43ff2500f76 will fix your tests.

grokjc:
SCA CI is highlighting:
  1. pnpm 11.3.0 — 10 high-severity GHSAs. All from a single outdated tool version. Bump pnpm to fix the whole cluster.
  2. pip 26.1.1 — 1 high (GHSA-wf93-45jw-7689). Same story, bump pip.

d4cee71 is your friend.

unprovable:
Soryr for the delay - think this is done now, @grokjc

grokjc:
https://github.com/gadievron/raptor/commit/689a45344c55d89da727ff40b597210b50840931 needed ... (no rush!)

unprovable:
Done! Apologies for the delay!

unprovable:
ok think this is fixed - hope the tests pass :)

grokjc:
cryptography==48.0.1  (requirements-dev.txt line 39)

   * GHSA-jwv3-5hgf-82ww — exponential path-building via duplicate self-signed intermediates (High). Fixed in 49.0.0.
   * GHSA-g6cj-pr64-35w5 — PKCS#7 EnvelopedData Bleichenbacher oracle (High). Fixed in 50.0.0.
   
Fix: bump the pin to cryptography>=50.0.0 (also update the commented pin in requirements.txt so the two files don't drift). The bump-comment there cites the OpenSSL advisory that motivated 48.0.1 — 50.x clears that one too.

p3-challenger 0.3.2-succinct  (core/zkpox/Cargo.lock, pulled in transitively by the sp1 zkVM stack)
   * GHSA-vj64-rjf3-w3v7 — Plonky3 MultiField32Challenger transcript malleability and challenge entropy loss (High). Fixed in 0.4.3 and 0.5.3 — there is no fixed release on the 0.3.x-succinct line, so a lockfile-only bump won't reach it.

Fix: upgrade the sp1 dependency stack to a release that pulls p3-challenger >= 0.4.3 (or >= 0.5.3), then regenerate Cargo.lock. Given ZKPoX's soundness rests on the Fiat-Shamir challenger, this one is worth fixing rather than waiving — a malleable transcript is exactly the property a proof-of-exploit scheme can't tolerate.

grokjc:
`Cargo:p3-challenger@0.4.3-succinct` (GHSA-vj64-rjf3-w3v7) still a problem 🤬

grokjc:
It was a (mildly interesting) false positive (OSV) that I've worked around - see https://github.com/gadievron/raptor/commit/be9c11d9914987b4bed02dd9bb7aa7762062646c and https://github.com/gadievron/raptor/commit/a4c6876f2a68823c228be9df2cf68c0ef926182b.

Meanwhile /exploit is coming along nicely, so this code will get used very soon! One thing I'm wondering is if all of this should target `packages/` - as the split between `core/` and `packages/` seems arbitrary and there's really only one consumer?

PR #486, opened 2026-05-14
feat: add raptor doctor self-check

Hinotoi-agent:
## Summary
- add a `raptor doctor` self-check command for setup diagnostics
- route `./bin/raptor doctor` without requiring Claude Code first
- cover the doctor checks with focused unit tests

## Why
This gives new and returning RAPTOR users a quick way to identify common local setup issues before running heavier scans. The output is intentionally actionable and copy-paste friendly.

## Validation
- `python3 -m ruff check core/doctor.py core/tests/test_doctor.py raptor.py`
- `python3 -m pytest -q core/tests/test_doctor.py core/security/tests/test_log_sanitisation.py`
- `python3 raptor.py doctor`
- `./bin/raptor doctor`

### DIFF
--- a/bin/raptor
+++ b/bin/raptor
@@ -82,6 +82,7 @@ if [[ "${1:-}" == "-h" || "${1:-}" == "--help" || "${1:-}" == "help" ]]; then
 
 Usage: raptor [options] [target]
        raptor project <subcommand> [args]
+       raptor doctor
 
 Arguments:
   target             Path or URL to scan (optional)
@@ -117,6 +118,13 @@ if [[ "${1:-}" == "project" ]]; then
     shift
     cd "$RAPTOR_DIR" || exit 1
     exec "$RAPTOR_DIR/libexec/raptor-project-manager" "$@"
+fi
+
+# Route doctor directly (no Claude needed)
+if [[ "${1:-}" == "doctor" ]]; then
+    shift
+    cd "$RAPTOR_DIR" || exit 1
+    exec python3 "$RAPTOR_DIR/raptor.py" doctor "$@"
 fi
 
 # Claude required for everything else
--- a/core/doctor.py
+++ b/core/doctor.py
@@ -1,0 +1,139 @@
+"""Environment self-checks for RAPTOR operators.
+
+The doctor command is intentionally lightweight: it checks local prerequisites
+and filesystem assumptions without contacting external services or reading any
+secrets.  It is safe to run before an analysis session to explain likely setup
+problems in one place instead of failing later inside a scanner wrapper.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+import os
+from pathlib import Path
+import platform
+import shutil
+import sys
+from typing import Iterable
+
+
+@dataclass(frozen=True)
+class DoctorCheck:
+    """One doctor check result."""
+
+    name: str
+    status: str
+    detail: str
+
+
+_REQUIRED_DIRS = ("core", "packages", "libexec", "bin")
+_OPTIONAL_TOOLS = (
+    ("claude", "Claude Code launcher"),
+    ("semgrep", "static-analysis scans"),
+    ("codeql", "CodeQL scans"),
+    ("afl-fuzz", "binary fuzzing"),
+)
+
+
+def _status_line(check: DoctorCheck) -> str:
+    marker = {"pass": "✓", "warn": "!", "fail": "✗"}.get(check.status, "?")
+    return f"{marker} {check.status.upper():4} {check.name}: {check.detail}"
+
+
+def _repo_root(explicit_root: Path | None = None) -> Path:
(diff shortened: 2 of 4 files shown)


grokjc:
I do like the idea, but this implementation reinvents (and contradicts) a lot of what is already present elsewhere in the codebase. Let me take a deeper look and come back to you on this one.

grokjc:
Thanks for raising this — the `raptor doctor` concept is one I've wanted in the tree for a while, and your PR (along with #57 from splinters-io's aborted Frida-integration work, which first surfaced the operator-facing self-check shape) was the prompt to actually build it.
                                                                                                                                                                                                                   
  I'm going to close this in favour of a from-scratch implementation on `feat/raptor-doctor` (commit landing shortly). Both of you are credited in the commit body. The reasons I didn't merge as-is:              
                                                                                                                                                                                                                   
  **Duplicate check infrastructure.** The PR's `core/doctor.py` re-implements tool detection, env var checks, and LLM key checks that already live in `core/startup/init.py` (powering the SessionStart banner).   Two sources of truth for "what RAPTOR knows about its environment" drift apart over time — a new tool added to `RaptorConfig.TOOL_DEPS` would light up in the banner but not the doctor, or vice versa. The shipped version wraps the existing `check_tools` / `check_llm` / `check_env` / `check_lang` / `check_active_project` functions, so one new check lights up in both surfaces.                                     
                                                               
  **Flat module placement.** `core/doctor.py` is a bare file in `core/`. The repo convention (enforced through a recent migration arc — see PR #447) is "no bare files in `core/`": every unit is a package directory with colocated `tests/`. The shipped version lives at `core/startup/doctor.py` alongside `init.py` because the doctor is a renderer over startup checks.
                                                                                                                                                                                                                   
  **Python version pin.** The PR's check looks for Python 3.9+, but RAPTOR actually requires 3.10+. `packages/exploitability_validation/schemas.py` uses PEP 604 union syntax (`str | None`) at function-definition time without `from __future__ import annotations`, so the module fails to import on 3.9 with a confusing `TypeError`. The shipped version pins 3.10+ and surfaces the version-mismatch warning before the operator hits the import trace. `docs/ARCHITECTURE.md` was also stale at "Python 3.9+" and has been updated.                                                                                                     
                                                               
  **`.raptor/tmp` writability check.** That path isn't a RAPTOR concept — RAPTOR uses `out/`, `~/.claude/projects/...`, and the target-repo dir for state. The check would always pass-by-default on a freshly-cloned repo and add noise without signal.
                                                                                                                                                                                                                   
  **Dangerous-env-var allowlist.** The PR introduces an ad-hoc list of `PYTHONPATH` / `LD_PRELOAD` / etc. The project's actual env-hazard convention is `RaptorConfig.get_safe_env()` (strips `TERMINAL` / `EDITOR` / `VISUAL` / `BROWSER` / `PAGER` — variables that scanner subprocesses may shell-evaluate) plus `LLM_API_KEY_VARS` for credential isolation. A second list in the doctor would drift from the real safety story; if the doctor needs to report env hygiene later, it should call into those existing APIs.                                                                                                                       
                                                               
  **`.claude/settings.json` and `.claude/raptor.env` checks.** I thought about these and dropped both. The only way `settings.json` is missing or unparseable is operator edit — they'd recognise their own damage  on sight. `raptor.env` is regenerated every SessionStart by the hook; the failure modes (operator wiped it / hook script broken / claude using wrong project's settings) are either wilful or not fixable by doctor advice. "Restart your claude session" doesn't help when the SessionStart hook itself is broken. The checks would emit noise without actionable signal.                                                    
                                                               
  **Install hints.** The PR's output suggests `apt install <tool>` for missing binaries. RAPTOR's audience is technical operators who can install software themselves — and the hints would often be wrong (`rr` is Linux-only, so a macOS operator gets a contradiction; some tools need build-from-source for newer kernels; Debian's package is sometimes outdated). The shipped version reports the missing dependency and which features degrade, leaving the install path to the operator.                                                                                                                                                     
                                                               
  What does land, paralleling your design:
    * `raptor doctor` CLI route on `bin/raptor` (and `raptor.py` for bare-shell use).
    * `--strict` and `--verbose` flags.                                                                                                                                                                            
    * Non-zero exit on real failure so CI can gate on a clean state.
    * Tool / LLM / env / language / active-project sections.                                                                                                                                                       
                                                                                                                                                                                                                   
  Plus some additions beyond the original proposal:
    * Python 3.10+ enforcement (as above).                                                                                                                                                                         
    * `RAPTOR_DIR` defensive check that includes the computed value so the operator can copy-paste `export RAPTOR_DIR=…`.
    * `escape_nonprintable` pass on every operator-visible string — defence in depth against future ANSI / control-byte injection from tainted upstream sources (subprocess stderr, provider error messages, project names read from disk).                                                                                                                                                                                   
                                                                                                                                                                                                                   
47 tests in `core/startup/tests/`. Crediting you and splinters-io in the commit body. Thanks again for the nudge.

PR #487, opened 2026-05-14
test: harden log sanitisation regressions

Hinotoi-agent:
## Summary
- add regression tests for terminal-control sanitisation edge cases
- cover OSC 8 hyperlinks, OSC 52 clipboard writes, DCS/PM/APC controls, partial escape sequences, and long hostile strings
- keep the PR test-only because the existing sanitizer already neutralizes these sequences

## Why
RAPTOR processes output from tools and untrusted targets. These tests lock in the expected terminal-safety behavior so future sanitizer changes do not accidentally re-enable control-sequence side effects.

## Validation
- `python3 -m ruff check core/security/log_sanitisation.py core/security/tests/test_log_sanitisation.py`
- `python3 -m pytest -q core/security/tests/test_log_sanitisation.py`

### DIFF
--- a/core/security/tests/test_log_sanitisation.py
+++ b/core/security/tests/test_log_sanitisation.py
@@ -47,6 +47,44 @@ class TestEscapeNonprintable(unittest.TestCase):
         # isprintable()-based check catches them.
         self.assertEqual(escape_nonprintable("\x9b[31m"), "\\x9b[31m")
 
+    def test_escape_osc8_hyperlink_sequence(self):
+        # OSC 8 can turn harmless-looking text into a clickable URL in
+        # supporting terminals. Both ESC and BEL must be rendered inert.
+        self.assertEqual(
+            escape_nonprintable("safe \x1b]8;;https://evil.example\x07click\x1b]8;;\x07"),
+            "safe \\x1b]8;;https://evil.example\\x07click\\x1b]8;;\\x07",
+        )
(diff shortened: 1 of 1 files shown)


grokjc:
Thanks - useful addition.  A few thoughts:

  test_escape_very_long_hostile_string has a brittle length-math assertion:
```
  self.assertEqual(len(escaped), len(payload) + len("\\x1b") - 1 + len("\\x07") - 1)
```
  If the escape format ever changes (e.g., \xHH → \u00HH for some path, or sanitiser merges adjacent escapes), this breaks even though the test still passes its actual semantic claims (assertNotIn + assertIn). The other three assertions in the same test cover the real properties; this one could go.

I also note another coverage gap: No has_nonprintable symmetry check - every escape-target should also trigger has_nonprintable=True.

Hinotoi-agent:
Thanks — addressed both points in the latest push.

- Removed the brittle exact length assertion from `test_escape_very_long_hostile_string`; the test now focuses on the semantic guarantees that raw ESC/BEL are absent and the inert escapes are present.
- Added `test_escape_targets_are_detected` so the escape-target samples are also checked against `has_nonprintable()`.

Local validation: `python3 -m pytest core/security/tests/test_log_sanitisation.py -q` → 24 passed.

grokjc:
Thanks!

PR #488, opened 2026-05-15
Feat/fuzzing v2

danielcuthbert:
Phun time!

**What Changed**

- Generates agentic seed corpora from binary strings, nearby source files, READMEs and help-style documentation.
- Persists AFL++ stdout/stderr per instance so forkserver and startup failures are diagnosable from RAPTOR output.
- Improves AFL++ telemetry parsing for executions, paths, coverage, corpus size and crash counts.
- Adds clearer fuzzing campaign summaries, including crash paths, coverage, corpus details and validation handoff artefacts.
- Deduplicates fuzzing crashes by root cause/stack hash.
- Replays crash inputs against ASAN/debug sibling builds where available.
- Moves radare2 binary understanding from packages/fuzzing into shared packages/binary_analysis. This could help wider projects coming in RAPTOR.
- Emits reusable context-map.json / binary-context-map.json artefacts for fuzzing, /understand, validation and future binary workflows.
- Keeps backwards compatibility for existing fuzz imports.
- Makes radare2 availability and execution explicit in the campaign plan and logs.

**Validation**

Validated with focused unit tests and live fuzzing against the vulnerable testbench, where the agentic corpus materially improved coverage and crash discovery.
Screenshot 2026-05-15 at 11 04 16


grokjc:
Just landing a few fixes on branch before we merge (more to come)

libexec/raptor-binary-understand
grokjc, line 1: I added the trust marker stuff to save you a job 😄

packages/binary_analysis/radare2_understand.py
```diff
@@ -0,0 +1,608 @@
...
+        """Run the full analysis pipeline.
+
+        max_decompile bounds the number of high-priority functions we ask
+        the decompiler for, since decompilation is the slowest step.
+        """
+        import r2pipe
+
+        ctx = BinaryContextMap(binary_path=self.binary)
+        logger.info(f"radare2 analysis: opening {self.binary}")
+        r2 = r2pipe.open(str(self.binary), flags=["-2"])  # -2: silence stderr
```
grokjc: radare2 process not sandboxed. This runs radare2 directly against the target binary. Historical CVEs parsing malformed binaries (e.g. heap overflows in ELF parser). On a hostile target this is real exposure as every other binary-touching surface in RAPTOR (libfuzzer_runner, crash_analyser) goes through core.sandbox.run.

This needs some design work - on it!

packages/binary_analysis/radare2_understand.py
```diff
@@ -0,0 +1,608 @@
...
+        try:
+            r2.cmd("aaa")    # full auto-analysis
```
grokjc: Might be worth having a timeout here. If r2.cmd_with_timeout isn't available we could use a subprocess timeout.

packages/binary_analysis/radare2_understand.py
```diff
@@ -0,0 +1,608 @@
...
+            self._extract_metadata(r2, ctx)
+            self._extract_imports_exports(r2, ctx)
+            self._extract_functions(r2, ctx)
+            self._extract_entry_points(ctx)
+            self._extract_strings(r2, ctx, limit=max_strings)
+            self._tag_dangerous_callers(r2, ctx)
+            self._decompile_priorities(r2, ctx, limit=max_decompile)
+            if self.llm:
+                self._llm_prioritise(ctx)
```
grokjc, line 423: I already pushed a commit wrapping the prompt content with wrap_tool_result here

raptor_agentic.py
```diff
@@ -211,6 +212,370 @@ def stream_output(pipe, storage, prefix=""):
...
+                stdout_path.write_bytes(e.stdout or b"")
+                stderr_path.write_bytes(e.stderr or b"")
+                entries.append({
+                    "binary": str(candidate),
+                    "returncode": "timeout",
+                    "stdout": str(stdout_path),
+                    "stderr": str(stderr_path),
+                    "reproduced": True,
+                })
+            except Exception as e:
```
grokjc: Suggest narrowing to (OSError, subprocess.SubprocessError, ValueError) and re-raising anything else.

packages/fuzzing/harness_generator.py
```diff
@@ -0,0 +1,264 @@
...
+        spec: HarnessSpec,
+        language: str = "cpp",
+    ) -> str:
+        compiler = "clang++" if language == "cpp" else "clang"
+        includes = " ".join(f"-I{p}" for p in spec.include_paths) if spec.include_paths else ""
+        lib_link = f"-l{spec.library_name}" if spec.library_name else ""
+        sanitisers = "-fsanitize=fuzzer,address,undefined"
+        opt = "-g -O1"
+
+        return (
```
grokjc, line 254: Probably need shlex.quote here to avoid command injection

packages/binary_analysis/radare2_understand.py
```diff
@@ -0,0 +1,608 @@
...
+            )
+            info = FunctionInfo(
+                name=name,
+                address=int(addr or 0),
+                size=int(raw.get("size", 0) or 0),
+                type=str(raw.get("type", "fcn")),
+                is_imported=name.startswith(("sym.imp.", "imp.")),
+                is_exported=name in ctx.exports,
+            )
+            ctx.interesting_functions.append(info)
```
grokjc: interesting_functions is the full function inventory, not a curated list. 

Rename to ```functions: List[FunctionInfo]```, or filter at population (e.g. exclude sym.imp.*, exclude functions with size < 8 bytes, etc, etc) ?

packages/binary_analysis/radare2_understand.py
```diff
@@ -0,0 +1,608 @@
...
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+logger = logging.getLogger(__name__)
+
+
+# Functions that are high-value sinks for fuzzing -- if the binary
+# imports any of these, they are interesting to trace flows toward.
+_DANGEROUS_IMPORTS = {
```
grokjc: We could have a module with these constants so the whole codebase has a list of such things?

Also 'malloc', read, etc is going to lead to a lot of FPs?

packages/binary_analysis/radare2_understand.py
```diff
@@ -0,0 +1,608 @@
...
+            if not text:
+                continue
+            if len(text) < 4 or len(text) > 200:
+                continue
+            strings.append(text)
+            if len(strings) >= limit:
+                break
+        ctx.strings_sample = strings
+
+    def _tag_dangerous_callers(self, r2, ctx: BinaryContextMap) -> None:
```
grokjc, line 557: One from Claude:

  No transitive call-graph analysis. A function is marked calls_dangerous only if it directly calls a flagged import. But the interesting fuzz targets are often N hops from the sink — a parse_message routine builds a struct that gets passed through 2-3 internal helpers before reaching strcpy. Without a transitive walk, those parse routines never appear in fuzz_priorities even though they're exactly what the fuzzer should hammer.
  
r2 already has the data: aflcj (callee graph) or axt @ <addr> (xrefs-to) lets you do a 2-3 hop reachability pass cheaply. Suggest adding a second flag transitively_reaches_dangerous: bool with a depth cap of 3  — flag any function that can reach a dangerous import within that depth. Surface both flags in fuzz_priorities ranking.

grokjc (commented):
I like it 😄

grokjc:
rebased onto main so we pick up the sandbox changes (amongst other things)

packages/binary_analysis/radare2_understand.py
grokjc: Sorted sandbox with [a555e44](https://github.com/gadievron/raptor/pull/488/commits/a555e443e762fe7f626e3012143ffc959bb1ee65)

grokjc:
ok, that's the libexec stuff sorted and r2 running sandboxed, tests passing again 👍

packages/binary_analysis/radare2_understand.py
grokjc: sorted in [54f9001](https://github.com/gadievron/raptor/pull/488/commits/54f900145ac7134b53d56066ee7ce300073e50a7)

grokjc:
Function taxonomy module created + added.

packages/binary_analysis/radare2_understand.py
grokjc: Addressed in #530.

packages/binary_analysis/radare2_understand.py
grokjc: Done in #532

raptor_agentic.py
grokjc: Resolved in #536

packages/fuzzing/harness_generator.py
grokjc: Resolved in #536

packages/binary_analysis/radare2_understand.py
grokjc: Resolved in #536

grokjc:
ffuf restored.

PR #489, opened 2026-05-15
feat(web): add opt-in ffuf discovery

### LINKED ISSUE #67: feat. Addition of ffuf
We can't have a web app function without https://github.com/ffuf/ffuf, so it's time we integrate it

Hinotoi-agent:
## Summary
- add an opt-in `ffuf` runner for RAPTOR web scans, enabled only with `--ffuf-wordlist`
- scope-check ffuf URL templates against the configured target origin and invoke ffuf via argv, not a shell
- save raw ffuf JSON to `ffuf_results.json` and include a compact, redacted summary in `web_scan_report.json`
- add CLI controls for ffuf binary, path template, threads, rate, timeout, and report result limit
- add focused tests for scoping, command construction, validation, redaction, result limiting, and scanner integration

Closes #67.

## Safety / scope notes
- ffuf does not run unless the operator provides `--ffuf-wordlist`
- relative templates are anchored to the scanner base URL
- absolute templates must match the configured target origin
- generated commands use `subprocess.run([...])` without `shell=True`
- the embedded report summary redacts common secret-bearing URL parameters by default; raw ffuf JSON remains available as the scanner artifact

## Validation
- `python3 -m pytest packages/web/tests/test_ffuf.py packages/web/tests/test_scanner_none_llm.py` → 19 passed, 1 local LibreSSL/urllib3 warning
- `python3 -m pytest packages/web/tests` → 49 passed, 1 local LibreSSL/urllib3 warning
- `git diff --check`
- `python3 -m py_compile packages/web/ffuf.py packages/web/scanner.py packages/web/tests/test_ffuf.py packages/web/tests/test_scanner_none_llm.py`
- `python3 packages/web/scanner.py --help | grep -A10 -B2 ffuf`

### DIFF
--- a/packages/web/ffuf.py
+++ b/packages/web/ffuf.py
@@ -190,7 +190,7 @@ class FfufRunner:
             "reported_result_count": len(summarized_results),
             "omitted_result_count": max(0, len(results) - len(summarized_results)),
             "results": summarized_results,
-            "stderr": self._redact(completed.stderr.strip()),
+            "stderr": self._redact((completed.stderr or "").strip()),
         }
 
     def _summarize_result(self, result: dict[str, Any]) -> dict[str, Any]:
--- a/packages/web/ffuf.py
+++ b/packages/web/ffuf.py
@@ -1,0 +1,204 @@
+#!/usr/bin/env python3
+"""Narrow ffuf integration for RAPTOR web scans.
+
+The runner is intentionally small and opt-in: operators must provide a
+wordlist, and RAPTOR constrains the ffuf URL template to the configured target
+origin before spawning the external binary.
+"""
+
+from __future__ import annotations
+
+import json
+import shutil
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+from urllib.parse import urljoin, urlparse
+
+from core.logging import get_logger
+from core.sandbox import run_untrusted
+from core.security.redaction import redact_secrets
+
+logger = get_logger()
+
+
+@dataclass(frozen=True)
+class FfufConfig:
+    """Configuration for an explicit ffuf content-discovery run."""
+
+    wordlist: Path
+    path_template: str = "FUZZ"
+    threads: int = 10
+    rate: int | None = None
+    timeout: int = 30
+    max_runtime: int = 300
+    report_limit: int = 50
+    binary: str = "ffuf"
+    auto_calibration: bool = True
+    match_status: str | None = "200,204,301,302,307,401,403,405,500"
+    filter_status: str | None = "404"
+    filter_size: int | None = None
+
+
+class FfufRunner:
+    """Run ffuf against a single in-scope target origin."""
+
+    def __init__(self, base_url: str, out_dir: Path, reveal_secrets: bool = False):
+        self.base_url = base_url.rstrip("/")
+        self.out_dir = out_dir
+        self.reveal_secrets = reveal_secrets
+
+    def _origin(self, url: str) -> tuple[str, str, int]:
+        parsed = urlparse(url)
+        default_port = 443 if parsed.scheme == "https" else 80
+        return (
+            parsed.scheme.lower(),
+            (parsed.hostname or "").lower(),
+            parsed.port or default_port,
+        )
+
+    def _redact(self, value: object) -> str:
+        return redact_secrets(value, reveal_secrets=self.reveal_secrets)
+
+    def build_url_template(self, path_template: str) -> str:
+        """Build and scope-check the ffuf URL template.
+
+        ffuf marks the replacement point with ``FUZZ``. Accepting a raw URL from
+        the CLI without checking it would let a saved RAPTOR config or copied
+        command accidentally aim ffuf at a different host. Treat the template
+        like WebClient paths: relative paths are anchored to ``base_url``;
+        absolute URLs are allowed only when their normalized origin matches.
+
+        ``urljoin`` intentionally normalizes ``..`` segments before the origin
+        check. That can move a relative template outside the base path while
+        staying on the same origin; this integration scopes ffuf to the target
+        host/origin rather than to a specific subpath.
+        """
+        if "FUZZ" not in path_template:
+            raise ValueError("ffuf path template must include FUZZ")
+
+        url_template = urljoin(self.base_url + "/", path_template)
+        probe_url = url_template.replace("FUZZ", "raptor-scope-probe")
+        if self._origin(probe_url) != self._origin(self.base_url):
+            raise ValueError(
+                "ffuf path template is outside configured target scope: "
+                f"{self._redact(probe_url)}"
+            )
+        return url_template
+
+    def build_command(self, config: FfufConfig, output_file: Path) -> list[str]:
+        """Return argv for a safe, non-shell ffuf invocation."""
+        if not config.wordlist.is_file():
+            raise FileNotFoundError(f"ffuf wordlist not found: {config.wordlist}")
+        if config.threads < 1:
+            raise ValueError("ffuf threads must be >= 1")
+        if config.rate is not None and config.rate < 1:
+            raise ValueError("ffuf rate must be >= 1 when set")
+        if config.timeout < 1:
+            raise ValueError("ffuf timeout must be >= 1")
+        if config.max_runtime < 1:
+            raise ValueError("ffuf max runtime must be >= 1")
+        if config.report_limit < 0:
+            raise ValueError("ffuf report limit must be >= 0")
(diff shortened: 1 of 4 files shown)


grokjc:
Nice clean foundation — security primitives are good (argv subprocess, origin scope check, host-suffix/protocol-relative test cases, `redact_secrets` plumbing). Two structural asks before this is mergeable, plus a roadmap suggestion.
  
  ### 1. Run ffuf in the sandbox
  
  RAPTOR routes all external-tool invocations through `core.sandbox` — `codeql` (#418), `semgrep` (#440), `git clone` (`core/git/clone.py`), `cve-diff` forge (#441), `sca` resolvers all use it. ffuf should too.  It's *more* important here, not less, because ffuf deliberately makes egress: the application-layer origin check is great defence-in-depth, but the OS-layer egress allowlist is the belt to that braces.
  
  The closest existing analog is `core/git/clone.py:175-185` — it uses `run_untrusted()` with `use_egress_proxy=True` and `proxy_hosts=[…]`. ffuf is the same shape: a deliberate-egress subprocess whose only legitimate destination is a known host. Sketch:
  
      from core.sandbox import run_untrusted
      from urllib.parse import urlparse
  
      target_host = urlparse(self.base_url).hostname.lower()
  
      proc = run_untrusted(
          cmd,
          target=str(config.wordlist.parent),  # readable scope (wordlist dir)
          output=str(self.out_dir),            # writable scope (ffuf_results.json)
          readable_paths=[str(config.wordlist.parent)],
          use_egress_proxy=True,
          proxy_hosts=[target_host],           # already origin-scope-checked above
          caller_label="web-ffuf",
          timeout=config.max_runtime,          # add to FfufConfig
          capture_output=True,                                                                                                                                                                                     
          text=True,
      )                                                                                                                                                                                                            
                                                                
  What this buys: filesystem reads pinned to the wordlist dir + system libs (no `~/.aws/credentials`, no `~/.ssh`), writes pinned to `out_dir`, network egress pinned to the configured target host via the egress proxy with seccomp-blocked UDP (no DNS exfil), no shell exec / fork-bomb / setuid escalation, and a `proxy-events.jsonl` audit trail of every CONNECT ffuf attempted. Also auto-aligns with the planned MITM mode - once that lands, ffuf-in-sandbox + `-replay-proxy` becomes a clean Burp-style intercept pipeline for free.
                                                                                                                                                                                                                   
Three subtleties worth knowing:                               
  
  - `run_untrusted()` forces `block_network=True` at the namespace level (strips all interfaces). The egress proxy works *alongside* that — child sees no real interfaces, just `HTTPS_PROXY=127.0.0.1:<port>`.  Don't flip to `sandbox()` with `block_network=False` thinking you need to "enable network".
  - `proxy_hosts=[target_host]` is hostname-only. The proxy matches by hostname regardless of port; no port logic needed at the call site.                                                                         
  - ffuf must be findable inside the sandbox. The sandbox bind-mounts host `/usr/bin`, `/usr/local/bin`, etc. read-only, so a packaged `ffuf` from those paths just works. If it's installed somewhere weird (e.g. `~/go/bin/ffuf`), add `tool_paths=[str(Path(shutil.which("ffuf")).parent)]`.                                                                                                                                     
   
  ### 2. The ffuf surface is too thin to be useful as-shipped                                                                                                                                                      
                                                                
  ffuf has ~30 useful flags; the PR exposes 7. For directory discovery against any real target, two omissions make the output unreadable on the first run:                                                         
   
  - **`-mc` / `-fc` status code matchers/filters** — without these the report is hundreds of 404s.                                                                                                                 
  - **`-ac` auto-calibration** — without this, any site with a templated soft-404 page (Rails, Django, Express, …) produces thousands of false positives.
                                                                                                                                                                                                                   
  Please add at least these two before merge. `-fs`/`-fw`/`-fl` size/word/line dedup is near-essential too (one extra `--ffuf-filter-size` flag covers most cases).                                                
                                                                                                                                                                                                                   
  ### 3. Roadmap suggestion — file companion issues                                                                                                                                                                
  
  Even with (1) and (2) addressed, this PR is only directory discovery. Worth tagging it as such so we/others can open companion issues so the slices are reviewable independently:
  
  | Slice | Flags | Use case |
  |---|---|---|
  | Parameter discovery | `-X POST/PUT/...`, `-d`, `-H Content-Type` | API endpoint enumeration |
  | Authenticated fuzzing | `-H Authorization`, `-H Cookie`, `-b` | Behind-login surface (where the real bugs live) |
  | Recursion | `-recursion`, `-recursion-depth`, `-recursion-strategy` | Tree walking |
  | Replay proxy | `-replay-proxy` | Pairs with the planned MITM mode |
  | Multi-wordlist | `-w a:K1 -w b:K2`, `-mode pitchfork/clusterbomb` | Combinatorial fuzzing |
  | Stop conditions | `-maxtime`, `-maxtime-job`, `-sf` | Operational + DoS-against-self guard |
  | Content matchers | `-mr`, `-fr`, `-mt`, `-ft` | Regex / response-time |
  | Vhost discovery | `-H "Host: FUZZ.target"` | Single-IP virtual-host enum (distinct from URL-hostname FUZZ, which you correctly block) |
  
  Each would be a small, self-contained PR in the future (and that list isn't exhaustive!)
  
  ### Minor
  
  - `assert "shell=True" not in captured["cmd"]` is checking literal-string presence in the argv list — that string is never going to be there regardless. Better: assert against captured call kwargs.
  - No Python-side `subprocess.run(..., timeout=…)` — ffuf's `-timeout` is per-request; a hung ffuf process hangs the scanner. Belt-and-braces.
  - `urljoin` permits `..` segments, so `../../FUZZ` escapes the configured subpath but stays same-origin — fine if the scope is "host", but worth a docstring note saying so.
  - `redact_secrets` signature matches (`core/security/redaction.py:158`) ✓.
  - Rebase needed — PR is 36 commits behind main, `packages/web/scanner.py` was reformatted.

Hinotoi-agent:
Updated in 4193eb6, thanks for the detailed review.

Changes made:
- rebased onto current main and resolved the scanner wiring conflict while preserving both crawl limits and ffuf config wiring
- routed ffuf through `run_untrusted()` with egress proxying, target-host allowlisting, wordlist read scope, output write scope, tool path binding, and `max_runtime`
- added `-ac`, `-mc`, `-fc`, and `-fs` support with CLI wiring
- updated tests to assert the sandbox call kwargs instead of checking for a literal `shell=True` argv entry
- kept the URL-template docstring explicit about same-origin `..` normalization

Local verification:
- `uv run python -m pytest packages/web/tests/test_ffuf.py packages/web/tests/test_scanner_none_llm.py -q` -> 19 passed
- `git diff --check`

The GitHub checks on the pushed commit are now passing/skipped as expected.

grokjc:
Thanks!

Only one real issue I can see:
 - FfufConfig now has match_status, filter_status, filter_size, auto_calibration, and max_runtime as overridable fields but main() in scanner.py only wires the 7 original flags. Operators running the scanner CLI get the defaults baked in with no way to override from the command line. Defaults are reasonable, so it's not catastrophic  but the moment someone hits a templated soft-404 that escapes -ac, they'll want --ffuf-filter-size 1234, and there's no way to pass it without editing code. We should either add --ffuf-mc, --ffuf-fc, --ffuf-fs, --ffuf-no-autocal, --ffuf-max-runtime now, or call that out explicitly as next-PR.

Two minor things:
  - completed.stderr.strip() would AttributeError if stderr is ever None; (completed.stderr or "").strip() is safer.
  - The parametrize for test_build_command_rejects_invalid_numeric_options didn't get extended with filter_size: -1 or max_runtime: 0 cases for the new validators.

Hinotoi-agent:
Updated in `32a9d56`.

## Changes made
- Added scanner CLI overrides for the new ffuf tuning fields:
  - `--ffuf-max-runtime`
  - `--ffuf-no-auto-calibration`
  - `--ffuf-match-status`
  - `--ffuf-filter-status`
  - `--ffuf-filter-size`
- Factored scanner CLI parsing/config construction into testable helpers.
- Covered full option wiring, including empty match/filter status values to omit `-mc` and `-fc`.
- Made ffuf stderr handling tolerate `None` via `(completed.stderr or "").strip()`.
- Extended invalid numeric option coverage for `filter_size=-1` and `max_runtime=0`.

## Local verification
```bash
python3 -m pytest packages/web/tests/test_ffuf.py -q
# 17 passed

python3 -m pytest packages/web/tests/test_scanner_none_llm.py packages/web/tests/test_client_scope.py -q
# 14 passed

python3 -m compileall -q packages/web core/security *.py
git diff --check
```

grokjc:
LGTM. Look forward to some followups 😃 Thanks!

PR #498, opened 2026-05-16
test(web): honor scanner crawl limits

Hinotoi-agent:
## Summary
- Thread scanner-level `--max-depth` / `--max-pages` controls into `WebCrawler` instead of always using crawler defaults.
- Add a local fake-target regression test proving `WebScanner(max_depth=0, max_pages=1)` only requests the seed page and persists bounded crawl artifacts.

## Why
The CLI exposes crawl bounds and prints them, but `WebScanner` constructed `WebCrawler(self.client)` without passing those options. That made scanner-level limits ineffective, so small fake-target scans could crawl more pages than requested.

## Validation
- `python3 -m pytest packages/web/tests/test_scanner_fake_target.py -q`
- `python3 -m pytest packages/web/tests/test_scanner_fake_target.py packages/web/tests/test_scanner_none_llm.py packages/web/tests/test_client_scope.py -q`
- `python3 -m pytest packages/web/tests -q`
- `python3 -m compileall -q packages/web`
- `git diff --check`

### DIFF
--- a/packages/web/scanner.py
+++ b/packages/web/scanner.py
@@ -144,7 +144,7 @@ Examples:
     parser.add_argument("--url", required=True, help="Target web application URL")
     parser.add_argument("--out", help="Output directory for results")
     parser.add_argument("--max-depth", type=int, default=3, help="Maximum crawl depth (default: 3)")
-    parser.add_argument("--max-pages", type=int, default=50, help="Maximum pages to crawl (default: 50)")
+    parser.add_argument("--max-pages", type=int, default=100, help="Maximum pages to crawl (default: 100)")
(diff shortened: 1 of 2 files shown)


grokjc:
One thing worth aligning before merge (left as an inline comment): the new max_pages default in WebScanner.\_\_init__ is 100, but the CLI's --max-pages default at scanner.py:121 is 50. Pre-PR this didn't matter because the value was ignored; post-PR, a CLI run without the flag gets 50 while a direct Python caller gets 100. Let's pick one (I'd say 100 since that matches WebCrawler) and use it in both places.

Hinotoi-agent:
Aligned the CLI default with `WebScanner`/`WebCrawler` at 100 and updated the help text. Local `python3 -m pytest packages/web/tests -q` passes (36 passed).

grokjc:
Thanks! LGTM.

PR #565, opened 2026-05-21
 fix(smt/path-validator): dedup textually-identical call subterms within one batch

unprovable:
Pre-fix _substitute_calls allocated a fresh `_anon_N` for every balanced `<ident>(...)` match, so a path like `[strlen(input) > 100, strlen(input) < 50]` encoded `as _anon_0 > 100 AND _anon_1 < 50` — trivially satisfiable when both `strlen(input)` texts were meant to denote the same value. Tier-4 SMT refinement then upgraded these inconclusive findings to "feasible with witness" against the LLM's writer intent.

Use the existing `anon_map` (placeholder -> call-text) as a dedup oracle; identical call substrings inside one `check_path_feasibility` batch share one placeholder, whether they appear in the same condition string or across conditions.  Any `impure-call` default is preserved at batch boundaries — fresh `check_path_feasibility` calls reset `vars_ and anon_map` — because that's the only granularity at which the LLM has no syntactic way to get shared identity.

The verb path (`smt_verbs._build_operand_bv` → `make_anon_call_var`) still allocates per-call without text-based dedup; tracked in the `make_anon_call_var` docstring as a follow-on.

Tests: +5 covering intra-condition, cross-condition, cross-batch reset, distinct-text retention, and whitespace/argument distinction. Other tests replaced in line with the above.


grokjc:
How do we cope with something like `[strlen(input) > 100, input = realloc(input, ...), strlen(input) < 50]` where the LLM summarises the conditions but the path actually mutates the argument between calls? 🤔

unprovable:
You are asking the good questions... ok, setting a beady eye and a clanger's soul on it.

grokjc:
Thanks, that is a good solution and an elegant one at that.

**One observation re regex:** the assignment-shape coverage is a mix of designed and incidental, since several Python/Java compound operators contain a designed match as a substring. 

  | operator | matched? | how |
  | --- | --- | --- |
  | bare `=` (C/C++/Java/Python assignment) | yes | designed (bare-`=` alternation with `(?<![=<>!])(?!=)` guards) |
  | `+= -= *= /= %= &= \|= ^=` | yes | designed (`[+\-*/%&\|^]=`) |
  | `<<= >>=` | yes | designed |
  | `++ --` | yes | designed |
  | `:=` (Python walrus / assignment expression) | yes | **incidental** — `:` isn't in the bare-`=` lookbehind, so the bare-`=` alternation matches. Reasonable since walrus binds, but not by our deliberate design. |
  | `**=` (Python pow-assign) | yes | **incidental** — trailing `*=` matched |
  | `//=` (Python floordiv-assign) | yes | **incidental** — trailing `/=` matched |
  | `>>>=` (Java unsigned-right-shift-assign) | yes | **incidental** — trailing `>>=` matched |
  | `@=` (Python matrix-mul-assign, PEP 465) | **no** | not in regex |

  The incidental matches mean a future "simplify this regex" refactor that drops e.g. the `*` from `[+\-*/%&|^]` would silently regress Python `**=` coverage along with the obvious `*=`. The exclusion-side guards `(?<![=<>!])=(?!=)` for `==`/`<=`/`>=`/`!=` are the other half of the same fragility — well-tested on the inclusion side, decays silently on the exclusion side.

  `@=` is the only operator with no current coverage; very unlikely to appear in path conditions surfaced by this validator.

  **Minor:** `--x` (numeric double-negation as a condition) routes to ASSIGNMENT_SHAPED. Astonishingly rare; flagging the behaviour.

Might be worth adding extra docstrings just to make sure there are no surprises later? But this is looking good and I'm being pedantic 😄

grokjc:
Thanks! LGTM

Issue #485, opened 2026-05-13
Appreciation


Splinters-io:
I've had some bangers with raptors help, there should be a wall or a way to attribute issues found / co-authored / guided you name it 

maybe even a buy raptor a coffee link or something

danielcuthbert:
I love this idea and I love the fact you found bangers too ;) I'm open to suggestions

unprovable:
This isn't a kofi link, but the easy way to add a CITATION.cff file that lets someone easily reference the repo in a 'known format' way. See PR #573 .

grokjc:
I wonder if we have a Hall of Fame (or Hall of Shame 😄) where we list CVE, Author, Description, etc for stuff that was found with Raptor. People can submit PRs to add their stuff if they wanted to ...? 🤔

Splinters-io:
I’d be happy to list my recent CVE’s with my mildly modified variant I have a blogpost speaking to it but need vendor to give me the nod Wall of fame or I’ll feel even more attribution guilt 😂Perhaps just a simply name, social, CVE / ID, Blogpost ?

Splinters-io:
I had a really long post (Raptor Advert) but having spoken to vendor this is all I could get away with https://thecontractor.io/cve-2026-34910/

PR #573, opened 2026-05-22
(FEAT) Add CITATION.cff for RAPTOR project

unprovable:
After issue #485 it's probably worth having this so it can be properly referenced. M.

### DIFF
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -34,6 +34,7 @@ authors:
     family-names: Evron
   - given-names: Daniel
     family-names: Cuthbert
+    orcid: "https://orcid.org/0009-0000-6443-0251"
   - given-names: Thomas
     family-names: Dullien
     alias: "Halvar Flake"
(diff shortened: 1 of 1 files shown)


unprovable:
@danielcuthbert technically this is a feat note...

grokjc:
Not sure if anyone has an ORCID, could include those too ('orcid') and/or github handles in 'alias'. But do folks want to cite RAPTOR or us? 🤔

unprovable:
Ah, it's an easy attribution method. Can't hurt, especially with lots of papers about 'AI cyber cyber' floating around. And yes, just remembered @danielcuthbert has an ORCID, added it.

danielcuthbert:
@thanks @unprovable please also add yourself here too

grokjc:
Seeing as all the cool kids have one: https://orcid.org/0009-0004-3478-3207

Issue #583, opened 2026-05-23
Extend the L1 source table to include C-level inputs (read/recv/fgets/argv/env) on /understand and  /validate


dinosn:
Hello, 

   One suggestion on improving Raptor with the following additions from vbsec, [https://github.com/tanviet12/vbsec], which has some nice concepts that maybe worth porting, one is the post data classification to /understand, which will add shared L1-L4 trust vocabulary to /understand + /validate. 

Extend the L1 source table to include C-level inputs (read/recv/fgets/argv/env), which will change the state as:

<img width="2076" height="470" alt="Image" src="https://github.com/user-attachments/assets/4e18ad21-9a31-4063-84da-4efd276c0ca4" />

Regards,
Nicolas

dinosn:
Adding 2 files as per suggestions, one data-flow-classification.md the above method and a second new detection suggested for Ai 'slop' creations slopsquatting.md. 

[data-flow-classification.md](https://github.com/user-attachments/files/28174792/data-flow-classification.md)

[slopsquatting.md](https://github.com/user-attachments/files/28174796/slopsquatting.md)

grokjc:
I see four pieces in `data-flow-classification.md`:

### 1. Four-Level Trust Taxonomy
- **L1**: Attacker-controlled
- **L2**: Persistent storage, semi-trusted
- **L3**: Internal code/config
- **L4**: System/runtime constants

### 2. Per-Framework Source Enumeration
Function-name granularity for:
- Express, Fastify, Next, NestJS, Flask, Django, FastAPI, Go-net-http, Gin, Echo, Fiber, PHP, Laravel, Rails, Spring entry methods
- C-syscall extension: `read`, `recv`, `fgets`, `argv`, `getenv`, `ioctl`, `copy_from_user`, etc.

### 3. L2 Cross-Class Rule
- L2 in same-class sink (DB→DB parameterized) = **safe**
- L2 cross-class (DB→HTML, DB→shell, DB→URL, DB→path) = **escalates to L1**

### 4. Five-Item Sanitization Checklist
- [ ] Parameterization
- [ ] Escape-encode
- [ ] Whitelist-validation
- [ ] Length-check + bounds-clamp
- [ ] Type-narrowing-cast

**Combination Rule**: L1 ⊕ L3 ⊕ L4 → L1 (highest wins), threads through all four

---

## RAPTOR's Current Coverage

### ✅ What RAPTOR Already Has

| Feature | Location | Details |
[...]

PR #615, opened 2026-05-24
feat(source-intel): surface C-level input sources

Hinotoi-agent:
## Summary

- Adds informational C/C++ L1 source observations for fd/socket/stream input calls, process/environment inputs, IPC inputs, kernel user-copy reads, and device-control entry points.
- Reuses the reviewed `core.function_taxonomy` function-name catalog instead of introducing a parallel source catalog.
- Keeps source-only ubiquitous reads such as `read`/`fread` as a scanner-local union, because they are useful at source call sites but deliberately excluded from import/fuzz-priority categories.
- Renders those observations into `/understand` and `/validate` evidence strings so candidate flows can carry explicit source/taint context.
- Keeps the change evidence-only: no verdict policy changes, no L2/L3/L4 combination rules, and no fuzz-priority policy changes.

Refs #583.

## Issues covered

- Extends the L1 source table to include common C/C++ user-controlled inputs that are visible in source, including fd/socket/stream reads and process/environment sources.
- Retargets the implementation after maintainer feedback so shared function-name lists come from `core.function_taxonomy`:
  - `STREAM_INPUT_FUNCS`
  - `NETWORK_INGEST_FUNCS`
  - `PROCESS_BOUNDARY_FUNCS`
  - `KERNEL_USERSPACE_FUNCS`
  - `IPC_FUNCS`
  - `SCAN_FAMILY_FUNCS`
- Adds `ioctl` / `unlocked_ioctl` / `compat_ioctl` to the shared function taxonomy as `DEVICE_CONTROL_FUNCS`.
- Keeps `argv` / `envp` / `environ` scanner-local because they are identifiers, not function names.
- Complements existing kernel/user boundary evidence instead of replacing it.

## Review notes

- The scanner is bounded to C/C++ files and capped at 500 files, matching the lightweight sidecar posture used elsewhere in `source_intel`.
- The helper scans files deterministically and avoids recording examples inside comments, strings, character literals, and simple function prototypes.
- Unit coverage exercises the scanner directly rather than depending on local `spatch` availability.
- This intentionally does not implement the broader L2 cross-class rule or sanitization-checklist semantics from the issue thread; those should remain separate follow-up work.

## Type of change

- [x] Feature / evidence enrichment
- [x] Tests
- [x] Maintainer-feedback adjustment
- [ ] Documentation-only change

## Validation

- [x] `uv run python -m pytest packages/source_intel/tests/test_c_level_sources.py packages/source_intel/tests/test_axis4_boundary_lsm.py packages/source_intel/tests/test_render.py core/function_taxonomy/tests/test_function_taxonomy.py -q`
- [x] `uv run python -m compileall -q core/function_taxonomy packages/source_intel`
- [x] `uv run ruff check core/function_taxonomy/__init__.py core/function_taxonomy/tests/test_function_taxonomy.py packages/source_intel/analyze.py packages/source_intel/render.py packages/source_intel/tests/test_c_level_sources.py`
- [x] `git diff --check`


Hinotoi-agent:
Thanks — I updated this PR after reviewing the discussion on #583 and wanted to call out exactly what changed from the initial version.

## What changed in this PR

- Added C/C++ L1 source evidence to the `/understand` and `/validate` source-intel context so RAPTOR can surface concrete input-origin hints instead of only showing sink/import-oriented signals.
- Added a bounded deterministic scanner for common C/C++ input APIs, including fd/socket reads, stdio-style reads, argv/env access, environment lookup, device-control inputs, and kernel/user boundary helpers.
- Added rendering for these findings as `C/C++ L1 source` lines in source-intel output.
- Added regression coverage for both positive detections and false positives, including comments, strings, character literals, prototypes, and simple declarations.

## Improvements made after reviewing #583 comments

- Moved the source API list into a neutral `core.input_taxonomy` module instead of keeping it embedded in `packages/source_intel/analyze.py`. This follows the issue discussion that source-side taxonomy should be separate from existing sink/import-focused taxonomy.
- Kept the implementation informational-only: it improves the evidence available to analysts and prompt context, but does not change verdicts or exploitability scoring by itself.
- Expanded the catalog beyond the initial issue examples (`read`, `recv`, `fgets`, `argv`, `env`) to include additional C/kernel boundary inputs mentioned in discussion, such as `ioctl`, `copy_from_user`, and related helpers.
- Kept the PR narrowly scoped to L1 source surfacing so follow-up L2/L3/L4 correlation rules can be reviewed separately.

## Validation

I reran the focused validation after the update:

```bash
uv run python -m pytest packages/source_intel/tests/test_c_level_sources.py packages/source_intel/tests/test_axis4_boundary_lsm.py packages/source_intel/tests/test_render.py -q
uv run python -m compileall -q core/input_taxonomy packages/source_intel
uv run ruff check core/input_taxonomy/__init__.py packages/source_intel/analyze.py packages/source_intel/render.py packages/source_intel/tests/test_c_level_sources.py
git diff --check
```

Current PR head: `3c8343dd4f76b0e4cfdf8e0315becc2b61795448`.

grokjc:
Unfortunately this clashes with work in progress and the PR as-is duplicates #585 and other private work. This is the fun of trying to react to a new Issue before the design is fully complete (or at least public).

  One important piece of context that isn't your fault — it wasn't visible on the issue: the L1 source catalog itself already landed  in core/function_taxonomy/ (PR #585), and we extended it further across #588/#589/#590. We chose to put it there (rather than a new module) because that module's existing shape — frozensets of security-relevant function-name categories — already fit, and #583's suggestion of a separate core/input_taxonomy/ was superseded by that decision. We never mirrored that back to #583, so you reasonably built against the older intent. Apologies for the gap - but it hasn't been 24 hours yet since we first started work there! 😄 

  The result is that core/input_taxonomy/ overlaps the existing catalog — including a literal name collision (STREAM_INPUT_FUNCS now exists in both modules) — and the new copy is missing refinements the existing
   one already has: the full memdup_user/iovec/pages kernel set, the IPC family, scanf/fscanf kept in SCAN_FAMILY_FUNCS for disjointness, and the getenv-vs-secure_getenv/getauxval marker split.

  Could we retarget the PR to keep the mechanism, drop the duplicate catalog? Concretely:

  1. Source the function lists from core/function_taxonomy/ — STREAM_INPUT_FUNCS, NETWORK_INGEST_FUNCS, PROCESS_BOUNDARY_FUNCS, KERNEL_USERSPACE_FUNCS, IPC_FUNCS, SCAN_FAMILY_FUNCS are all already there and
  reviewed.
  2. You correctly spotted that source-side scanning wants the ubiquitous reads (read/fread/pread) that the import-table catalog deliberately excludes. That's a real distinction — the right pattern is to compose them at the scanner, exactly like packages/exploit_feasibility/constants.py:INPUT_FUNCTIONS does (_T_STREAM | frozenset({"read", "fread", ...})). So no parallel catalog needed, just a scanner-local union.
  3. ioctl/unlocked_ioctl/compat_ioctl aren't in function_taxonomy yet — those are a genuine addition. Suggest adding them there (one home for the catalog) rather than in a separate module; happy to take that as a quick follow-up if easier.
  4. argv/envp/environ are identifiers, not function names, so they legitimately stay scanner-local — function_taxonomy is a function-name catalog and shouldn't hold them. Keep that part as-is.

  Also minor: Fixes #583 → Refs #583 — the issue is a multi-part arc and this is one slice.

  Net: the scanner + render + identifier detection are the valuable core and we'd like to land them. It's just the catalog that should point at the existing source of truth instead of forking it.

Hinotoi-agent:
Thanks — I made the precision refinement on the same branch.

Follow-up update:

- kept `core.function_taxonomy` unchanged for its binary-fingerprint consumers
- added scanner-local L1-source filters so `_C_L1_SOURCE_CALLS` only labels direct ingress call sites:
  - socket keeps `recv`, `recvfrom`, `recvmsg`, `recvmmsg`, `SSL_read`, `BIO_read`
  - socket excludes `accept`, `bind`, `listen`
  - stream keeps `scanf`, `vscanf`, `fscanf`, `vfscanf`, `wscanf` plus the stream-input taxonomy entries
  - stream excludes `sscanf`, `vsscanf`, `swscanf`
- added regression coverage that confirms `wscanf` remains a stream source while the socket setup and in-memory scanf variants are not surfaced as L1 source evidence

Validation rerun locally:

```bash
uv run python -m pytest packages/source_intel/tests/test_c_level_sources.py packages/source_intel/tests/test_axis4_boundary_lsm.py packages/source_intel/tests/test_render.py core/function_taxonomy/tests/test_function_taxonomy.py -q
uv run python -m compileall -q core/function_taxonomy packages/source_intel
uv run ruff check core/function_taxonomy/__init__.py core/function_taxonomy/tests/test_function_taxonomy.py packages/source_intel/analyze.py packages/source_intel/render.py packages/source_intel/tests/test_c_level_sources.py
git diff --check
```

Result: `88 passed, 4 deselected`; compile, ruff, and diff-check passed.

grokjc:
This is exactly the retarget I was hoping for — composing from the reviewed function_taxonomy sets, read/fread kept scanner-local, ioctl family added to the shared catalog as DEVICE_CONTROL_FUNCS with disjointness/size tests, argv/envp/environ kept as identifiers. Nicely done, and thanks for the quick turnaround.
  
  **Slight precision refinement before merge** 

  Two of the sets you compose into _C_L1_SOURCE_CALLS are somewhat over-inclusive when relabelled as "L1 source", because they were originally curated for binary-capability bucketing (where "is this binary doing network I/O" is the question), not for "this call site reads attacker-controlled data":

  - NETWORK_INGEST_FUNCS → "socket" also contains accept / bind / listen. Those are socket setup / server-side markers, not data ingress — the attacker bytes arrive later via recv/read on the fd accept returns (already covered). So a bind() or listen() call site would surface as an "L1 socket source", which isn't true.
  - SCAN_FAMILY_FUNCS → "stream" also contains the string-parsing variants sscanf / vsscanf / swscanf. Those parse an in-memory buffer, not an external stream — sscanf(internal_buf, …) isn't an L1 source by  itself.

  Since this is informational evidence fed to the /understand + /validate prompt, the over-inclusion is low-harm (the LLM can disregard a spurious line), so I don't want to over-rotate on it. But the clean fix is a scanner-local subtraction — symmetric with how you already add read/fread locally — keeping the data-ingress members only:

  - socket: keep recv / recvfrom / recvmsg / recvmmsg / SSL_read / BIO_read; drop accept / bind / listen.
  - stream: keep scanf / vscanf / fscanf / vfscanf / wscanf; drop sscanf / vsscanf / swscanf.

  I'd keep the subtraction at the scanner (not in function_taxonomy) — those sets are deliberately bundled the way they are for the binary-fingerprint consumers, so splitting them upstream would ripple. A scanner-local frozenset(...) - {…} with a one-line comment is the contained fix. Please double-check the wide-char split (wscanf reads a stream, swscanf parses a string) since that pair is easy to flip.

TL;DR  "exclude accept/bind/listen from the socket tag and sscanf/vsscanf/swscanf from the stream tag, because those aren't data-ingress calls."

With that final set of changes, this will be good to merge — the scanner + render + identifier detection are exactly the value we wanted.

grokjc:
Thanks - LGTM!

Issue #617, opened 2026-05-24
More ffuf


grokjc:
As per my comments on https://github.com/gadievron/raptor/pull/489, we have more work to do in using the excellent ffuf tool to its full potential. Just capturing that here so we don't forget!

danielcuthbert:
Agreed and I see our amazing friend @Hinotoi-agent got stuck in already

grokjc:
**New capability**

- **Recursion** — `--ffuf-recursion` / `--ffuf-recursion-depth` / `--ffuf-recursion-strategy`. The template must end in `FUZZ` (we mirror ffuf's own strict `HasSuffix` rule so bad configs fail at parse time, not mid-scan), and it never ships bare: `-maxtime-job` defaults to `max(60, max-runtime/4)` and a default `-rate 50` applies when none is set.
- **Full matcher/filter family** — `-e` extensions (`--ffuf-extensions '.php,.bak'`), `-fw`/`-fl`, `-mr`/`-fr` (Go-regexp values passed through untouched — Python-side validation would be neither necessary nor sufficient), `-mt`/`-ft` for response-time matching.
- **Request shaping** — `-X` + `-d`, so POST/JSON parameter discovery works. The scope rule generalizes: the URL template must always be same-origin, but the FUZZ keyword may now live in the URL, body, a header, or a cookie value.
- **Multi-wordlist** — `--ffuf-wordlist` is repeatable with `path:KEYWORD` syntax; `--ffuf-mode clusterbomb|pitchfork`. Dead wordlists (a keyword that appears nowhere) are config errors, keywords can't collide (`FUZZ` inside `FUZZ2` would corrupt substitution), and clusterbomb shares the recursion rate guard.
- **Vhost discovery** — `--ffuf-vhost` fuzzes `Host: FUZZ.<target-host>` against a fixed URL; a custom template must stay under the target host, so a saved config can't quietly probe a sibling namespace.
- **Stop conditions** — `-sf`/`-se`/`-sa` exposed, and every run now carries ffuf's own `-maxtime`. That last one fixed a real bug: previously the only runtime bound was the Python subprocess timeout, and when it fired it killed the *whole scan* and discarded ffuf's report. Now ffuf exits cleanly and flushes; a `timed_out` field marks runs the backstop had to kill.
- **Parse-time preflight** — every ffuf config rule is validated when arguments are parsed, so a typo'd flag combination dies in under a second instead of after the crawl/fuzz phases have spent their budget.

[...]

PR #652, opened 2026-05-25
feat(web): add authenticated ffuf options

Hinotoi-agent:
## Summary
- add repeatable `--ffuf-header` and `--ffuf-cookie` options for authenticated ffuf discovery
- thread the options through scanner CLI -> `FfufConfig` -> ffuf argv
- validate header/cookie values before invoking ffuf
- keep sandboxing, target-host proxy allowlisting, and same-origin URL template checks unchanged
- redact authenticated ffuf options from the command log by default

## Safety
- ffuf remains opt-in via `--ffuf-wordlist`
- ffuf still runs through `run_untrusted` with target-host egress allowlisting
- headers/cookies are passed as non-shell argv entries
- CR/LF in header and cookie values is rejected before execution
- authorization headers and cookie values are redacted in the logged command unless `reveal_secrets` is explicitly enabled

Further improvement for #617; this is a narrow authenticated-ffuf slice and does not close the broader issue.

## Test plan
- `uv run python -m pytest packages/web/tests/test_ffuf.py packages/web/tests/test_scanner_none_llm.py -q`
- `uv run python -m pytest packages/web/tests -q`
- `uv run ruff check packages/web/ffuf.py packages/web/scanner.py packages/web/tests/test_ffuf.py`
- `git diff --check`

### DIFF
--- a/packages/web/ffuf.py
+++ b/packages/web/ffuf.py
@@ -17,7 +17,7 @@ from urllib.parse import urljoin, urlparse
 
 from core.logging import get_logger
 from core.sandbox import run_untrusted
-from core.security.redaction import redact_secrets
+from core.security.redaction import is_secret_field_name, redact_secrets
 
 logger = get_logger()
 
@@ -38,6 +38,8 @@ class FfufConfig:
     match_status: str | None = "200,204,301,302,307,401,403,405,500"
     filter_status: str | None = "404"
     filter_size: int | None = None
+    headers: tuple[str, ...] = ()
+    cookies: tuple[str, ...] = ()
 
 
 class FfufRunner:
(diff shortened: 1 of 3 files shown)


danielcuthbert:
I’ve pushed a small follow-up patch for the auth bits. Love this and it also introduces the ugliness that is how we work with modern auth tokens and identities. 

The original change was good, but I spotted one awkward edge: `Authorization` was only being redacted if the value matched one of the known token shapes. That meant something like `Authorization: Token abc123` could still show up in logs. Not ideal, as the header name itself tells us it is secret-ish, even if the value is weird, short, old-school, or just generally cursed.

The patch does three small things:

- always redacts `Authorization` and `Proxy-Authorization` values
- rejects malformed `--ffuf-header` input that is not in `Name: value` form
- tidies cookie redaction so spacing does not get a bit wonky

I also added regression tests for `Token`, `ApiKey`, `Digest`, and `Negotiate` style auth headers so this does not sneak back in later.

Validation on my side:

- `packages/web/tests/test_ffuf.py` passes
- full `packages/web/tests` passes
- `compileall` passes
- `git diff --check` passes

So with this patch, I’m much happier with the authenticated ffuf path. Small fix, but worth doing because auth material in logs is exactly the sort of tiny paper cut that turns into a proper headache later.

Now to think how we handle MFA/U2F and so on

PR #696, opened 2026-05-27
feat: add AWS Bedrock as an LLM provider option

owen10380:
## Summary

Adds **AWS Bedrock** as an LLM provider option for RAPTOR. Same Claude models as Anthropic-direct, but auth via SigV4 through the `boto3` credential chain instead of `ANTHROPIC_API_KEY`. Useful for AWS shops where API keys can't leave the account boundary, or where existing IAM roles already grant Bedrock access.

This is **opt-in only** — no behaviour change for existing users. Bedrock is intentionally absent from `_DEFAULT_PROVIDER_ORDER` so users with ambient `AWS_REGION` set for unrelated reasons aren't auto-promoted.

## Design

`BedrockProvider` is a subclass of `AnthropicProvider` that overrides only `__init__` to swap `anthropic.Anthropic` for `anthropic.AnthropicBedrock`. The official Anthropic SDK ships `AnthropicBedrock` with the same `.messages.create` / `.beta.messages.create` surface, so all inherited behaviour (tool use, prompt caching, structured output, cost tracking, transient-error retry) works unchanged.

```python
class BedrockProvider(AnthropicProvider):
    def __init__(self, config: ModelConfig):
        LLMProvider.__init__(self, config)  # skip parent __init__
        self.client = AnthropicBedrock(timeout=config.timeout, ...)
        # mirror parent's instructor + caching state
```

Bedrock model IDs (e.g. `us.anthropic.claude-opus-4-7`, `eu.anthropic.claude-haiku-4-5-20251001-v1:0`) are aliased to the canonical Anthropic IDs in `MODEL_COSTS` / `MODEL_LIMITS`, so cost tracking and limit enforcement work identically across both providers without duplicated price data.

## Configuration

```bash
pip install boto3 anthropic
export AWS_REGION=us-east-1     # or eu-west-1, ap-southeast-2
export AWS_PROFILE=my-profile   # optional
```

`~/.config/raptor/models.json`:
```json
{
  "models": [
    {"provider": "bedrock", "model": "us.anthropic.claude-opus-4-7", "role": "analysis"},
    {"provider": "bedrock", "model": "us.anthropic.claude-haiku-4-5-20251001-v1:0", "role": "fallback"}
  ]
}
```

Or via the existing `RAPTOR_PROVIDER_PREFER=bedrock` env var.

## What's in the diff (11 files, +697 / -6)

| File | Change |
|---|---|
| `core/llm/providers.py` | `BedrockProvider` class · factory branch · `AnthropicBedrock` import · `supports_temperature` gate in `generate()` and `generate_structured()` |
| `core/llm/detection.py` | `BEDROCK_AVAILABLE` flag · `has_bedrock` detection · misconfig warnings · config validity check |
| `core/llm/config.py` | `_build_bedrock_config` builder + registration |
| `core/llm/model_data.py` | 31 verified Bedrock model IDs aliased to Anthropic costs/limits · `MODELS_WITHOUT_TEMPERATURE` registry · `supports_temperature()` helper |
| `core/security/llm_family.py` | `provider_of()` detects Bedrock-prefixed IDs and routes to `bedrock` provider; family resolution unchanged so cross-family logic still treats Bedrock-Claude as anthropic lineage |
| `core/config/__init__.py` | Adds `AWS_REGION` / `AWS_DEFAULT_REGION` / `AWS_PROFILE` / `AWS_DEFAULT_PROFILE` / `AWS_CONFIG_FILE` / `AWS_SHARED_CREDENTIALS_FILE` to `LLM_API_KEY_VARS` so the dispatcher subprocess can authenticate |
| `packages/llm_analysis/orchestrator.py` | `_resolve_model()` exempts Bedrock from the "must have api_key" check (boto3 chain handles auth) |
| `core/llm/tests/test_bedrock_provider.py` | 55 new tests (7 classes) |
| `requirements.txt` | comment-only optional dep note |
| `README.md` | "AWS Bedrock" subsection under "Using a different LLM" |
| `docs/DEPENDENCIES.md` | Bedrock + AnthropicBedrock listed as optional tools |

## Bug fixes that came along for the ride

Two real bugs surfaced while wiring this up. Both affect Anthropic-direct users, not just Bedrock — the fixes are in this PR because they block any use of newer models:

1. **Claude Opus 4.7 deprecated the `temperature` parameter** (returns `400 BadRequest "temperature is deprecated for this model"`). `AnthropicProvider.generate()` and `.generate_structured()` always passed `temperature` unconditionally, so Opus 4.7 was unusable. Fixed via the new `MODELS_WITHOUT_TEMPERATURE` registry + `supports_temperature()` gate. Forward-compatible: future no-temperature models just need adding to the set. `turn()` already omitted temperature so it was unaffected.
2. **`LLM_API_KEY_VARS` was missing `AWS_REGION` / `AWS_PROFILE`** — the cred isolation dispatcher's subprocess could see `AWS_ACCESS_KEY_ID` etc. but not the region/profile vars boto3 needs to actually pick an endpoint. Pre-existing gap; only surfaced once a Bedrock provider tried to use them.

## Testing

**Unit tests** — 55 new tests covering:
- SDK availability flag
- Model alias map (cost/limit parity with Anthropic, parametrised across 11 IDs)
- `supports_temperature` gate (Opus 4.7 + 5 Bedrock variants reject; Sonnet 4.5/4.6/Haiku/GPT/Gemini accept; unknown models default-true)
- `_build_bedrock_config` env-driven activation (with/without `AWS_REGION`, `AWS_DEFAULT_REGION` fallback)
- Factory dispatch (returns `BedrockProvider`, raises clear error when SDKs missing)
- `provider_of` Bedrock routing (8 prefix variants → bedrock; non-Bedrock IDs unaffected)
- Orchestrator validation (env-var allowlist correctness)

**Regression** — `pytest core/llm/tests/` reports **579 existing tests still pass + 55 new = 634 total, no regressions**.

**Live verification:**
- ✅ `us.anthropic.claude-opus-4-7` round-trip in `us-east-1`
- ✅ `eu.anthropic.claude-opus-4-7` round-trip in `eu-west-1`
- ✅ Full `raptor.py analyze` pipeline e2e against a 2-finding Semgrep SARIF: 12 LLM calls (Stages A→D validation + exploit + patch generation), ~3 min, $1.83, all using `us.anthropic.claude-opus-4-7` — orchestrated report, exploit PoCs, and patches generated successfully.

## Backwards compatibility

- No changes to existing provider behaviour.
- `_DEFAULT_PROVIDER_ORDER` unchanged — Bedrock is opt-in only.
- `AnthropicProvider` only changes are the `supports_temperature` gates in `generate()` / `generate_structured()`. For all models currently in `MODEL_COSTS` except Opus 4.7, the gate evaluates True and the existing behaviour is preserved bit-for-bit.
- Bedrock model IDs alias to existing Anthropic cost/limit entries via a registration helper that's idempotent and runs at module import — no impact on cold-start time.

## License

- `boto3` (Apache 2.0) and `anthropic` (MIT) — both compatible with RAPTOR's MIT.
- Optional deps only — installed with `pip install boto3 anthropic`. No change to base requirements.

## Open question for the maintainers

Happy to split this into two PRs if preferred:
1. The `supports_temperature` registry + AnthropicProvider fix (Opus 4.7 is broken without it on the Anthropic-direct path too).
2. The Bedrock provider proper.

The bugs in (1) only surfaced because of (2), but they're independently useful. Let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,46 @@ export MISTRAL_API_KEY=...             # Mistral
 export OLLAMA_HOST=http://localhost:11434  # Local Ollama
 ```
 
+### AWS Bedrock
+
+Run Claude on Bedrock instead of the Anthropic API. Same models, same wire format, AWS-native auth via SigV4 — no API keys flow through models.json.
+
+Install the SDKs:
+
+```bash
+pip install boto3 anthropic
+```
+
+Set your region (boto3 picks up credentials from the standard chain — env vars, profile, IAM role):
+
+```bash
+export AWS_REGION=us-east-1
+export AWS_PROFILE=my-profile          # optional
+```
+
+Configure a model in `~/.config/raptor/models.json`:
+
+```json
+{
+  "models": [
+    {
+      "provider": "bedrock",
+      "model": "us.anthropic.claude-opus-4-7",
+      "role": "analysis"
+    },
+    {
+      "provider": "bedrock",
+      "model": "us.anthropic.claude-haiku-4-5-20251001-v1:0",
+      "role": "fallback"
+    }
+  ]
+}
+```
+
+Bedrock is opt-in only — RAPTOR will not auto-detect Bedrock from ambient `AWS_REGION` unless you explicitly select it via `models.json` or `RAPTOR_PROVIDER_PREFER=bedrock`. This avoids surprising users with AWS env vars set for unrelated reasons.
+
+Cross-region inference profiles (`us.`, `eu.`, `apac.`, `global.` prefixes) work the same as region-pinned IDs. Pricing tracks the underlying Anthropic models and is shared with the Anthropic provider's cost tables.
+
 Model roles let you assign different models to different tasks:
 
 | Role | What it does |
--- a/core/config/__init__.py
+++ b/core/config/__init__.py
@@ -737,6 +737,15 @@ class RaptorConfig:
         "AWS_ACCESS_KEY_ID",
         "AWS_SECRET_ACCESS_KEY",
         "AWS_SESSION_TOKEN",
+        # Bedrock additionally needs region + profile resolution to work.
+        # Without these, boto3 falls back to bare defaults and the
+        # ``AnthropicBedrock`` client cannot pick a regional endpoint.
+        "AWS_REGION",
+        "AWS_DEFAULT_REGION",
+        "AWS_PROFILE",
+        "AWS_DEFAULT_PROFILE",
+        "AWS_CONFIG_FILE",
+        "AWS_SHARED_CREDENTIALS_FILE",
         "AZURE_OPENAI_API_KEY",
         "AZURE_OPENAI_ENDPOINT",
         "GOOGLE_APPLICATION_CREDENTIALS",  # GCP service account JSON path
--- a/core/llm/config.py
+++ b/core/llm/config.py
@@ -283,6 +283,42 @@ def _build_anthropic_config() -> Optional['ModelConfig']:
     )
 
 
+def _build_bedrock_config() -> Optional['ModelConfig']:
+    """AWS Bedrock provider config builder.
+
+    Triggered when boto3 + AnthropicBedrock SDKs are importable AND
+    AWS_REGION (or AWS_DEFAULT_REGION) is set. boto3's credential
+    chain (env vars, profile, IAM role) handles auth — RAPTOR never
+    sees the credentials directly.
+
+    The default model is a cross-region inference profile for broadest
+    regional availability. Override in models.json if you need a
+    region-specific or non-default model.
+    """
+    from .detection import BEDROCK_AVAILABLE
+    if not BEDROCK_AVAILABLE:
+        return None
+    region = os.getenv("AWS_REGION") or os.getenv("AWS_DEFAULT_REGION")
+    if not region:
+        return None
+    default_model = PROVIDER_DEFAULT_MODELS.get(
+        "bedrock", "us.anthropic.claude-opus-4-7"
+    )
+    limits = MODEL_LIMITS.get(default_model, {})
+    costs = MODEL_COSTS.get(default_model, {})
+    return ModelConfig(
+        provider="bedrock",
+        model_name=default_model,
+        # api_key intentionally None — Bedrock auth is via boto3
+        # credential chain, not key passthrough.
+        api_key=None,
+        max_tokens=limits.get("max_output", _DEFAULT_MAX_OUTPUT_FRONTIER),
+        max_context=limits.get("max_context", _DEFAULT_MAX_CONTEXT_FRONTIER),
+        temperature=0.7,
+        cost_per_1k_tokens=(costs.get("input", 0.003) + costs.get("output", 0.015)) / 2,
+    )
+
+
 def _build_openai_compat_config(provider_name: str) -> Optional['ModelConfig']:
     """Generic builder for OpenAI / Gemini / Mistral — same shape, different env var + endpoint."""
     env_var_map = {"openai": "OPENAI_API_KEY", "gemini": "GEMINI_API_KEY", "mistral": "MISTRAL_API_KEY"}
@@ -401,6 +437,7 @@ def _build_claudecode_config() -> Optional['ModelConfig']:
 
 _PROVIDER_BUILDERS = {
     "anthropic":  _build_anthropic_config,
+    "bedrock":    _build_bedrock_config,
     "openai":     lambda: _build_openai_compat_config("openai"),
     "gemini":     lambda: _build_openai_compat_config("gemini"),
     "mistral":    lambda: _build_openai_compat_config("mistral"),
--- a/core/llm/detection.py
+++ b/core/llm/detection.py
@@ -38,6 +38,19 @@ except ImportError as _e:
     logger.debug("openai SDK probe failed: %s", _e)
     OPENAI_SDK_AVAILABLE = False
 
+# Bedrock requires both boto3 (for the AWS SDK and credential chain)
+# and AnthropicBedrock from the anthropic SDK (for the Messages API
+# wire format that BedrockProvider proxies to). If either is missing,
+# bedrock is not usable — even when AWS_REGION is set — and the
+# detection layer flags the misconfiguration.
+try:
+    import boto3 as _boto3_module  # noqa: F401 — availability probe
+    from anthropic import AnthropicBedrock as _ab  # noqa: F401 — availability probe
+    BEDROCK_AVAILABLE = True
+except ImportError as _e:
+    logger.debug("Bedrock SDK probe failed: %s (need boto3 + anthropic)", _e)
+    BEDROCK_AVAILABLE = False
+
 try:
     import anthropic as _anthropic_module  # noqa: F401 — availability probe
     ANTHROPIC_SDK_AVAILABLE = True
@@ -480,6 +493,9 @@ def _config_has_keyed_models() -> bool:
         if provider == "anthropic":
             if not (ANTHROPIC_SDK_AVAILABLE or OPENAI_SDK_AVAILABLE):
                 continue
+        elif provider == "bedrock":
+            if not BEDROCK_AVAILABLE:
+                continue
         elif provider == "ollama":
             if not OPENAI_SDK_AVAILABLE:
                 continue
@@ -532,8 +548,16 @@ def detect_llm_availability() -> LLMAvailability:
     has_openai = bool(os.getenv("OPENAI_API_KEY")) and OPENAI_SDK_AVAILABLE
     has_gemini = bool(os.getenv("GEMINI_API_KEY")) and (GENAI_SDK_AVAILABLE or OPENAI_SDK_AVAILABLE)
     has_mistral = bool(os.getenv("MISTRAL_API_KEY")) and OPENAI_SDK_AVAILABLE
+    # Bedrock signal: any AWS_REGION (or AWS_DEFAULT_REGION) plus the
+    # required SDKs. The boto3 credential chain handles actual auth —
+    # IAM role, profile, or env-var creds all work without any
+    # RAPTOR-side key handling.
+    has_bedrock = (
+        BEDROCK_AVAILABLE
+        and bool(os.getenv("AWS_REGION") or os.getenv("AWS_DEFAULT_REGION"))
+    )
 
-    has_cloud_keys = has_anthropic or has_openai or has_gemini or has_mistral
+    has_cloud_keys = has_anthropic or has_openai or has_gemini or has_mistral or has_bedrock
 
     # Phase B credential-isolation: a worker spawned via the
     # dispatcher has ``RAPTOR_LLM_SOCKET`` set but no API keys in
@@ -594,6 +618,7 @@ def _warn_unusable_keys():
         "openai": ("openai", OPENAI_SDK_AVAILABLE),
         "gemini": ("google-genai or openai", GENAI_SDK_AVAILABLE or OPENAI_SDK_AVAILABLE),
         "mistral": ("openai", OPENAI_SDK_AVAILABLE),
+        "bedrock": ("boto3 + anthropic", BEDROCK_AVAILABLE),
     }
 
     for provider, env_var in PROVIDER_ENV_KEYS.items():
--- a/core/llm/model_data.py
+++ b/core/llm/model_data.py
@@ -194,13 +194,100 @@ MODEL_LIMITS = {
     "ministral-3b-latest":     {"max_context": 256000,  "max_output": 256000},
 }
 
-# Provider -> env var mapping for API key lookup
+# Provider -> env var mapping for API key lookup.
+# Bedrock uses AWS_REGION as its "key marker" — there is no API key,
+# but the boto3 credential chain needs a region to operate. The detection
+# layer treats any AWS_REGION (or AWS_DEFAULT_REGION) as the presence
+# signal and the boto3 credential chain handles the actual auth.
 PROVIDER_ENV_KEYS = {
     "anthropic": "ANTHROPIC_API_KEY",
     "openai": "OPENAI_API_KEY",
     "gemini": "GEMINI_API_KEY",
     "mistral": "MISTRAL_API_KEY",
+    "bedrock":  "AWS_REGION",
 }
+
+
+# ---------------------------------------------------------------------------
+# AWS Bedrock model IDs.
+# Bedrock hosts the same Anthropic Claude models as the Anthropic API
+# but uses AWS-prefixed model IDs and SigV4 auth via boto3. Costs and
+# context/output limits are identical to the underlying Anthropic models;
+# we add the Bedrock IDs as duplicate entries so cost tracking and limit
+# enforcement work with no special-casing in callers.
+#
+# Inference profile prefixes (us./eu./apac.) enable cross-region
+# inference for higher throughput. Region-specific (non-prefixed) IDs
+# are also valid in regions that support on-demand for that model.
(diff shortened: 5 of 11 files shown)


grokjc:
Thanks for this - it's a genuinely well-built contribution. 

Since you'll be running this on your fork, you should know about a few real bugs before you depend on it:

**1. (Important) `family_of()` returns `"unknown"` for every Bedrock model ID - silently weakens cross-family validation.**

You patched `provider_of` with a prefix loop (good), but `family_of`/`same_family` in `core/security/llm_family.py` were left untouched, and they only match `claude-` / `anthropic/` prefixes:

```python
from core.security.llm_family import family_of, same_family
family_of("us.anthropic.claude-opus-4-7")               # -> "unknown"  (want "anthropic")
same_family("us.anthropic.claude-opus-4-7", "claude-opus-4-7")  # -> False  (same model!)
```

That module exists to keep a producer model and its cross-family checker independent (Attacker-Moves-Second, arXiv 2510.09023). With this, a Bedrock-Claude producer and a direct-API-Claude checker - literally the same model - read as *different* families, so they could be paired as "independent." Your `provider_of` docstring actually claims the family stays `anthropic`; the code doesn't do that. And `test_family_unchanged_for_bedrock_claude` asserts `in {"anthropic", "unknown"}`, so it passes on the broken branch and hides it. Fix: strip the Bedrock prefixes in `family_of` too, and tighten the test to `== "anthropic"`.

**2. (Medium) Detection treats ambient `AWS_REGION` as Bedrock intent -works against your own opt-in design.**

- `has_bedrock` feeds `has_cloud_keys` → `external_llm`. An AWS user with boto3+anthropic installed and `AWS_REGION` set but no LLM keys gets `external_llm=True`, yet `_get_default_primary_model()` returns `None` (Bedrock isn't in the default order). Detection says "LLM available"; selection finds none.
- `_warn_unusable_keys` now fires for *any* AWS user whose `AWS_REGION` is set but who hasn't installed boto3 - and `AWS_REGION` is set for countless unrelated reasons (unlike `ANTHROPIC_API_KEY`, which signals intent). The hint is also malformed: `"boto3 + anthropic".split(' or ')[0]` yields `pip install boto3 + anthropic`, which isn't a valid command.

Fix: gate `has_bedrock` (and the warning) on the same opt-in signal selection uses, not bare `AWS_REGION`.

**3. (Medium) `au.` vs `apac.` mismatch across files.**

Your cost map and `MODELS_WITHOUT_TEMPERATURE` use `au.` (correct for Claude 4.x, per your own comment), but the README, the `BedrockProvider` docstring, and `provider_of` all say/allow `apac.`. So a user following the README picks `apac.anthropic.claude-opus-4-7` → it routes to Bedrock but has no cost/limit alias (cost tracking breaks) and isn't in the temperature-skip set → it'll send `temperature` and 400. Suggest you pick one (`au.` for 4.x) and make it consistent everywhere.

---

Now the bigger picture, and the reason I can't merge this upstream as-is - and it's architectural, not anything particularly wrong with your (excellent) code.

RAPTOR is mid-way through a **credential-isolation** effort. Analysis runs spawn worker subprocesses that handle untrusted content; those workers route LLM calls through a Unix-socket **dispatcher** that holds the credentials in the trusted parent and injects auth on the worker's behalf, so the worker's address space never contains a key. That's already wired for Anthropic/OpenAI/Gemini and 9 aggregators.

Bedrock breaks that model, and your PR (correctly, given the current state) works *around* it by having `BedrockProvider` bypass the dispatcher and authenticate via direct boto3. The problem: the dispatcher relays a static **bearer token**, but Bedrock auth is **SigV4** - a per-request signature over method/path/headers/body/timestamp. You can't relay that the way you relay a bearer token. Hence the boto3 dependency. But, boto3's credential chain resolves creds from env **and** `~/.aws` files, SSO cache, container creds, and IMDS (`169.254.169.254`) - none of which env-scrubbing can isolate. So a Bedrock worker is exempt from the isolation guarantee by construction, and on an EC2 instance role it can even harvest creds from IMDS regardless of env. Merging this would bake in a provider that's permanently carved out of the very thing we're trying to close.

So the path forward is to build the missing piece - a **SigV4 signing proxy in the dispatcher** - and re-introduce Bedrock through it. The good news is it's mostly precedented. The Anthropic→Bedrock wire delta is tiny and mechanical (add `anthropic_version` to the body, move `model` into the URL path `/model/<id>/invoke`, point host at `bedrock-runtime.<region>`, sign with botocore's `SigV4Auth` instead of `x-api-key`). The one genuinely new piece is **AWS credential refresh** in the parent (STS/SSO/IMDS rotation, vs. static bearer keys) - which we'll share with the deferred Vertex work.

The consequence worth flagging: once the dispatcher owns Bedrock egress, **most of this PR's worker-side surface goes away.** The worker just uses the standard Anthropic provider over the socket (it thinks it's talking to Anthropic; the dispatcher translates + signs on the way out). That dissolves `BedrockProvider`, the 31-entry model-ID alias map, and the `family_of`/`provider_of` prefixes - and the cross-family bug disappears with them, since the worker only ever sees canonical `claude-*` IDs.

What that means for your effort: the pieces that survive either architecture and are worth keeping polished are the **temperature gating** (`MODELS_WITHOUT_TEMPERATURE` / `supports_temperature`), the **model cost/limit metadata**, and the **docs**. The `BedrockProvider` class itself and the family/provider routing are likely to be superseded, so I wouldn't sink more time into those as-is.

To be clear: I'd love to land Bedrock, and your work materially de-risks it - you've proven the client path end-to-end and built the model coverage. We just need it to enter through the signing proxy (which was already on the TODO list) so it's isolated from day one rather than as an exception we have to remember to carve out. I'm going to take on that proxy/credential-refresh work; once it lands I'll wire Bedrock in on top of it and credit this PR. Happy to keep you in the loop on that, and very open to your input on the design - thanks again for pushing this forward and giving me a well-placed kick to bring this back to the short-term task list. I wasn't sure there was a consumer for Bedrock but now I know there are users out there that need/want it!

---

As you said you'd be happy to split the PRs, can I take you up on the offer to land the fixes outside of this one, and then we can keep this PR open to track progress on getting Bedrock support into RAPTOR.  I know I could merge this and then replace it later, but that doesn't seem fair to you either and I'd rather land this right the first time given the constraints (which are all my fault in the first place 😆 )

grokjc:
Update on the credential-isolation blocker from my earlier review: the missing piece — a SigV4 signing proxy in the dispatcher — is now built and validated, so Bedrock can go through the dispatcher instead of authenticating with boto3 in worker address space.

What landed (dispatcher side): the bedrock rule gains a prepare_request hook that rewrites a worker's stock-Anthropic /v1/messages request into Bedrock's /model/<id>/invoke shape and attaches auth one of two ways:
  - SigV4 — the rewritten request is signed in the trusted parent with botocore's SigV4Auth (static keys or the profile/SSO/IMDS chain).
  - Bedrock API key (AWS_BEARER_TOKEN_BEDROCK) — a static Authorization: Bearer header; no botocore, no signing; takes precedence over SigV4, matching the AWS SDKs.

  The parent's AWS secrets (keys and bearer token) are read-and-erased at CredentialStore construction like every other provider key, so they never reach a spawned worker. botocore is an optional, parent-only dependency needed only for the SigV4 mode. Non-streaming InvokeModel only. I validated both modes with a live round-trip through the dispatcher against real Bedrock (eu-west-1, claude-haiku-4-5) — real 200 completions in each — plus an offline check that the signature is byte-identical to a re-sign of the exact wire request and to boto3's own InvokeModel request.

  What this means for this PR. The signing proxy makes the in-worker BedrockProvider (the boto3 path) unnecessary — that was the isolation blocker. The pieces that carry over and are still wanted are exactly your worker-facing surface:
  - the model cost/limit metadata,
  - temperature gating (MODELS_WITHOUT_TEMPERATURE / supports_temperature),
  - the family_of() fix (strip Bedrock prefixes → "anthropic"),
  - the docs,
  - and an alias → Bedrock-model-id mapping — note current on-demand Claude models generally need a regional inference-profile id (eu./us. prefix), which I hit firsthand in testing.

  So the cleanest path is to rebase this on top of the dispatcher work: drop the boto3 provider, have the worker select Bedrock via config and use the stock Anthropic provider over the dispatcher, and keep the metadata/temperature/docs/family_of changes. The temperature + bug-fix subset is still good to land on its own in the meantime.

  Thanks again for pushing this — your contribution is what surfaced the gap, and it carries through into the result.

grokjc:
I had to bring forward the temperature stuff in #716 as it was blocking various things. I gave you co-author credit - appreciate you surfacing the bug (looks like Anthropic have decided 4.7 onwards isn't going to use that param)

grokjc:
Two findings to flag from poking at this before we lock the porting plan.

  **1. Unprefixed entries aren't callable for on-demand.** I ran a quick `invoke-model` sweep on the 7 unprefixed Claude entries from your model_data.py additions, since I wasn't sure whether they were actually callable or just catalogue artefacts from `list-foundation-models`.

  Result: **7/7 return HTTP 400 — not callable for on-demand invocation.** AWS's error message is explicit:

  > "Invocation of model ID `anthropic.claude-opus-4-7` with on-demand throughput isn't supported. Retry your request with the ID or ARN of an inference profile that contains this model."

  That message came back for 6 of them (opus-4-7, sonnet-4-6, haiku-4-5, opus-4-5, opus-4-6, sonnet-4-5). The 7th — `anthropic.claude-opus-4-1-20250805-v1:0` — returned a different error: *"The provided model identifier is invalid"*, suggesting Opus 4.1 wasn't in the tested region's catalogue.

  So `list-foundation-models` returns the unprefixed forms as catalogue entries, but AWS requires an inference profile (`us.` / `eu.` / `au.` / `global.` prefix) for actual `invoke-model` calls on Claude 4.x. If you saw something different during your testing — e.g., if the unprefixed forms still work in some specific region/account configuration — I'd like to know, because the implication is the 8 unprefixed entries should probably come out of the data map.

  **2. Bedrock pricing tiers differ from direct Anthropic, by prefix.** Checking the AWS Bedrock pricing page surfaced two on-demand tiers for Claude:

  - **Global Cross-region Inference (`global.anthropic.…`)** — matches direct Anthropic pricing exactly. Opus 4.7: $5.00/$25.00 per 1M input/output. Sonnet 4.6: $3.00/$15.00. Haiku 4.5: $1.00/$5.00.
  - **Geo / In-region Cross-region (`us.` / `eu.` / `au.anthropic.…`)** — uniform +10% surcharge across all Claude models. Opus 4.7: $5.50/$27.50. Sonnet 4.6: $3.30/$16.50. Haiku 4.5: $1.10/$5.50.

  The current alias-to-canonical map (each Bedrock ID points at the canonical Anthropic name and inherits its cost row) under-quotes the `us./eu./au.` variants by 10%. Not a rounding error — operators tracking spend on regional inference profiles would see RAPTOR's totals diverge from their AWS console.

  Cleanest fix is a tiny multiplier — global = 1.0, everything else = 1.10 — applied at cost lookup. One rule, captures all current Claude-on-Bedrock IDs, and survives future Claude versions automatically since the pattern is AWS-structural rather than per-model.

  The 25 prefixed entries themselves look solid — your live round-trip tests cover the prefixed forms directly.

owen10380:
Thanks — this all makes sense, and the signing proxy is clearly the right home for
Bedrock. The plan sounds right to me: retire the in-worker boto3 `BedrockProvider` and
bring Bedrock in through the dispatcher, keeping the worker-facing pieces that carry over
(cost/limit metadata, the `family_of()` fix, docs, alias → Bedrock-id map). Good that the
work surfaced the gap and fed into something cleaner. Thanks for the co-author credit on
#716 too — useful to know 4.7+ dropped the temperature param.

**Q1 — the unprefixed `anthropic.claude-*` entries:** you're right, they should come out.
I don't have any evidence they worked on-demand — that was my mistake. I pulled them from
`list-foundation-models` and assumed callability rather than verifying with an actual
`invoke-model`. Your 7/7 HTTP 400 matches AWS's documented behaviour (Claude 4.x
on-demand needs an inference-profile prefix), so only the prefixed
`us.`/`eu.`/`au.`/`global.` forms should stay.

**Q2 — the +10% regional surcharge:** your call on implementation, but for what it's worth
I checked it against AWS's own docs and it aligns with my understanding:

- The [cross-Region inference guide](https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html)
  says global profiles give "approximately 10% cost savings" vs geographic profiles.
- The [global cross-Region inference guide](https://docs.aws.amazon.com/bedrock/latest/userguide/global-cross-region-inference.html)
  repeats "approximately 10% cost savings compared to geographic cross-Region inference."
- The [pricing page](https://aws.amazon.com/bedrock/pricing/) splits Claude into exactly
  two on-demand tiers — "Global Cross-region Inference" and "Geo Cross-region inference
  and in-region" — which matches your global-vs-(us/eu/au) split.

Two things I noticed while looking, in case they're useful:
1. AWS phrases it as "*approximately* 10%", so 1.10 reads as a documented estimate rather
   than an exact figure.
2. Global CRIS seems to have rolled out per-model (the docs currently scope the global
   discount to specific Claude models), so "global = 1.0, else = 1.10" holds wherever a
   `global.` profile exists — for a geo/in-region-only model there's no cheaper global
   baseline, so its geo price is effectively the base.

Makes sense to keep this open as the tracking PR. Happy to be a sounding board on the
design if it's useful.

grokjc:
We landed full AWS Bedrock support based on the credential-isolation proxy concept from your PR. The final shape ended up as two coexisting paths:

  - Mantle (default) — bedrock-mantle.<region>.api.aws/anthropic/v1/messages. Native Anthropic Messages API with bare model IDs, full feature parity (SSE streaming, tool use, prompt caching, vision, extended thinking, computer use).
  - Runtime — your original bedrock-runtime InvokeModel transform path, restored as the coexistence option for models not yet on Mantle and for cross-region inference profile IDs (us./eu./global.).

Operators pick per-run via RAPTOR_BEDROCK_API=mantle|runtime or per-model via a bedrock_api field in models.json. Bearer + SigV4 both work on both APIs.

The InvokeModel transform + signing path is essentially your code. Co-authored you on the landing commit (feat(llm/bedrock): full provider support — Mantle + InvokeModel coexistence) — your push to get Bedrock working is what got this rolling, even though the final design ended up routing through Mantle as the default. A follow-up commit added security guards (path-traversal, allow-list) and an expiry-aware credential UX (JWT exp pre-flight, refresh-friendly chain lookup, launcher warnings).

This all works fine E2E for me, but would appreciate some additional testing.

danielcuthbert:
Running my own tests now against latest main, 

```
git fetch origin main
git fetch origin pull/696/head:refs/remotes/origin/pr-696

python3 -m py_compile core/llm/providers.py core/llm/config.py core/llm/detection.py core/llm/model_data.py core/security/llm_family.py packages/llm_analysis/orchestrator.py

python3 -m pytest core/llm/tests/test_bedrock_provider.py
# 51 passed, 4 skipped

python3 -m pytest core/llm/tests
# 558 passed, 23 skipped, 12 deselected
```

we have some conflicts in:

```
core/llm/detection.py
core/llm/providers.py
core/security/llm_family.py
requirements.txt
```

main already contains a newer Bedrock path, including dispatcher-based credential handling and Bedrock signing hardening from #791 so this needs to be looked into. 
PR #696 uses direct AnthropicBedrock inside the provider/worker process. That weakens the current main design, where AWS credentials stay at the parent/dispatcher boundary so yeah need to fix that too. 
also worried about AWS_REGION / AWS_DEFAULT_REGION as enough to auto-build Bedrock config. That is risky because lots of users have AWS region variables set for unrelated work. main is more explicit and avoids promoting Bedrock just because region env vars exist.

grokjc:
To clarify I think we are 'done' re Bedrock, but would like some feedback re the implementation. I'm going to close this PR to avoid any further confusion but please do test further if you're able.

PR #697, opened 2026-05-28
feat(reachability): close per-language entry/dispatch gaps (PHP/Python/Go/Rust)

grokjc:
FN-safe add-entries/hedge fixes from a per-language audit:
- PHP: Laravel/Symfony framework-entry model (bases + #[Route]/#[AsCommand]
  + *Controller convention) — handlers no longer read not_called
- Python: capture class bases + class-based-view convention (Django/DRF/Flask)
- Go: structural-interface CHA-analog — an interface-dispatched method reads uncertain not no_path_from_entry (seed override_methods from receiver type + intercept the entry stage; also tightens Java/TS)
- Rust: flag Any::downcast/transmute as reflection masking (parity)

All move verdicts only toward reachable/uncertain, never suppress. _CACHE_VERSION bump (Go override_methods). 1076 tests pass.


grokjc:
Fixing the non-treesitter path ...

PR #765, opened 2026-06-03
fix(llm): use max_completion_tokens for OpenAI reasoning models (gpt-5.x, o-series)

dinosn:
## Problem

`OpenAICompatibleProvider` hardcoded the legacy `max_tokens` (and a non-default `temperature`) on every `chat.completions` call. The OpenAI **reasoning-tier** families — `gpt-5.x` and `o1`/`o3`/`o4` — rejected both:

```
HTTP 400 — Unsupported parameter: 'max_tokens' is not supported with this model.
Use 'max_completion_tokens' instead.
```

(reasoning models also only accept the default `temperature`). The effect: **every `gpt-5.x` call failed**, silently knocking the strongest OpenAI models out of `/understand --hunt` and `/agentic` multi-model runs. In practice this forced a fallback to weaker `gpt-4` models, which a multi-model audit specifically does not want.

## Fix

Add two small helpers in `core/llm/providers.py`:

- `_is_openai_reasoning_model(name)` — prefix match on `gpt-5`, `o1`, `o3`, `o4` (bare model id, so `openai/gpt-5.5` and date suffixes are tolerated).
- `_openai_sampling_kwargs(name, max_tokens, temperature=None)` — returns `{"max_completion_tokens": …}` (no temperature) for reasoning models, else the legacy `{"max_tokens": …, "temperature": …}`.

Applied at **all three** OpenAI call sites: `generate()`, the Instructor structured-output path, and `turn()` (the multi-model hunt/agentic codepath).

Non-OpenAI compat models (Ollama `qwen3`, `claude-*` via OpenAI-compat) and classic OpenAI models (`gpt-4o`, `gpt-4.1`) are **unchanged** — they keep `max_tokens` + `temperature`.

## Verification

- New unit test `core/llm/tests/test_openai_reasoning_tokens.py` pins the classifier + kwargs builder.
- Live: `gpt-5.5` and `gpt-5.4` now return **200** through both `generate()` and `turn()`; `gpt-4o` / `gpt-4.1` / `claude-opus-4-8` / `qwen3` behaviour unchanged.

### DIFF
--- a/core/llm/providers.py
+++ b/core/llm/providers.py
@@ -873,6 +873,40 @@ def _dict_schema_to_pydantic(schema: Union[Dict[str, Any], Type['BaseModel']]):
     return model
 
 
+def _is_openai_reasoning_model(model_name: str) -> bool:
+    """True for OpenAI reasoning-tier models (gpt-5.x, o1/o3/o4 families).
+
+    These models changed the chat.completions contract: they reject the
+    legacy ``max_tokens`` param (require ``max_completion_tokens``) and only
+    accept the default ``temperature`` (1) — passing ``temperature=0.7``
+    returns HTTP 400. Matched by bare model-name prefix so aggregator/
+    provider prefixes (``openai/gpt-5.5``) and date suffixes are tolerated.
+    Non-OpenAI compat models (Ollama ``qwen3``, ``claude-*`` via compat) do
+    not match and keep the legacy params.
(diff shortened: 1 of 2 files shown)


grokjc:
Thanks, appreciated. I will follow up with a small tweak to future proof this, which I'd appreciate you testing.

PR #767, opened 2026-06-03
fix(llm): version-gate OpenAI reasoning detection so future gpt-6/o5 don't 400

dinosn:
Hi, 
   This should be better handled now.

## Problem

`_is_openai_reasoning_model()` matched a **literal prefix list** (`gpt-5`/`o1`/`o3`/`o4`):

```python
return m.startswith(("gpt-5", "o1", "o3", "o4"))
```

So `gpt-6`, `gpt-10`, `o5`, `o6` would silently fall back to the classic `max_tokens` + `temperature` contract and return **HTTP 400 on every call** the day OpenAI shipped them — re-introducing the exact regression #765 fixed, just deferred a model generation.

## Fix

Gate on the version number instead, mirroring `supports_temperature`:

```python
_OPENAI_REASONING_GPT_FROM = 5
_GPT_VERSION_RE = re.compile(r"^gpt-(\d+)")
_O_SERIES_RE = re.compile(r"^o\d")
```

- **Whole o-series is reasoning** — the `o\d` anchor requires a digit right after `o`, so `o1`…`o6` match but `olmo`/`orca` don't.
- **`gpt` is reasoning from major version >= 5** — `gpt-5.x`, `gpt-6`, `gpt-10`. `gpt-4o` / `gpt-4.1` / `gpt-4-turbo` (major 4) stay classic.
- Bare-name match (`rsplit("/")`) still tolerates provider prefixes (`openai/gpt-6`) and date suffixes (`gpt-6-2026-01-01`).

## Tests

- Fixed the test's path bootstrap to use `Path(__file__).resolve().parents[3]` (the dir convention) instead of an `os.getcwd()` fallback.
- Extended the unit tests with future `gpt-6`/`gpt-10` and `o5`/`o6` cases plus `olmo`/`orca` negatives.

Validated: all five test functions pass; direct boundary sweep of 23 model names has no mismatches; `None` → `False`; `gpt-6` → `{'max_completion_tokens': ...}`, `gpt-4o` → `{'max_tokens': ..., 'temperature': ...}` (classic unchanged).

### DIFF
--- a/core/llm/providers.py
+++ b/core/llm/providers.py
@@ -873,19 +873,41 @@ def _dict_schema_to_pydantic(schema: Union[Dict[str, Any], Type['BaseModel']]):
     return model
 
 
+_OPENAI_REASONING_GPT_FROM = 5
+_GPT_VERSION_RE = re.compile(r"^gpt-(\d+)")
+_O_SERIES_RE = re.compile(r"^o\d")
+
+
 def _is_openai_reasoning_model(model_name: str) -> bool:
-    """True for OpenAI reasoning-tier models (gpt-5.x, o1/o3/o4 families).
+    """True for OpenAI reasoning-tier models (the o-series + gpt major >= 5).
 
     These models changed the chat.completions contract: they reject the
     legacy ``max_tokens`` param (require ``max_completion_tokens``) and only
     accept the default ``temperature`` (1) — passing ``temperature=0.7``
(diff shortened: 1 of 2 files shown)


grokjc:
I already fixed in #766 😄 But your regex is better, so I'm going to patch that and give you co-author credit on that change.

dinosn:
Thank you ! :) I rushed on the changes as I had noticed as well the case.  Appreciated!

PR #768, opened 2026-06-03
Improve SCA reachability reporting

danielcuthbert:
This makes the SCA report much easier to triage.

Before this, reachability was technically present in each individual finding, but you had to dig through the report to work out what was actually reachable. That is not great when the headline says something like '21 vulnerable dependencies' and the useful question is really 'which of these can I probably hit in this codebase?'.

What changed:

- Adds a reachability breakdown to the SCA report summary.
- Groups vulnerable dependencies into 'Reachable / likely used', 'Present, needs review', and 'Probably not reachable'.
- Adds render-time filters for old findings files: --only-reachable, --hide-not-reachable, and --reachability likely_called,imported.
- Keeps non-vulnerability SCA rows visible when filtering, including hygiene, supply-chain, and licence findings.
- Documents the new report behaviour and render flags in docs/sca.md.

Example:

Instead of just seeing:

21 vulnerable dependency findings

The report can now show:

Reachability breakdown
- Likely called: 3
- Imported: 5
- Not reachable: 9
- Not evaluated: 4

That gives the operator a much better starting point: fix the reachable bits first, review the uncertain ones, and avoid wasting time on dependency noise that is present but probably not in play.

Validation:

- python3 -m pytest packages/sca/tests/test_report.py packages/sca/tests/test_render.py packages/sca/tests/test_thresholds.py packages/sca/tests/test_sarif.py packages/sca/tests/test_output_schemas.py
- 106 tests passed

### DIFF
--- a/docs/sca.md
+++ b/docs/sca.md
@@ -120,6 +120,54 @@ Reports include a reachability breakdown for vulnerable dependencies and group
 the detailed vulnerability section into `Reachable / likely used`, `Present,
 needs review`, and `Probably not reachable`.
 
+## Reachability analysis
+
+SCA reachability is mechanical. RAPTOR does not ask an LLM whether a dependency
+is reachable; it derives the verdict from source evidence and advisory metadata.
+LLM review can still run later in the pipeline, unless `--no-llm` is set, but it
+is not the source of truth for the reachability verdict.
+
+Best results come from scanning the full source tree. SBOM-only scans can still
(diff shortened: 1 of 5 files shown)


danielcuthbert:
also amended the docs a bit as we ain't giving ourselves enough credit here for the different approaches.

grokjc:
Yes! LGTM

PR #776, opened 2026-06-05
feat: add proper threat modelling to agentic

danielcuthbert:
Build a project-owned threat model artefact for /agentic, keep operator edits unless --threat-model-refresh is asked for, and turn unchecked /understand flows into SARIF handoffs so the pipeline still has something useful when scanners miss the good stuff.

Also tightens the rough edges: stale context maps are refused unless explicitly allowed, bad agentic flags now fail before lifecycle setup, threat model prompt context is wrapped safely, and the new docs/tests cover the flow without making the whole thing feel like a policy document.

Good addition to /understand too i feel


grokjc:
`core/sandbox/tests/test_audit_cli_flags.py::test_only_four_profiles` fails on this branch:

      Extra items in the left set:
      'strict'

      Full diff:
        {
            'debug',
            'full',
            'network-only',
            'none',
      +     'strict',
        }

  The test hard-codes the expected PROFILES set:

  ```python
  class TestProfileSetIsTrimmedToFour:
      """Confirms `audit` and `audit-verbose` are GONE from PROFILES."""

      def test_only_four_profiles(self):
          from core.sandbox.profiles import PROFILES
          assert set(PROFILES) == {"full", "debug", "network-only", "none"}
```
  It was pinning the post-#269 "trimmed to four" contract — adding strict deliberately extends that contract but the test wasn't updated.

  Minimal fix:
```
  class TestProfileSet:
      """Pins the canonical profile names."""

      def test_canonical_profiles(self):
          from core.sandbox.profiles import PROFILES
          assert set(PROFILES) == {
              "full", "strict", "debug", "network-only", "none",
          }
```
  (Renaming the class drops the now-misleading "TrimmedToFour" framing; the test_no_audit_mode_field_in_profile_dicts test below it still pins the original "audit/audit-verbose are GONE" guarantee, so the post-#269 contract isn't lost.)

danielcuthbert:
Thing im adding today because it is saturday and raining:

1: Add native threat records
Each threat should have an ID, status, affected assets, affected flows, risk score, likely bug class, validation plan and linked evidence.

2: Add /threat-model diff:
Compare the saved model against the latest /understand output and show drift because this is important

3: Add /threat-model report:
right now it isn't good at all, so we should generate a proper Markdown report with diagrams, top threats, control gaps, confirmed findings and residual risk.

grokjc:
Going to bolt a few more features on, I like this one a lot 😄

PR #777, opened 2026-06-05
fix(sandbox): env-overridable default profile for rootless podman/distrobox

ferr079:
## Problem

On rootless podman / distrobox, the sandbox's kernel-isolation layers (mount-ns spawn, Landlock) cannot engage. With the default `full` profile, the sandbox half-engages and falls back to a broken path: `semgrep`/`codeql` produce **empty output**, and `raptor scan` silently reports **0 findings in 0 files** instead of failing loudly. The result looks like a clean scan when nothing actually ran.

There's no way to set a coherent default profile for such an environment short of passing `--sandbox none` on every single invocation.

## Fix

Make the default sandbox profile overridable via a `RAPTOR_SANDBOX` env var, so environments that can't run the kernel layers can opt into a coherent profile once:

- **`core/sandbox/profiles.py`** — `DEFAULT_PROFILE` is now resolved by `_resolve_default_profile()` from `RAPTOR_SANDBOX`. Unknown/empty values fall back to `"full"`, so a typo can never silently downgrade isolation.
- **`core/config/__init__.py`** — allowlist `RAPTOR_SANDBOX` in `SAFE_ENV_ALLOWLIST` so the operator's choice survives into the `scanner` `__main__` subprocess (otherwise the child re-resolves `DEFAULT_PROFILE` to `"full"` and the override is silently lost). Read only as a `PROFILES` enum key → no injection surface, identical in effect to passing `--sandbox`.
- **`core/sandbox/context.py`** — when no caller `profile=` and no CLI `--sandbox` flag, fall back to `DEFAULT_PROFILE` across **all** isolation layers, not just the seccomp default seeded at function top. Scoped to `DEFAULT_PROFILE != "full"` so the **stock hardened default path is byte-for-byte unchanged** — only an explicit operator downgrade takes the new branch.
- **`bin/raptor`** — inside a distrobox named `raptor` (guarded by `CONTAINER_ID`), default `RAPTOR_SANDBOX=none`, since the container is itself the isolation boundary. An explicit `RAPTOR_SANDBOX` / `--sandbox` still wins.

## Safety / scope

- Default behaviour with no env var set is unchanged: `DEFAULT_PROFILE == "full"`.
- Does **not** touch the deliberate `PYTHONUSERBASE` strip in `DANGEROUS_ENV_VARS`. A HOME-independent `semgrep` install (system / venv / pipx) remains required — a `pip --user` install breaks under the sandbox's `HOME` redirect regardless of this change.

## Testing

- `core/sandbox/tests/test_audit_cli_flags.py`, `test_context_backend_dispatch.py`, and `core/config/tests/test_config.py` all pass (35 + 39).
- Full run: **920 passed**, 25 skipped. The 55 failures are confined to `test_spawn_mount_ns.py`, `test_spawn_audit.py`, and `test_fingerprint_e2e.py` — they require mount-ns/Landlock/ptrace and fail identically on unmodified `main` in a rootless container (verified by checking out `main` and re-running). They are environmental, not introduced by this change. `test_tracer.py` fails to collect under Python 3.11 (3.12+ nested-quote f-string) — also pre-existing and unrelated.
- End-to-end: with `RAPTOR_SANDBOX=none`, `raptor scan` returns real findings again (was 0).

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/bin/raptor
+++ b/bin/raptor
@@ -170,6 +170,19 @@ export RAPTOR_DIR
 # and resolved RAPTOR_DIR; mark this invocation as trusted.
 export _RAPTOR_TRUSTED=1
 
+# Rootless podman / distrobox cannot engage mount-ns + Landlock, so the
+# default 'full' sandbox half-engages and the semgrep/codeql scanners emit
+# empty output. Inside the dedicated `raptor` distrobox, default the sandbox
+# profile to 'none' (the container itself is the isolation boundary). Only a
+# default — an explicit `export RAPTOR_SANDBOX=...` or `--sandbox <profile>`
+# still wins. Scoped via CONTAINER_ID so host / other distroboxes are
+# unaffected. RAPTOR_SANDBOX is allowlisted in core.config so it survives
+# into the scanner subprocess.
+if [ "${CONTAINER_ID:-}" = "raptor" ]; then
+    : "${RAPTOR_SANDBOX:=none}"
+    export RAPTOR_SANDBOX
+fi
+
 # Parse RAPTOR-specific args, collect Claude args
 STARTUP_ARGS=("--caller-dir" "$RAPTOR_CALLER_DIR")
 CLAUDE_ARGS=("-n" "RAPTOR")
--- a/core/config/__init__.py
+++ b/core/config/__init__.py
@@ -403,6 +403,20 @@ class RaptorConfig:
         #                    read as an enum (any other value → auto); no
         #                    injection surface.
         "RAPTOR_TARGET_KIND",
+        #   RAPTOR_SANDBOX   operator's default sandbox-profile override
+        #                    (full|debug|network-only|none), read by
+        #                    core/sandbox/profiles._resolve_default_profile().
+        #                    Must survive the subprocess boundary: the
+        #                    scanner runs as a __main__ subprocess with a
+        #                    get_safe_env() environment, and without this
+        #                    in the allowlist the child re-resolves
+        #                    DEFAULT_PROFILE to "full" — silently ignoring
+        #                    the operator's choice (e.g. "none" on rootless
+        #                    podman/distrobox where mount-ns + Landlock
+        #                    cannot engage). Read only as a PROFILES enum
+        #                    key (any other value → "full"); no injection
+        #                    surface, identical to passing --sandbox.
+        "RAPTOR_SANDBOX",
     })
 
     # Name prefixes — any variable whose name starts with one of these is
--- a/core/sandbox/context.py
+++ b/core/sandbox/context.py
@@ -683,6 +683,19 @@ def sandbox(block_network: bool = False, target: str = None, output: str = None,
         effectively_disabled = disabled or state._cli_sandbox_disabled
         if effectively_disabled:
             profile = "none"
+        elif profile is None and DEFAULT_PROFILE != "full":
+            # No explicit caller profile and no CLI --sandbox flag: when the
+            # operator has set a non-default DEFAULT_PROFILE (via
+            # RAPTOR_SANDBOX), honour it across ALL isolation layers, not
+            # just the seccomp default seeded at function top. Without this,
+            # `RAPTOR_SANDBOX=none` disabled seccomp but left Landlock +
+            # network-block engaged — which breaks the semgrep scanner on
+            # rootless podman / distrobox, where mount-ns + Landlock cannot
+            # operate and the half-engaged sandbox yields empty output.
+            # Scoped to `!= "full"` so the stock default path is byte-for-
+            # byte unchanged: only an explicit operator downgrade takes it.
+            profile = DEFAULT_PROFILE
+            effectively_disabled = (profile == "none")
     if profile is not None:
         if profile not in PROFILES:
             raise ValueError(
--- a/core/sandbox/profiles.py
+++ b/core/sandbox/profiles.py
@@ -12,6 +12,7 @@ a plain dict would let an external caller mutate `PROFILES["full"]
 sandbox() invocation in the process.
 """
 
+import os
 import types
 
 # Profile definitions
@@ -46,7 +47,24 @@ PROFILES = types.MappingProxyType({
     "network-only": types.MappingProxyType({"block_network": True,  "use_landlock": False, "seccomp": ""}),
     "none":         types.MappingProxyType({"block_network": False, "use_landlock": False, "seccomp": ""}),
 })
-DEFAULT_PROFILE = "full"
+def _resolve_default_profile() -> str:
+    """Default sandbox profile, overridable via ``RAPTOR_SANDBOX``.
+
+    Hardened default stays ``full`` (network blocked + Landlock +
+    seccomp). Environments where the kernel-isolation layers cannot
+    engage — notably **rootless podman / distrobox**, where mount-ns
+    spawn and Landlock are unavailable and the sandbox would otherwise
+    fall back to a half-broken path that yields empty scanner output —
+    can export ``RAPTOR_SANDBOX=none`` (or any valid profile name) to
+    set a coherent default without passing ``--sandbox`` on every call.
+    Unknown / empty values fall back to ``full`` so a typo never
+    silently downgrades isolation.
+    """
+    env = os.environ.get("RAPTOR_SANDBOX", "").strip().lower()
+    return env if env in PROFILES else "full"
+
+
+DEFAULT_PROFILE = _resolve_default_profile()
 
 
 # Kwargs that configure isolation. Callers must not pass these to


grokjc:
Hi! Thanks for flagging this. There are several things to unpick here: 

a) there's clearly a bug that needs hunting down and fixing - the silent failure. In fact, I already found it.
b) The env var should NEVER pass into the sandbox, because the sandboxed process shouldn't see sandbox machinery or any simple marker to tell it that it is sandboxed. And future code that spawns child processes inside sandbox propagates/inherits that marker, which isn't the case with a CLI flag.
c) We can't downgrade the sandbox by default because the process is running in a container. That is not grounds for that decision - some containers are 100% compatible with the sandbox/privs
d) In the absence of landlock/seccomp, the network sandbox is even more important - so 'None' may not be the right answer for everyone.

You are correct that right now the answer is to pass the --sandbox none argument to every command, but that isn't a great workaround for your situation (and was never intended to be)

The good news is that I'm going to take all of this on board and craft a fix that solves your exact problem, without breaking our trust model or introducing dangerous defaults. But I can't accept this PR as-is due to the concerns above and I'd rather not delegate security-critical engineering tasks to you. Already working on it - stand by!

ferr079:
Thanks for the fast and detailed response — and for already being on it.

Your reasoning makes sense and I'd rather your fix than mine:

- **(b)** Agreed, this is the part I got wrong. An env marker is ambient and inherited by spawned children, so the sandboxed process (and anything it launches) could detect or act on it — that breaks the opacity the sandbox depends on. A per-invocation CLI flag stays outside that boundary; an env var doesn't. Hadn't thought through the child-process propagation.
- **(c)** Fair — "runs in a container" was the wrong signal. The actual failing condition here is *rootless podman where mount-ns/Landlock can't engage*, not containerization per se, and plenty of container setups keep full sandbox capability.
- **(d)** Good point, and the one I'd have missed: with Landlock/seccomp unavailable, the network sandbox is carrying more of the weight, so defaulting it off would quietly widen exposure exactly where it matters most.

So the real bug is the **silent** part — `0 findings in 0 files` reading as a clean scan when the sandbox half-engaged and the analyzers produced nothing. Failing loudly (or refusing to run) there is the right primitive; how to safely degrade is your call to make inside the trust model.

Happy to test your fix against the environment that surfaced this — rootless podman in a distrobox on Fedora Silverblue (Bluefin), where `semgrep`/`codeql` came back empty. Just ping me. Closing thoughts aside, feel free to close this PR whenever suits — the issue's in better hands now.

grokjc:
Thanks - have (hopefully) landed the fix for the underlying silent-failure bug. I'd be grateful if you can confirm that is the case on your setup - it still won't work, but you should get an error .

Following on from that I'm ready to land the opt-out logic, but this is a good opportunity to test the first fix in isolation.

grokjc:
Also @ferr079 I'd like to understand your use case. Are you using `bin/raptor`, running `claude`, or using e.g. ```python raptor.py``` to start your sessions in this context? I have a solution in mind but need to know which of those applies please 😄

ferr079:
Tested your fix on my setup — **the silent false-clean is gone**. Details below, plus the use-case answer.

## Use case

`python3 raptor.py <mode>` (scan / sca / doctor / agentic), always from inside the distrobox — either an interactive shell, or wrapped from the host as `distrobox enter raptor -- bash -c 'cd ~/raptor && python3 raptor.py scan ...'` for automation. I don't start sessions with `claude` directly, and `bin/raptor` only occasionally.

## Test results (main @ `0e770371`)

Host: Fedora Bluefin (SELinux), container: distrobox/rootless podman, Debian 12. Same repro as before, default sandbox profile, no flags:

- ✅ Engagement failure is now loud and precise:
  ```
  [INFO] Sandbox: mount-namespace isolation UNAVAILABLE — unshare(CLONE_NEWNS) refused at runtime (...). Fallback: Landlock-only.
  [WARNING] RAPTOR: sandbox running in Landlock-only mode — ...
  ```
- ✅ Per-pack failures surface in console **and** report:
  ```
  ⚠️  semgrep: 4 pack(s) failed or produced no SARIF: category_secrets, semgrep_owasp_top_10, semgrep_secrets, semgrep_security_audit
  ```
  plus `metrics.semgrep_failed_packs` in the JSON. No more "0 findings in 0 files" that looks like a clean audit.
- ✅ Control with `--sandbox none` still works (semgrep runs, 33 files scanned).

## Possibly useful for your opt-out design

The Landlock-only fallback actually gets semgrep **very close** to working here — the remaining per-pack failure is just:

```
[INFO] Sandbox: write outside allowed paths denied to ~/.config/.semgrep/semgrep.log during: /usr/local/bin/semgrep scan --config (rc=1)
```

So on this setup the blocker is semgrep writing its log/settings under `~/.config/.semgrep`. If that path (or an XDG redirect into the sandbox-writable area) were allowed, scans might work fully *with* Landlock engaged — no downgrade needed at all, which seems even better aligned with your trust model.

Two minor observations:

1. **Exit code is still 0** when all packs fail. Fine interactively, but for CI/scripted use a non-zero exit (or an opt-in `--strict` style flag) would prevent a wrapper from treating a no-op scan as a pass.
2. The hint `likely kernel.apparmor_restrict_unprivileged_userns=1` misattributes on Fedora-family hosts (SELinux; that sysctl doesn't exist). Here the `unshare(CLONE_NEWNS)` EPERM comes from rootless podman's container confinement itself (`unshare -m` → EPERM inside the container).

Happy to re-test whatever you land for the opt-out.

grokjc:
Thanks for the feedback, appreciate it. I'll get on with the next bit!

ferr079:
Closing this — the per-invocation flag is the right boundary and your trust-model reasoning (b/c/d) convinced me. The real win is already landed: the silent false-clean is fixed and now fails loudly on my setup. Happy to test whatever you land for the opt-out — ping me anytime. Thanks for the thoughtful review! 🙏

grokjc:
Have made some more changes - interested to know if they work for you!

ferr079:
Tested `d223b176` on the same setup — **the default-profile scan now completes cleanly, no crash.** 🎉

## Environment
- Host: Fedora Bluefin 44 (kernel `7.0.8-200.fc44`, **SELinux**), podman 5.8.2 rootless
- Container: distrobox / Debian 12 (bookworm), Python 3.11.2
- `python3 raptor.py scan --repo <dir> --out <dir>`, **default sandbox profile, no flags**

## Result — fixed ✅
Previously the operator `XDG_CONFIG_HOME` survived the allowlist, semgrep wrote outside the writable policy and Landlock denied it → crash. With `fake_home` the Landlock-only path now runs end to end:

```
[INFO] Sandbox: mount-namespace isolation UNAVAILABLE — unshare(CLONE_NEWNS) refused at runtime ... Fallback: Landlock-only.
[INFO] Sandbox (landlock+tcp+seccomp:full+limits): /usr/local/bin/semgrep scan --config
...
[INFO] Scan complete: 6 findings
"semgrep_failed_packs": []
```

Exit 0, 6 findings (4× `subprocess-shell-true`, 2× `eval-detected`), 8.9s, `doctor`: 0 failures. So the full arc — silent false-clean → loud failure → actually working — is complete on my end. Nicely done routing it through the sandbox layer rather than per-tool.

## Two small things I noticed (non-blocking)

1. **A warning fires once per semgrep pack (~12×):**
   ```
   Sandbox: caller supplied custom env= for /usr/local/bin/semgrep scan --config —
   get_safe_env() not applied; caller env passed through. Pass strict_env=True ...
   ```
   Looks like the scanner now passes `env=` explicitly (the `fake_home_env` merge?), so `get_safe_env()` is skipped. Intentional, or should the scanner pass `strict_env=True` here?

2. **The Landlock-only hint points at `kernel.apparmor_restrict_unprivileged_userns=1`** — on Fedora/SELinux that sysctl doesn't exist (`n/a` on my box); the `unshare(CLONE_NEWNS)` refusal comes from the SELinux + rootless-podman combo instead. Might be worth softening the message so it doesn't assume AppArmor, e.g. "likely `apparmor_restrict_unprivileged_userns=1` or an equivalent LSM/SELinux restriction".

Happy to test the `--extra-config` path too if that's useful. Thanks again for the thoughtful iteration! 🙏

grokjc:
Hopefully those two small things are also resolved now 😄 - and yes, please test the --extra-config path if you can.

ferr079:
Tested `bb6ddd6c` (#800) on the same setup — **both small things are gone, and `--extra-config` works end to end.** 🎯

**Env:** Fedora Bluefin 44 (kernel `7.0.8-200.fc44`, SELinux), podman 5.8.2 rootless, distrobox / Debian 12, Python 3.11.2. Default sandbox profile (Landlock-only fallback), no flags.

**1. `custom env=` warning — silenced ✅**
The ~12× per-pack `get_safe_env() not applied` warning is gone on both a plain `scan` and an `--extra-config` run. Routing it through `strict_env` was the right call.

**2. `--extra-config` peer scan — works ✅**
Passed a one-rule YAML (`pattern: eval(...)`). It ran as its own peer scan and merged into `combined.sarif` alongside the registry packs:
```
findings_by_rule:
  ...subprocess-shell-true: 4
  tmp.custom-no-eval: 1        # ← the --extra-config rule, matched at L2
semgrep_failed_packs: []
```
Exit 0, `doctor`: 0 failures. The custom SARIF is correctly namespaced and merged — no collision with the registry rule IDs.

So the full arc is closed on my end: silent false-clean → loud failure → working default scan → working custom rules, all inside the sandbox layer. Thanks for seeing this through! 🙏

PR #793, opened 2026-06-08
[Feat] - calibrated aggregation

unprovable:
feat(scorecard,multi_model): calibrated aggregation - Replaces vote-based multi-model consensus with a Dawid–Skene calibrated estimator: scorecard sufficiency audit (libexec/raptor-scorecard-audit) + pure-Python Beta priors with no scipy dep + panel-log loader reading existing orchestrated_report.json (no new sink needed) + binary EM estimator with log-space E-step and posterior-mean M-step + per-finding calibrated_aggregation field on `/agentic` output (additive, behind RAPTOR_CALIBRATED_AGGREGATION opt-out) + posterior-weighted scorecard updates via new MULTI_MODEL_CONSENSUS_CALIBRATED event slot, closing the structural circularity at consensus.py:131 where majority defined truth → graded dissenters → updated scorecard → scorecard was unused at the next vote. 

Also ships `libexec/raptor-panel-replay` as a read-only research harness that re-aggregates historical reports and reports flip rates + per-model inferred reliability. 131 new tests across 7 files; all 230 existing scorecard tests pass; design doc at `docs/design-aggregation-dominators-wp.md`. 

This should clear up some circularity in the pipeline:  majority defines truth → truth grades dissenters → grades populate scorecard → scorecard isn't used to weight the next majority vote. The scorecard is a passive observer of a system that ignores it.

Dawid–Skene breaks the circle: truth is latent, model reliabilities and the latent label are estimated jointly via EM, and the scorecard's per-class confusion data becomes the input rather than a downstream report. That changes what RAPTOR is; from "ensemble of opinions" to "calibrated estimator with a credible interval per finding." 

Triage ordering by posterior × severity falls out for free. This of course re-shapes a load-bearing pipeline, but with some improvement I hope!


grokjc:
As I am time-poor this week, here's what Claude thinks 😄  Will follow up with a proper review ASAP.

### Architectural

1. **One consensus mode — collapse, don't fork.** Drop the dual `MULTI_MODEL_CONSENSUS` / `MULTI_MODEL_CONSENSUS_CALIBRATED` event types. Replace the old path with the new one. Soft credits always; legacy = `correct=1.0, incorrect=0.0` is what `record_event_soft` already stores. Existing readers see the same schema. No fork, no audit of downstream consumers, no migration burden.

2. **No env var.** `RAPTOR_CALIBRATED_AGGREGATION` should not exist. If a runtime switch is needed, it's a CLI flag — but with the above (one mode), there's nothing to switch.

3. **Shadow mode for the cutover.** Phase 1a's audit returned no-data — D–S will be prior-dominated on fresh installs. Ship Phase 3 (emit calibrated field as telemetry) first; gate Phase 4's scorecard-update change behind measurement. Validate on real data via the replay harness, then flip in a follow-up PR.

4. **Priors from `/validate`, not the scorecard.** `/validate`'s `exploitable` / `disproven` is labeled ground truth — the right substrate for informed priors. Uniform Beta(1,1) stays as cold-start fallback. Document the path.

5. **Reposition the replay harness.** It's the validation mechanism for the gate-flip decision, not a research side project. Reframe in docs + design doc Phase 2d.

### Specific

- `libexec/raptor-scorecard-audit` + `libexec/raptor-panel-replay` — wrong placement. Both are operator-typed research/diagnostic tools, not slash-command dispatch. Move to `core/llm/scorecard/scripts/` and `core/llm/multi_model/scripts/` as standalone Python scripts (matching the `core/inventory/scripts/binary-oracle-closed-loop` pattern).
- `orchestrator.py:880-885` — silent `try/except Exception` + log-warning is the same pattern we previously rejected for sandbox. Either return a structured result type, or add a `scan_metrics["calibrated_aggregation_failures"]` counter so silent fallback is observable.
- `calibrated_aggregation.py:~80` — vote-fallback stats line should break down by reason: "5 fell back (3: single-model panel, 2: all-error)" not "5 fell back."
- `docs/design-aggregation-dominators-wp.md` Phase 1b — note that operators landing this with sparse cells get posteriors near 0.5 and near-zero soft updates. Not a bug — prior dominating — but document it.

unprovable:
So there's some deferred work here once there's some tangible data, but I think this is a reasonable start to better utilise the scorecards :)

grokjc:
Apart from the rebase/issue re slow tests (see #794) is the plan to wire in ```priors_from_validation()``` later? If so, all good, this is very close to merge o'clock 😄

grokjc:
OK, to summarise this is where I think we are:

  1. Rebase on current main to drop `95dc63e7` — slow-mark band-aids fixture amortization that already shipped in #732 / #799.
  2. `priors_from_validation` is API-only — orchestrator calls `calibrate_results(results_by_id)` with no `priors_by_class`. Confirm: wire-up deferred with Phase 4, or missed?
  3. Phase 4 design question: when scorecard updates flip to posterior-weighted, replace the vote path rather than coexist as an operator mode. Confirm before starting the follow-up.

  4. **Orchestrator mutation loop isn't atomic.** `for fid, verdict in verdicts.items(): primary["calibrated_aggregation"] = verdict_to_json(verdict)` is inside the `try`. If `verdict_to_json` raises on finding N, findings 0..N-1 carry the field, N+1.. don't. `calibrated_summary["failed"]=True` tells consumers to disregard, but individual findings are in mixed state. Either build all conversions into a dict first  then mutate (atomic), or document partial-state as the expected failure mode.

  5. **No test for the failure-path summary.** The `except` branch building `{"failed": True, "error": ...}` and landing it in `orchestrated_report.json` isn't covered. Worth one test that mocks `calibrate_results` to raise and verifies the summary shape.

grokjc (commented):
Comments already left on thread: this is very close to being ready to merge, and a good piece of work too!

unprovable:
Thanks for the detailed pass — went through the punch list. :)
  
  ### ✅ Done (commit `aff9e0ea`)
  
  **4 — Atomic mutation loop.** Confirmed the bug: the convert-and-assign ran per-finding inside the `try`, so a `verdict_to_json` failure on finding N left 0..N-1 carrying `calibrated_aggregation` and N+1.. without it. Fixed by serialising **all** verdicts up front, then assigning in a second loop that can't raise — now it's all-or-nothing, and any failure lands cleanly in the summary as `failed: true`. Extracted the block into `_attach_calibrated_aggregation(results_by_id)` to do this without duplicating logic.
  
  **5 — Failure-path coverage.** The extraction makes both paths unit-testable without booting `orchestrate()`. Added tests for: success-summary shape, the `{"failed": True, "error": ...}` path (mock `calibrate_results` to raise), and the atomicity guarantee (mock `verdict_to_json` to raise on the 2nd finding → no finding mutated). Existing wire-up tests now drive the real helper instead of an inline replica. `ruff-pr` clean.
  
  ### ✅ Confirmations
  
  **2 — `priors_from_validation` is API-only.** Intentional — wire-up is **deferred with Phase 4**. The design doc's Phase 4 section names it as the integration point (feed `priors_by_class` into `calibrate_results` from  `/validate` ground truth). Not missed.
  
  **3 — Phase 4 replaces the vote path.** Agreed, and that's the documented plan: collapse to **one** consensus mode recording soft credits via `record_event_soft` (legacy discrete update = the `{correct=1.0, incorrect=0.0}` special case), not a coexisting operator mode. Will not start the follow-up as a fork. 
  
I think that the slow marks should now be resolved... the CI being my ultimate litmus test.

unprovable:
I gues it's fine and the slow marks were overkill?

grokjc:
I think the only thing missing is a plan/commitment - when are we going to 'flip the switch' on this? Once we've nailed that down we can merge this as-is.

unprovable:
Ok here goes... 

## Phase 4 flip commitment.

So we're all agreed, this is the last open item. The flip is data-gated, not calendar-gated, because Phase 1a came back 'no-data'; flipping before there's real `/validate` ground truth just feeds the EM its own priors.

Where the thresholds sit today (shipped in this PR, all additive):
- Dawid–Skene runs only on findings with ≥2 valid panel members (`calibrated_aggregation.py`); fewer → `aggregation_method="vote"` fallback (`single_panel_member` / `insufficient_panel_size_N`). No verdict changes either way.
- The per-class prior is the uniform Beta(1,1) cold-start, and `priors_by_class` is empty, nothing feeds it yet. So the EM is prior-dominated and calibrated output tracks majority. That's intended for telemetry-only, and it's the concrete reason the scorecard flip would be near-inert today.

Calibration steps needed before the flip (the wiring Phase 4 adds):
1. Harvest /validate rulings into per-class (`n_exploitable, n_disproven`) counts.
2. Feed them through `priors.priors_from_validation(counts_by_class)` → `{decision_class: BetaPrior}` (the factory already exists and is unit-tested; this is the deferred call site).
3. Pass that map as `calibrate_results(..., priors_by_class=…)` so the EM is data-informed instead of Beta(1,1)-dominated; classes still without labels keep the uniform cold-start.
4. Run panel-replay over accumulated `orchestrated_report.json history` → confirm the posterior-weighted path moves credit in the right direction (mis-graded-dissenter case) with no regression on agreed findings.
5. Only then swap consensus.record_consensus_outcomes to posterior-weighted soft credits via record_event_soft (legacy discrete = the `{1.0, 0.0}` special case). One mode, no fork.

Gate (all three): ≥30 /validate labels/class on the active classes (so step 2 is data-driven, not cold-start). Replay clears with no regression, single-mode soft-credit shape.

Checkpoint: I'll re-run panel-replay and post flip-rate + per-class label counts here at the next monthly review (say we target 2026-07-18). If the gate clears the follow-up PR opens then; if not, the posted counts show how far off we are. Merging 793 as-is now; it's pure additive telemetry, nothing waits on the flip.

unprovable:
I may have an idea.

unprovable:
So, #794 was actually a second part to this project initially, so with this done, there's another small project that I'm doing now that basically helps with the phase 4 issue. Working on it now with the Tin Frenchman

unprovable:
WP extraction → /exploit is up in its own PR now - #814 
 - Phase 4 is the one carve-out, and it's intentional — not missing. Phase 1a returned no-data, so flipping the posterior-weighted scorecard update now would be prior-dominated noise. The doc now spells out the activation gate (≥30 /validate labels/class so priors are data-driven, replay-harness clears with no regression, one consensus mode via record_event_soft) plus the checkpoint to re-measure. 

Phases 1–3 in this PR are pure additive telemetry and change no verdict, so nothing here waits on the flip. This PR is good to merge as-is on that basis with the note above. I think this closes off a lot of the things I was trying to do to extend the efficacy of the solver infra.

grokjc:
Merged - thanks!

PR #794, opened 2026-06-08
feat(inventory,dataflow): sanitizer-cut suppressor

unprovable:
Adds a pure-graph substrate that decides "every dynamic path from source to sink crosses a sanitizer" without an LLM call. We add Lengauer–Tarjan *A Fast Algorithm for Finding Dominators in a Flowgraph*, ACM TOPLAS 1(1), 1979. to our CFG's to eliminate FPs, reduce LLM calls. 

Graph theory layer:
    - core/inventory/dominators.py Iterative Lengauer-Tarjan O(E·log V) over a generic Graph[N] Protocol. DomTree exposes idom / dominates / dominators_of. Pure Python, no scipy. Iterative DFS so deep CFGs don't trip cpython's 1000-frame recursion limit. AdjacencyGraph helper for callers with raw {node: [succs]} maps.
    - core/inventory/cfg_builder.py Two producers under one Protocol: build_python_cfg(source, function_name) — intra-procedural CFG via stdlib ast: if/elif/else, while, for (with break/continue), try/except/finally, with, return, raise. Statement-level granularity with per-node `calls` extraction. build_cpp_callgraph(binaries, entry) — inter-procedural call graph via existing core/inventory/binary_oracle_edges. Unions edges across multiple binaries (hybrid library + executable).

Sanitizer recognition:
    - core/dataflow/sanitizer_catalog.py Layered on top of the existing core/dataflow/known_safe_calls table — does NOT duplicate the 24 curated entries. Adds: sink_classes_for_cwe(cwe) — CWE-79/89/78/22/... mapping sanitizer_callables_for_cwe(cwe, language) — catalog query match_sanitizers_in_cfg(graph, cwe, language) — recognizer, duck-typed on node.calls (PyCFGNode) or node.name (CallGraphNode) so one function serves both producers. Tests assert data integrity in both directions: every catalog sink_class is reachable via some CWE, and every CWE maps to sink_classes the catalog actually has entries for.

FP reduction:
    - core/inventory/sanitizer_cut.py Pure vertex-cut: sink is unreachable from source in CFG \\ candidate_sanitizers. BFS-after-deletion, O(V + E). evaluate_finding() returns a SanitizerCutResult with the witness cut set; record_sanitizer_cut_suppression() bridges into the existing core/inventory/reach_chokepoint.record_suppression so the audit JSONL trail unifies with binary-oracle suppressions under verdict='sanitizer_dominated'.

Tests (82 new + 1 skipped optional NetworkX cross-check):
    - 16 dominators (Lengauer-Tarjan textbook Fig. 1 as golden, plus linear/diamond/loop/nested-loop/disconnected cases)
    - 21 CFG builder (Python control flow + C/C++ call graph end-to-end through dominators)
    - 24 sanitizer catalog (data integrity, CWE normalization, Python/Java/JS recognizer coverage)
    - 21 sanitizer-cut (canonical sibling-branch suppress + bypass, multi-source semantics, loop-body-only no-suppress, JSONL integration) Ruff clean across all 8 files.

Deferred ideas:
    - smt_barrier.py:746 / :940 upgrade: the existing lexical validator_dominates_sink and substitution_dominates_sink are production-path code; migrating them to the vertex-cut should ship behind a flag (like RAPTOR_CALIBRATED_AGGREGATION in Project A) as a separate, smaller change.
    - Live chokepoint wiring into the binary-oracle pipeline: the record_sanitizer_cut_suppression helper exists and writes the right JSONL shape, but no caller invokes it yet.

**NB** - `smt_barrier.py` already encodes the concept of this PR — it just does it with line numbers. Promoting it to Lengauer–Tarjan over `binary_oracle_edges.py`’s call edges (and intra-procedural CFGs) generalizes an existing mechanism rather than introducing a new one. Slots into suppressions.jsonl cleanly. Real FP reduction, no LLM cost. But it doesn't reshape the system — it extends an existing chokepoint.


danielcuthbert:
Nice bit of work this. The graph substrate is useful, and the sibling-branch sanitizer case is exactly the sort of false positive we should be killing without spending LLM tokens. I would just be careful that this does not become a suppressor before it has value binding.

The current cut proof is control-flow only: it proves every path crosses a node that contains a sanitizer call. It does not yet prove that the tainted value went into that sanitizer, or that the sanitizer return value is the value that reaches the sink. That distinction matters a lot. This currently suppresses a case like:

```python
def handle(user, other):
    safe_other = html.escape(other)
    render(user.name)
```

For CWE-79 the graph cut removes the `html.escape(other)` node and the sink becomes unreachable, but `user.name` was never sanitised. So I think the next nice step is to split this into candidate recognition and value-bound suppression.

Suggested shape:

```python
@dataclass(frozen=True)
class CallSite:
    name: str
    arg_names: frozenset[str]
    assigned_names: frozenset[str]
    lineno: int

@dataclass(frozen=True)
class PyCFGNode:
    kind: str
    lineno: int
    label: str
    calls: frozenset[str] = frozenset()
    call_sites: tuple[CallSite, ...] = ()
    defs: frozenset[str] = frozenset()
    uses: frozenset[str] = frozenset()
```

Then `match_sanitizers_in_cfg()` can return sanitizer bindings rather than raw nodes:

```python
@dataclass(frozen=True)
class SanitizerBinding:
    node: object
    callable: str
    input_symbols: frozenset[str]
    output_symbols: frozenset[str]
```

And `evaluate_finding()` should only return `suppress=True` when all of these are true:

1. the sanitizer callable matches the CWE/language catalogue;
2. the sanitizer input intersects the tainted source symbols or the source-to-sink path value;
3. the sanitizer output symbol reaches the sink argument;
4. removing those value-bound sanitizer nodes cuts every source-to-sink path.

If RAPTOR cannot prove those value facts, I would return something like `candidate_only` / `needs_value_binding` rather than suppressing. That keeps this safe while still giving us a really useful hint for the LLM or validation phase, which is kinda sexy time for me.

On plumbing: I would land the substrate first, then wire it behind a flag into the existing suppression chokepoint. The natural first integration point looks like the existing sanitizer logic in `core/dataflow/smt_barrier.py`: use the current lexical sanitizer-dominates-sink check as the fallback, and try the vertex-cut proof when we can resolve source/sink nodes inside an enclosing Python function CFG. Something like:

```python
if feature_flags.sanitizer_cut:
    proof = sanitizer_cut.evaluate_finding(
        cfg,
        source_nodes,
        sink_node,
        cwe=finding.cwe_id,
        language=finding.language,
        value_flow=finding.value_flow,
    )
    if proof.suppress:
        record_sanitizer_cut_suppression(out_dir, finding, proof)
        return Suppressed(verdict="sanitizer_dominated")
```

For `/validate` and `/agentic`, the caller already has SARIF-ish source/sink/CWE context in a few places, so the practical flow could be:

1. parse finding to get file, CWE, language, source line, sink line;
2. build `build_python_cfg(file, enclosing_function)`;
3. resolve source/sink nodes by line or SARIF code-flow location;
4. run the value-bound sanitizer-cut proof;
5. write `suppressions.jsonl` via the existing chokepoint;
6. surface the outcome in the final report as `Suppressed: Sanitizer Dominated`, including the sanitizer call and line numbers.

For C/C++ call graphs, I would be more conservative. Function-granularity call graphs can say "every route to this sink function calls a known sanitizer function", but they cannot prove argument/value binding. I would treat C/C++ as `candidate_only` until there is an intra-procedural value layer or binary/source-level data-flow witness.

danielcuthbert:
Adding a rough wiring sketch as well, because this is the bit that makes the feature properly useful and sexy

The shape I would go for is: Semgrep / CodeQL still find the noisy thing tho i love CodeQL more than anything else, RAPTOR builds the local CFG, then the sanitizer cut only gets to bin the finding if it can prove the same dirty value went through the sanitizer and the cleaned value is what hits the sink. Belt and braces, basically.

badda bing badda boom

```mermaid
flowchart TD
    A["Semgrep / CodeQL finding"] --> B["Normalise finding
file + CWE + language + source + sink"]
    B --> C["Find enclosing function"]
    C --> D["Build CFG
Python AST now, C/C++ later"]
    D --> E["Match known sanitizer calls
from known_safe_calls"]
    E --> F{"Value binding proved?"}
    F -- "Nope" --> G["Keep finding
needs validation / LLM"]
    F -- "Yes" --> H["Remove bound sanitizer nodes"]
    H --> I{"Can source still reach sink?"}
    I -- "Yes" --> J["Keep finding
bypass path exists"]
    I -- "No" --> K["Suppress as sanitizer_dominated"]
    K --> L["Write suppressions.jsonl
with sanitizer witness"]
    L --> M["/validate + /agentic report
shows why it was dropped"]
```

The important faff is the middle diamond. Control-flow proof alone says: every path crosses a line with `html.escape`. That is not quite enough. We need: the tainted input went into `html.escape`, and the return value from that exact call is what reaches `render`, `execute`, `open`, SQL concat, or whatever sink we are looking at.

So for the first real plumbing pass I would do this:

- wire it behind a flag into `core/dataflow/smt_barrier.py`, where we already have the old line-number style sanitizer logic;
- only enable live suppression for Python CFGs where source/sink/call binding is clean;
- emit `candidate_only` when the graph cut looks promising but value binding is missing;
- keep C/C++ as candidate-only for now, because function-level call graphs cannot honestly prove argument flow;
- put the sanitizer call, source value, output value, and sink arg into `suppressions.jsonl`, otherwise the operator has to trust the magic box, which is pants.

That gives us a nice progression:

```text
Finding found -> sanitizer candidate -> value proof -> graph cut -> auditable suppression
```

If we do it this way, this becomes a tidy false-positive killer without accidentally binning real bugs. Proper useful, but only once the value-binding bit is in place.

unprovable:
ok starting on this - here's the first foray. Some more to come.

unprovable:
This update closes the wrong-variable false-suppress class flagged in review. The sanitizer-cut gate now verifies that the symbol the sanitizer actually cleans is the symbol the sink reads — not just that a sanitizer call exists somewhere between source and sink. Concretely: for `safe_other = html.escape(other); render(user.name)`, the gate now refuses to suppress (correctly — the sink reads user.name, not safe_other), where the prior lexical check would have falsely suppressed.

The same gate now runs end-to-end on C/C++, not just Python. Findings in .c / .cpp files build a real intra-procedural CFG via tree-sitter (tree-sitter-c / tree-sitter-cpp, 540 KB of wheels, no LLVM headers needed), with statement-level defs / uses / call_sites over if/else, while/for/do-while + break/continue, switch with case fallthrough, goto + labeled targets, return, C++ methods inside classes, and namespace-wrapped functions. Previously every C/C++ finding degraded to candidate_only because the recognizer only had function-name granularity; it now produces real symbol bindings and the 4-condition gate fires.

Pointer / alias indirection is handled conservatively: any `*p / &x / a[i] / obj->field` or call to a bulk-copy function (memcpy, memmove, memset, bzero, the strcpy / strcat / strlcpy family, sprintf / snprintf and v-variants, wcscpy / swprintf) on the source→sink path downgrades the verdict from suppress to candidate_only — the cleaned value's identity at the sink is unprovable without an alias analysis we deliberately don't run. Python is bit-identical to the prior behaviour (the attribute defaults to absent and the helper short-circuits).

Catalog gained four C/C++ sanitizer entries with documented soundness contracts: g_markup_escape_text (xss), g_uri_escape_string (pathtrav), g_shell_quote (cmdi), sqlite3_mprintf (sqli). Selection rule documented inline — only transform-style sanitizers that return a newly-allocated value qualify; destination-buffer writers like mysql_real_escape_string and realpath are excluded because the alias-conservatism above would downgrade them anyway.

Everything stays behind the RAPTOR_SANITIZER_CUT env flag. The legacy lexical check is still the fallback when the value-bound resolver can't build a CFG (missing source, syntax errors, unsupported language) or when the gate returns candidate_only. No verdict regressions on any prior fixture; 261 tests across the arc, ruff clean.

Soundness witnesses pinned as regression tests for both languages, plus a may_escape-on-path fixture (sanitize correctly but a snprintf writes into a buffer on the same path → correctly downgrades). The wrong-variable case — the exact shape from the review comment — is `core/inventory/tests/fixtures/sanitizer_cut_corpus/wrong_variable.py` and …_cpp/wrong_variable.c. Both correctly return candidate_only. Happy to split the C/C++ half into a separate PR if that helps the reader.

unprovable:
## Update: inter-procedural taint tracking landed (Sub-arc C complete)
  
  The value-bound sanitizer-cut gate now reasons across function boundaries for Python. A sanitizer that lives inside a helper function is
  recognised and counts toward suppression — previously these were missed entirely and surfaced as false positives.

  ### What this fixes

  The motivating shape, which the gate previously could not suppress:

  ```python
  def _sanitize(s):
      return html.escape(s)

  def handle(x):
      y = _sanitize(x)    # sanitization happens in the callee
      render(y)
  ```

  Looking only at `handle`, there's no `html.escape` call in sight — so the intra-procedural gate gave `no_suppress` and the finding went to the
  LLM as a candidate. The gate now sees that `_sanitize` cleanly escapes its argument and that the escaped value is what reaches `render`, so it
  correctly suppresses.
  
  ### How it works

  Three pieces:

  1. **Module-local call graph** — resolves which functions call which within a file (bare names, `self.m`/`cls.m`, `Class.m`, constructors).
  Cross-module, dynamic dispatch, and builtins are dropped rather than guessed.
  2. **Per-function taint summaries** — for each function, which parameters taint the return value, and which sanitizers the taint passed through
  on the way. Computed with a fixed point over the call graph (recursion/mutual-recursion safe), and marked *unknown* on
  `getattr`/`eval`/`exec`/`**kwargs` so dynamic code never drives a suppression.
  3. **Synthetic bindings** — when the analysed function calls a helper whose summary proves it cleanly sanitizes a parameter for the finding's
  CWE, a sanitizer binding is synthesised at that call site. It carries the real input/output symbols, so the existing suppression gate handles it
  identically to a direct sanitizer call.

  ### Soundness

  A helper only counts as a sanitizer when **every** path through the relevant parameter is provably clean:

  - a helper that returns its argument unchanged on any branch → not counted
  - a helper that sanitizes on one branch but passes through on another → not counted
  - a chain through any unrecognised callable (e.g. `wrap(html.escape(x))`) → not counted
  - the wrong-variable check still holds end-to-end: a helper that sanitizes `other` does **not** suppress a finding whose sink reads `user`

  Positional argument mapping is exact — a helper that sanitizes its second parameter binds the second call argument, not the first. Transitive
  sanitization (helper calls helper) is handled automatically.

  ### Verdict deltas on the regression corpus

  | Fixture | Before | After |
  |---|---|---|
  | `sanitizer_in_helper.py` | no_suppress | **suppress** |
  | all other fixtures | unchanged | unchanged |

  Exactly one case flips — the one that was a known false positive. Nothing else moves.

  ### Scope and tests

  - Inter-procedural reasoning is **Python-only** for now; C/C++ findings are unaffected (intra-procedural value-binding from the earlier push 
  still applies).
  - Cross-module helpers (callee in a different file) stay conservative — no summary, no suppression.
  - 71 new tests across the call graph, taint summaries, and inter-procedural binding layers. Full arc suite at 339 passing, lint clean. The 
  change is gated behind the same feature flag as the rest of the work; the legacy check remains the fallback.

  Note: I escaped the inner code-fence (the python block) with  zero-width markers so the whole thing renders as one copyable block here. When you
  paste, replace the ```python / ````` lines with plain triple-backticks.

unprovable:
One more set of things to come, to tie up a few things. Also, gonna do some testing then indicate I think it's ready: )

unprovable:
# Value-binding extension to the sanitizer-cut suppressor
  
Closes a soundness hole in the sanitizer-cut suppressor: it proves **control flow** through a sanitizer but not **value  flow**, so it falsely suppresses findings where the sanitizer cleans the *wrong symbol*.

  ```python
  safe_other = html.escape(other)   # sanitizer runs...
  render(user.name)                 # ...but the sink reads a different value it never cleaned
  ```

The old check sees "a sanitizer dominates the sink → drop it." The new gate proves the sanitizer's *cleaned value* actually reaches the sink before suppressing. The wrong-variable case is pinned as a regression test in both the Python and C/C++ corpora.
  
Everything is gated behind `RAPTOR_SANITIZER_CUT`; the legacy lexical check remains the fallback. **476 tests across the work, all green; ruff clean.**

## Python value binding (intra-procedural)

  - Symbol-aware CFG — per-node `defs` / `uses` / `call_sites`
  - Intra-procedural reaching-definitions
  - Symbol-bound sanitizer recognition (`SanitizerBinding` carrying the call's input and output symbols)
  - A four-condition gate that suppresses only when the sanitizer's cleaned value provably reaches the sink, with a tri-state verdict (`suppress` / `candidate_only` / `no_suppress`)
  - A finding-normalisation adapter (SARIF / Semgrep / RAPTOR-native)
  - A JSONL audit trail recording the verdict and its witnesses

  ## C/C++ value binding (intra-procedural)

  - **tree-sitter** as the substrate (chosen over libclang / r2-decomp — 540 KB of wheels, no LLVM headers; spike + comparison committed)
  - A C/C++ CFG builder mirroring the Python one — if/else, while/for/do-while + break/continue, switch + fallthrough, goto + labels, return, C++ methods-in-class, namespace-wrapped functions
  - **Pointer / alias conservatism** — any indirection (`*p`, `&x`, `a[i]`, `obj->field`, bulk-copy calls like  `memcpy`/`strcpy`) marks the value as possibly-escaped and downgrades `suppress → candidate_only`; we don't claim a cleaned value survives aliasing we can't track
  - Resolver wiring, a small set of GLib/SQLite sanitizer catalog entries, and a C corpus

  ## Python inter-procedural taint

  - A module-local call graph — resolves bare names, `self.m` / `cls.m`, `Class.m`, and constructors; cross-module and dynamic dispatch are dropped rather than guessed
  - Per-function taint summaries — which parameters taint the return value and which sanitizers the taint passed through; computed with a fixed point over the call graph (recursion-safe); marked unknown on `getattr` / `eval` / `exec` / `**kwargs`
  - **A sanitizer inside a callee now counts.** When the analysed function calls a helper that cleanly sanitizes a parameter, a synthetic binding is created at the call site, so `y = _sanitize(x); render(y)` suppresses correctly. Helpers that sanitize on only some branches, pass their argument through unchanged, or sanitize for the wrong CWE produce no binding.

  ## Measurement before removal

  - **Shadow telemetry** — when `RAPTOR_SANITIZER_CUT_PARITY_LOG` is set, both the lexical decision and the value-bound verdict are recorded for every finding, independent of the suppression flag, with zero overhead when off and no change to return values
  - **A/B aggregation** — an agreement matrix plus per-method noise-suppression and bug-hiding rates with Wilson 95% confidence intervals
  - A gate that treats removal of the lexical fallback as safe only when the value-bound path catches at least as much noise, hides no more real findings, **and** regresses on no individual finding

The data shows the lexical check and the value-bound gate are **complementary, not equivalent** — each covers finding shapes the other doesn't — so the lexical fallback is kept. `RAPTOR_SANITIZER_CUT_NO_LEXICAL` makes the value-bound-only behaviour available as a flag-flip; `lexical_fallback_status()` reports the current state.
  
  ## Operating it

  ```bash
  export RAPTOR_SANITIZER_CUT=1                          # enable the value-bound gate
  export RAPTOR_SANITIZER_CUT_PARITY_LOG=parity.jsonl    # collect lexical-vs-value-bound telemetry (optional)
  # RAPTOR_SANITIZER_CUT_NO_LEXICAL=1                     # value-bound only; "can't decide" → don't suppress
  ```

  ## What's written
  
  - `docs/design-sanitizer-cut-value-binding.md` — the full design and status
  - `docs/phase-8-substrate-spike/` — the tree-sitter vs libclang vs r2-decomp comparison and decision
  - `docs/sanitizer-cut-parity/` — the telemetry horizon, the first A/B report, and the operating notes
  - Per-area corpora with `CORPUS.md` ablation tables (Python and C/C++)

  ## Notes for the reviewer

  - The Python inter-procedural layer is **module-local**; cross-module helpers stay conservative (no summary → no
  suppression).
  - C/C++ inter-procedural analysis is out of scope here; C/C++ findings carry no inter-procedural bindings.
  - The graph protocol the gate consumes means a future language frontend slots in without touching the gate itself.

unprovable:
@danielcuthbert - back to yourself for further review... I think this is taking shape?

grokjc:
More stuff direct from Claude as I'm v. busy atm 😄 

### 1. Multi-arg helper FN — `core/inventory/python_interproc.py`

**Issue:** When the same symbol feeds multiple helper params and only one is sanitized, the gate suppresses incorrectly.

**Repro:**
```python
def helper(a, b):
    return html.escape(a) + b   # only a sanitized

def handle(x):
    y = helper(x, x)            # same x into both args
    render(y)                   # taint via b reaches sink
```

**Root cause:** Synthetic binding gets `input_symbols={"x"}` (set, no position info). Conditions 2+3 pass, verdict = SUPPRESS. Real bug, same shape as the wrong-variable case but one inter-proc level deeper.

**Fix:** Track per-position sanitization (`param_index → arg_name`), intersect at the sanitized position; or conservatively downgrade to `candidate_only` when the same arg name appears at >1 position. Add a regression fixture matching the repro above.

---

### 2. Missing exception handler — `core/dataflow/smt_barrier.py:_value_bound_dominates`

**Issue:** Design promises "resolver failure → lexical fallback" but only the `None`-return path falls through. Any exception (ImportError on optional tree-sitter wheel, KeyError in inventory, AST parse error) escapes the gate and crashes `/agentic` mid-run.

**Fix:** 
- Wrap the `resolve_finding()` / `evaluate_finding()` call in `try` / `except Exception: return None`.
- Drop `assert result.verdict == VERDICT_CANDIDATE_ONLY` — that's a future-proofing footgun; replace with a defensive conditional that treats unknown verdicts as fallthrough.
- Add a test that monkeypatches `resolve_finding` to raise, asserts the lexical path runs.

---

### 3. `_positional_arg_names` uses `ast.walk()` — `core/inventory/python_interproc.py`

**Issue:** `ast.walk()` order is implementation-defined. When two calls appear on the same source line (e.g., `z = helper(a) if helper(b) else None`), the function can grab the wrong call's arg list and attach the synthetic binding to it.

**Fix:** Replace with an in-order recursive descent, or disambiguate by the CallSite's column offset (already captured).

---

### 4. Env vars should be CLI flags

**Issue:** `RAPTOR_SANITIZER_CUT`, `RAPTOR_SANITIZER_CUT_NO_LEXICAL`, `RAPTOR_SANITIZER_CUT_PARITY_LOG` are tied to per-invocation `/agentic` / `/validate` behaviour, not process-global state.

**Footguns:**
- `NO_LEXICAL=1 + SANITIZER_CUT` unset silently disables ALL suppression (value-bound off + lexical off → every finding passes through, no warning).
- `PARITY_LOG=1` (boolean-style) writes telemetry to a file literally named `1` in CWD.

**Fix:** Move to CLI flags on the consuming commands, defaulting the log path under the active run directory:
```
--sanitizer-cut=off|on|strict|shadow   (default: off)
--sanitizer-cut-parity-log=<path>       (default: <run_dir>/sanitizer_cut_parity.jsonl)
```

Eliminates both footguns; surfaces the feature in `--help`.

---

### 5. Phase 13 fixed-point bail-out floor — `core/inventory/python_taint_summaries.py`

**Issue:** `3 × len(target_nodes)` under-converges small modules. For a module with N=2 mutually recursive functions, that's 6 outer iterations — not enough for a real chain through a sanitizer. Result: `summary_unconverged` → conservative `candidate_only`, losing a real suppression.

**Fix:** Floor at `max(10, 3 * N)`. While there, name it as a module-level constant (`_TAINT_SUMMARY_OUTER_ITERS`) and document the empirical basis in a comment.

---

  ### 6. Move `libexec/raptor-sanitizer-cut-parity-report` to `core/dataflow/scripts/`

  **Issue:** Zero callers in Python or workflows. Every other `libexec/` entry is invoked by RAPTOR runtime (`raptor-sca-run`, `raptor-smt-validate-path`, `raptor-strategy-eval`, `raptor-run-lifecycle`).
  Convention is `libexec/` = runtime substrate; per-subsystem `scripts/` = operator-invoked (existing siblings in `core/dataflow/scripts/`: `corpus-run`, `trust-corpus`, `trust-report`).

  **Fix:** Move to `core/dataflow/scripts/sanitizer-cut-parity-report` (drop the `raptor-` prefix to match siblings). Update references in:
  - `docs/design-sanitizer-cut-value-binding.md`
  - `docs/sanitizer-cut-parity/HORIZON.md`
  - `core/dataflow/tests/test_sanitizer_cut_parity_report.py`

---

### 7. Phase number in test filename — `core/inventory/tests/test_phase10_may_escape.py`

**Issue:** Sibling tests use descriptive names (`test_cfg_builder.py`, `test_sanitizer_cut.py`).

**Fix:** Rename to `test_may_escape.py` (or `test_cpp_may_escape.py` if scope clarity helps). Phase number is temporal, not semantic.

---

### 8. Language qualifier inconsistency

**Issue:** 
- `core/inventory/python_callgraph.py`, `python_interproc.py`, `python_taint_summaries.py` use a `python_` prefix
- `cfg_builder.py` / `cfg_builder_cpp.py` use a `_cpp` suffix

**Fix:** Drop the `python_` prefix on Python-only files (matches the `cfg_builder` pattern); only suffix when a second-language variant lands.

---

### 9. Phase-spike artifacts — `docs/phase-8-substrate-spike/`

**Issue:** `fixture.c`, `prototype_tree_sitter.py`, `prototype_tree_sitter.out` are one-shot research outputs that won't be re-run.

**Fix:** Delete and keep only `DECISION.md` (git history preserves the spike). Same convention applies to `docs/sanitizer-cut-parity/first-report.md` once subsequent reports land — either `archive/` or single rolling file.

unprovable:
Ok I think these cover it off. And no worries - busy times! So glad as ever for your help and expertise :)

grokjc:
Ah, what you really wanted to do was rebase so you get the test fixes in #799 and then you don't need [0de4e88](https://github.com/gadievron/raptor/pull/794/commits/0de4e885e56ac06177917b7e04910f9df7b80838) at all 😄 And post those changes, the tests aren't slow and shouldn't be opted-out ideally - so suggest you drop that last commit. Same on the other PR. Sorry!

grokjc:
# More Review Findings

## 1. CRITICAL — sqlite3_mprintf catalog entry is documented-unsound

**File:** `core/dataflow/known_safe_calls.py`

The entry's soundness note flags the gap:

> "only the %q/%Q specifiers are SQL-safe — generic %s in sqlite3_mprintf does NOT escape. The catalog keys on the call name; per-format discrimination is a future refinement."

**Issue:** `sqlite3_mprintf("SELECT * FROM t WHERE id = %s", user_input)` will be suppressed as SQL-injection-safe when it's actually vulnerable. This is exactly the wrong-variable/wrong-format-specifier class of FN the arc was built to prevent — shipped with the gap acknowledged inline.

**Fix:** Either remove the entry until per-format-specifier discrimination lands, OR add %q/%Q discrimination to the recognizer before merging. A documented-unsound entry in a security-suppression catalog is a contradiction.

---

## 2. Documentation drift after the python_ rename

**File:** `docs/design-sanitizer-cut-value-binding.md`

Commit b5a4cd96 renamed `python_callgraph.py` → `callgraph.py` (and siblings), but the design doc still references the old paths in 6+ places — Phase 12/13/14 table rows + Shipped/Tests sections (lines ~59-61, 503, 533, 554, 602, 630, 658). Operators reading the doc will look for files that don't exist.

**Fix:** find/replace the following in the design doc:
- `python_callgraph`
- `python_taint_summaries`
- `python_interproc`
- `test_python_callgraph`
- `test_python_taint_summaries`
- `test_python_interproc`

---

## 3. Parity JSONL has no finding_id deduplication

**File:** `core/dataflow/sanitizer_cut_parity.py` (`append_parity_record` + `aggregate_parity`)

Append-only writes with no dedup on `finding_id`. Agentic retries (e.g., LLM timeout retry, sanitizer-cut → lexical fallback shadow log) emit the same `finding_id` twice. `aggregate_parity` then counts both occurrences in the agreement matrix and rate-criterion calculations.

**Issue:** This biases the Phase 15 parity criterion (Wilson-CI gate) over time. Phase 16 lexical-fallback removal depends on this criterion clearing; if duplicates skew agreement upward, the gate could clear prematurely.

**Fix:** Dedup by `finding_id` before `aggregate_parity` runs (keep last record), OR enforce idempotent append (no-op when a record for this `finding_id` already exists in the active log).

---

## 4. Persisted config file is world-readable

**File:** `core/dataflow/sanitizer_cut_config.py:persist()`

JSON written with default umask permissions (0o644). Contents are mode-name + parity-log path — no credentials, low risk — but hygiene-wise the file should be 0o600 since it's a per-run internal artifact, not something other UIDs need to read.

```
  fd = os.open(path, os.O_WRONLY | os.O_CREAT | os.O_TRUNC, 0o600)
  with os.fdopen(fd, "w", encoding="utf-8") as fh:
      json.dump(payload, fh)
```
---

## 5. Test gap on Blocker 2 fix

**File:** `core/dataflow/tests/test_lexical_removal_switch.py`

The regression test monkeypatches `resolve_finding` to raise, but `evaluate_finding` raising is NOT separately tested. Both call sites are inside the same try/except Exception so it works in practice, but a future refactor that splits the wrap could regress one path silently.

**Fix:** Add a parallel test that monkeypatches `evaluate_finding` to raise.

---

## 6. Binary-oracle vs sanitizer-cut interaction undocumented

**Files:** 
- `core/inventory/binary_oracle_edges.py`
- `core/inventory/reach_chokepoint.py`

Both write to `suppressions.jsonl` (per CLAUDE.md's "Audit trail" section). Order-of-operations between the two chokepoints isn't documented in either file. For a function marked absent by binary-oracle that ALSO would trip the sanitizer-cut gate, which audit-record wins? Does the parity log record findings the binary-oracle already dropped?

**Fix:** Add a comment block to each chokepoint stating the logical order and whether they observe each other's verdicts. Schema already accommodates both (additive design) — this is documentation, not code.

---

## 7. Stale-config edge case in /validate (pre-existing, low priority)

If stage 0 calls `configure(export_env=True)` then crashes BEFORE `persist()`, the env vars are mutated for the rest of the parent process. A subsequent `raptor-validation-helper <dir>` manual invocation would `load_persisted` → `None` → env fallback, using the leftover env from the failed stage 0. 

The happy path is correctly handled because the run is marked failed and pipeline-driven stages won't run; this only bites operators driving stages by hand. Worth a comment in the validation-helper.

unprovable:
Ok, Critical should be fixed in this - working on the rest :)

unprovable:
ok, I think just the documentation to sort now. 

- parity log has no finding_id dedup. Append-only, so agentic retries (LLM-timeout, sanitizer-cut → lexical fallback shadow log) emit the same finding_id twice and aggregate_parity double-counted them, biasing the Phase-15 Wilson-CI gate that gates Phase-16 lexical removal. aggregate_parity now collapses duplicates by finding_id (keeping the latest verdict) before counting; regression test added.
- persist() wrote the per-run config with the umask default (0o644, world-readable). Create it 0o600 via os.open instead.
- the resolver/evaluator failure path only tested resolve_finding  raising; both call sites share one try/except, so add a parallel test that makes evaluate_finding raise (resolve_finding succeeds first) and asserts the gate still falls through to lexical rather than crashing.

unprovable:
Second-round review addressed — commits per concern:

  - **1 (critical)** removed the `sqlite3_mprintf` catalog entry — call-name keying is unsound when the format is `%s`; re-add only behind
  `%q`/`%Q` discrimination. The other 3 C/C++ entries escape unconditionally and stay.
  - **3** `aggregate_parity` now dedups by `finding_id` (keep-last) so retries don't bias the Phase-15 gate (+ test).
  - **4** persisted config now created `0o600`.
  - **5** added the parallel `evaluate_finding`-raises test.
  - **6** documented the `suppressions.jsonl` order-of-ops. One note: there's a **single writer** (`reach_chokepoint.record_suppression`); `binary_oracle_edges.py` doesn't write it. The sanitizer-cut bridge writes the same schema but isn't wired into the pipeline yet, so there's no live interaction; intended order is binary-oracle (pre-LLM) first.
  - **7** documented the `/validate` configure-before-persist env-leak edge (hand-driven stages only).
  - **2** already resolved — the design doc has no stale `python_` paths (fixed in `b5a4cd96`).
  - **Slow-marks** dropped to rely on #732/#799 amortization; letting CI confirm the e2e test stays under the guard.

grokjc:
Thanks - LGTM!

PR #801, opened 2026-06-12
feat: add SiftRank skill helper

noperator:
Used GPT 5.5 to draft an initial implementation of a SiftRank integration into Raptor. I haven't used Raptor yet so I'm not very familiar with its architecture, but it seems like the best way is putting a SiftRank wrapper into `libexec/` that can be called by a Claude skill. That's the approach I took here, but feel free to suggest something different @danielcuthbert :) I also followed other examples in the codebase that fetched OpenAI API creds from the Raptor models config file. I tested this out with a few basic examples and it seems to work well. The code and tests can likely be tightened up. 

Coding agent summary here:
- Add a Claude Code skill for using SiftRank to prioritize candidate security items before deeper RAPTOR review.
- Add libexec/raptor-siftrank as a trusted helper that reads RAPTOR model configuration, ensures the external siftrank CLI is available, invokes it with an OpenAI-compatible analysis model, and writes ranked JSON while preserving original candidate items.
- The skill documents when SiftRank is useful, how to prepare candidate JSON, how to use named or literal prompts, and how to interpret ranked output as a review queue rather than a validation result.
- Add focused tests for deterministic helper behavior, model selection, SiftRank output mapping, and Go-based SiftRank installation paths.

### DIFF
--- a/.claude/skills/siftrank/SKILL.md
+++ b/.claude/skills/siftrank/SKILL.md
@@ -1,0 +1,71 @@
+---
+name: siftrank
+description: Prioritize many security-relevant candidate items with SiftRank when review attention is scarce, relative comparison is easier than absolute scoring, and RAPTOR needs an ordered queue before expensive validation or deeper analysis.
+user-invocable: false
+---
+
+# SiftRank Skill
+
+Use SiftRank for attention-constrained prioritization. It is useful when RAPTOR or Claude has many candidate security-relevant items and the next problem is deciding what deserves scarce review time first.
+
+SiftRank is a ranking aid, not a validator. It does not prove reachability, exploitability, severity, or correctness. Use it to choose review order, then continue normal RAPTOR validation or direct code review.
+
+## Why This Helps
+
+Use SiftRank when relative comparison is likely easier than absolute scoring. LLMs are often noisy when asked to judge one item in isolation or assign calibrated numeric scores, but they can still provide useful signal when asked to compare several items and rank them by relevance to a query.
+
+SiftRank is especially useful as a broad triage pass before expensive analysis. A lower-cost model can rank many candidates, allowing Claude, a human analyst, or a higher-tier model to spend deeper reasoning on the most promising prefix of the list.
+
+## When to Use
+
+Use this when there are many candidate items and manual or agentic review would be expensive, such as:
+
+- scanner findings
+- CodeQL or Semgrep results
+- suspicious files or functions
+- decompiled functions
+- call chains or traces
+- crash reports
+- dependency findings
+- exploit hypotheses
+- fuzz targets
+- injection points
+- validation notes or intermediate tool outputs
+
+Good SiftRank problems look like needle-in-a-haystack triage: the relevance criteria may be fuzzy, but a useful item should be recognizable when compared against nearby alternatives.
+
+Do not use it for:
+
+- only a few items
+- final exploitability verdicts
+- deterministic severity sorting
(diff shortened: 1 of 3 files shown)


grokjc:
Hey - thanks for putting this together. The core idea here is genuinely interesting: pairwise comparison as a triage primitive is a smart approach to the "too many findings" problem.

RAPTOR already has LLM-driven triage in its analysis pipeline, so one option worth exploring is implementing the pairwise-ranking primitive natively which would give us tighter control over prompts, cost tracking, and finding-schema integration. But the siftrank approach has the advantage of being battle-tested in your own workflows. 

A few thoughts on how to take this forward and potentially cement an approach:

  Dependency management: the auto-install via go install is clever for local experimentation, but for the shipped tool we'd want the operator to bring siftrank themselves (like we do other dependencies) with a clean probe, actionable error message, no runtime fetching.  We're also a big fan of pinning versioned dependencies to avoid nasty surprises.
That said, the right shape might not be an external binary at all.

  Model selection: the model config functions you're reaching into are internal/private, and the OpenAI filter means operators running Anthropic or via Bedrock would be locked out. RAPTOR is model-agnostic by design - so worth wiring through the public model-selection surface so any configured provider works 'for free'.

Integration point: this is the big one, and also the most intriguing question: where in the pipeline does this fire? After /scan? Before /validate? As a step inside /agentic? Once there's a concrete caller with a defined input/output contract, everything else (schema, test placement, prompts) falls into place naturally. This is where we'd love your ideas.

Test location: Small thing, but tests for libexec/ scripts should live closer to the owning module rather than in packages/llm_analysis/tests/. Easy to move.

If you want to keep going with this, I'd suggest starting from the integration point question - as once we know where it plugs in, we can figure out the ideal shape together.

danielcuthbert:
This is cool @noperator thanks for pushing this. I'm with @grokjc here, if you could help us with the wiring, that would be epic

grokjc:
Thanks for inspiring me to read the original SiftRank paper. I've created our own native implementation in 8960d82e7dd813ad4dda028761b49d3ebbca95db. SiftRank is now fully integrated.

**The core:** `core/llm/ranking.py` - an independent from-spec implementation of SiftRank (§3.2 of the paper) on RAPTOR's internal LLM layer. The landing commit credits you as co-author, and the paper + MIT reference implementation are cited in the module and in `docs/references.md`.

**Consumers wired so far** (all opt-in, all ordering-only):

- `/agentic --rank` — findings re-ranked most-promising-first before the budget caps, so `--max-findings` / `--max-cost` cut the least promising tail instead of scan order.
- `/audit` — `gaps --rank` for the in-session flow and `run --rank-gaps` for the orchestrator: the unreviewed-function
  queue re-ranked strictly within its mechanical priority tiers.
- `/validate --rank` (stage 0) — the attack-surface bridge's priority targets re-ranked before the validation stages assess  them in order.
- `/understand --hunt ... --rank` — merged multi-model variant candidates re-ranked before `hunt-result.json` is written.
- `/fuzz --rank-crashes` — the crash queue re-ranked before the per-crash deep-analysis cap, over a widened pool with  stack-dedup backfill.
- `raptor-binary harness --llm-rank` — ingress candidates re-ranked ahead of harness-boundary selection.
- `raptor-cve-diff localize` — your paper's §4 recipe end to end (call-chain expansion, k/r weighting, mass×density call-graph clustering) for advisory-to-patch localization when no fix commit is published.

Thanks again for pushing the idea!

PR #802, opened 2026-06-12
feat(cve-env): integrate cve-env as packages/cve_env (Phase 1 lift-and-shift)

gadievron:
Brings `cve-env` (an agentic CVE → Docker-environment builder) into raptor as `packages/cve_env`, wired like the sibling `cve_diff` tool. Phase 1 is a behavior-preserving lift-and-shift: cve-env keeps its own engine and adopts zero `core/` modules, so the import is auditable as "same tool, new home."

Imported from `gadievron/cve-env` @ `ba9f91c` (see `packages/cve_env/PROVENANCE.md`).

### Package import
**What:** vendored copy of cve-env at `packages/cve_env/cve_env/` (flat layout, mirroring `packages/cve_diff/cve_diff/`) + its test suite + `__init__`/`__main__` + `LICENSE` + `PROVENANCE.md`.
**Counts** (re-derived):
- `find packages/cve_env/cve_env -name '*.py' | wc -l` → 41 source files
- `find packages/cve_env/tests -name '*.py' | wc -l` → 89 test files
- `git diff --stat origin/main...HEAD | tail -1` → 142 files changed, 45379 insertions(+), 2 deletions(-)
**Scope note:** all `cve_env.` imports are absolute, so the `src/cve_env/` → `packages/cve_env/cve_env/` move needs no source-import rewrites. `grep -rE '^(from|import) (core|packages)\.' packages/cve_env/cve_env` → 0 matches (zero raptor-`core` adoption this phase).

### Pattern-B wiring (mirrors cve_diff)
**Solution:** `bin/cve-env` (launcher: env-strip + `PYTHONPATH` + trust marker), `libexec/raptor-cve-env` (non-interactive dispatcher with the inline trust-marker guard), `.claude/commands/cve-env.md` (`dispatch: libexec/raptor-cve-env <subcommand> [args]`). Subcommands are cve-env's own `build` / `doctor` (argparse), not Typer.
**Verification:** `python3 .github/scripts/check_command_metadata.py` → `OK: 28 command .md files lint-clean` (includes `cve-env.md`).

### Layout-coupled path fixes
**Problem:** three sites resolved paths by file-depth and would mis-place artifacts under the new nesting (audit JSONL, outcome sidecar, refusals log).
[...]

PR #803, opened 2026-06-13
fix(cve-env): floor interrupted-exit cost by turns under session auth

gadievron:
## What

Floor a build's reported `total_cost_usd` by its turn count when the run ends on
an interrupted status (`turn_cap` / `budget_exhausted` / `error` / `interrupted` /
`incomplete` / `rate_limited`) and the SDK reports no token usage — the Claude
Code session-auth case. Ports upstream `gadievron/cve-env` `89917d8` (PR #2) into
the vendored `packages/cve_env/`.

> **Depends-on: #802** — `packages/cve_env/` is introduced by the integration PR;
> this fix stacks on `feat/cve-env-integration`. GitHub will retarget to `main`
> once #802 merges.

## Why

Both `Outcome` finalization sites in `packages/cve_env/cve_env/agent/loop.py`
(the normal `_map_status` path and the exception/terminal path) floored
`total_cost_usd` on `max(last_cost_usd, continuation_cost, token_estimate)`.
Under Claude Code **session auth** the SDK sets `usage=None` on every message, so
`state.total_input/output_tokens == 0` and `estimate_cost_from_tokens(...) == 0`.
On an interrupted exit the SDK's `ResultMessage.total_cost_usd` is also
implausibly low, so `max()` collapsed to that value even though `num_turns`
(from the authoritative `state.turn`) stayed correct — a 46-turn build logged
`$0.013`.

## Root cause

The cost floor had no fallback derived from the engine turn count, so an
interrupted run with absent token usage lost its accumulated cost while keeping
an accurate turn count.

## How

- `config.py:estimate_cost_from_turns(num_turns, model)` — a turns-based lower
  bound (`5000` in / `500` out tokens per turn, sized from the existing
  `test_b19_b20_cost_extension.py` "~5K in, ~500 out per LLM round" figure),
  routed through `estimate_cost_from_tokens`; returns `0.0` for non-positive turns.
- `loop.py:_floor_cost(...)` — a shared helper applied at both `Outcome` sites.
  Base floor unchanged; adds the turns-based floor **only** when the status is in
  `_INTERRUPTED_EXIT_STATUSES = {turn_cap, budget_exhausted, error, interrupted,
  incomplete, rate_limited}` AND there is no token usage, bounded by
  `effective_max_cost_usd` so the estimate never exceeds the run's budget cap.
- Clean exits (`success`, `verified_partial`, `verify_failed`,
  `launched_no_verify`, `unresolvable`) end via `end_turn` with reliable SDK cost
  and are excluded so the floor never inflates a correctly-reported cost.

## Regression test

5 test functions in `packages/cve_env/tests/unit/test_cost_floor_non_clean_exit.py`,
including 2 parametrized over the full OutcomeStatus partition (floor fires for
every interrupted status, never for a clean one).

```
Vendored package (pytest tests/unit/test_cost_floor_non_clean_exit.py): 14 passed
Regression (verify + map_status + loop + cost_floor): 244 passed
raptor ruff-pr gate (--select F401,F811,F821,F841): All checks passed
```

## Compatibility

No API change. Clean exits and token-bearing (API-key) runs are untouched; the
floor only ever raises a collapsed cost. Vendored-copy change only — no raptor
`core/` modules touched.

## Author notes

**Q1 — which lines implement the fix?** `config.py` `estimate_cost_from_turns`
and `loop.py` `_floor_cost` gated on
`status in _INTERRUPTED_EXIT_STATUSES and input_tokens == 0 and output_tokens == 0`,
called at both `Outcome(... total_cost_usd=_floor_cost(...))` sites.

**Q2 — one concrete input handled wrong before, right now:** an interrupted
outcome with `ResultMessage(total_cost_usd=0.013, num_turns=46, usage=None)`
previously logged $0.013; it now logs a turns-proportional floor bounded by the
budget cap.

**Q3 — likely reviewer pushback + answer:** "Why not re-sync the whole vendored
package instead of cherry-picking?" This is a single targeted upstream fix with a
regression test; a full re-sync is a larger change better done deliberately. The
divergence is recorded in `PROVENANCE.md`.

Upstream: `gadievron/cve-env` `89917d8` (PR #2, merged).

---

> [!NOTE]
> **Low Risk**
> Telemetry-only change behind strict status and zero-token gates; clean exits and token-bearing runs are untouched, with regression tests covering the status partition.
> 
> **Overview**
> Ports upstream **cost-floor on interrupted exits** into vendored `cve-env`: when a build ends on an abnormal status and the SDK reports no token `usage` (Claude Code session auth), `Outcome.total_cost_usd` no longer collapses to a tiny SDK value while `num_turns` stays accurate.
> 
> **`estimate_cost_from_turns`** in `config.py` derives a conservative lower bound from turn count (5K in / 500 out tokens per turn via existing `estimate_cost_from_tokens`). **`_floor_cost`** in `loop.py` centralizes final cost: `max(SDK cost, continuation cost, token estimate)`, then optionally applies the turns floor only when `status ∈ _INTERRUPTED_EXIT_STATUSES` and both token totals are zero, capped by `effective_max_cost_usd`. Both normal and exception-path `Outcome` construction call `_floor_cost` instead of inline `max(...)`.
> 
> Clean exits (`success`, `verified_partial`, etc.) and API-key runs with real token usage are unchanged. **`PROVENANCE.md`** records the cherry-pick from upstream `89917d8`. New unit tests parametrize interrupted vs clean statuses and integration cases for turn_cap.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 93de23e9a1786b26ce8d2b9883bc2887ab6d7c38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/cve_env/PROVENANCE.md
+++ b/packages/cve_env/PROVENANCE.md
@@ -8,3 +8,9 @@ This package was imported from the standalone repository **gadievron/cve-env**.
 - Not copied: `pyproject.toml`, `uv.lock`, virtualenvs, caches, `cve-env.toml.example`. Dependencies are declared in the repo-root `requirements.txt` per raptor's "no per-package build config" convention.
 
 Phase 1 of the integration is a behavior-preserving lift-and-shift: cve-env keeps its own agent loop (claude-agent-sdk), Docker tooling, dockerfile generation, config, and HTTP layer. It adopts **zero** raptor `core/` modules in this phase. Selective `core/` adoption is deferred to a later phase behind behavior-equivalence checks.
+
+## Divergences from the imported snapshot
+
+The vendored copy tracks upstream `gadievron/cve-env` with cherry-picked fixes applied on top of the `ba9f91c` snapshot:
+
+- **Cost-floor on interrupted exits** (this PR) — ports upstream cve-env `89917d8` (PR #2): floors `total_cost_usd` by engine turn count when a build ends on an interrupted status with no token usage (the Claude Code session-auth case), so interrupted runs no longer log ~$0. Files: `cve_env/config.py` (`estimate_cost_from_turns`), `cve_env/agent/loop.py` (`_floor_cost` + `_INTERRUPTED_EXIT_STATUSES`), `tests/unit/test_cost_floor_non_clean_exit.py`.
--- a/packages/cve_env/cve_env/agent/loop.py
+++ b/packages/cve_env/cve_env/agent/loop.py
@@ -86,6 +86,7 @@ from cve_env.config import (
     TURN_EXTENSION_PCT,
     VERSION_ASSERTION_CMD_PATTERN,
     estimate_cost_from_tokens,
+    estimate_cost_from_turns,
     get_benign_verify_continuation_max,
     get_enable_benign_verify_continuation,
     get_enable_halt_on_verified_success,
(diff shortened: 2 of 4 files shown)


gadievron:
**Dependency: this PR depends on #802.**

#803 is stacked on `feat/cve-env-integration` (the branch #802 introduces) because the file it patches — `packages/cve_env/` — only exists on that branch, not yet on `main`. The diff here is the single fix commit `93de23e9` on top of #802.

Merge ordering (pick one):
1. **Merge #803 into `feat/cve-env-integration` first**, so #802 carries the cost-floor fix when it lands on `main`; or
2. **Merge #802 first** — GitHub then auto-retargets #803 to `main` and its diff stays just the fix.

Do not merge #803 to `main` before #802, or the base will be missing `packages/cve_env/`.

PR #813, opened 2026-06-16
chore: Use correct param for claude code system-prompt 

igama:
Update the system prompt param for Claude Code. It should be `--system-prompt`. When Anthropic first launched the [Claude Code](https://code.claude.com/docs/en/cli-reference) CLI, `--system` was the original flag used to override the default system prompt. Not supported anymore.

```
vscode ➜ /workspaces/raptor $ claude --help
Usage: claude [options] [command] [prompt]

Claude Code - starts an interactive session by default, use -p/--print for non-interactive output

Arguments:
  prompt                                            Your prompt

Options:
  --add-dir <directories...>                        Additional directories to allow tool access to
  --agent <agent>                                   Agent for the current session. Overrides the 'agent' setting.
(....)
  --system-prompt <prompt>                          System prompt to use for the session
  ```

### DIFF
--- a/core/llm/cc_adapter.py
+++ b/core/llm/cc_adapter.py
@@ -26,7 +26,7 @@ class CCDispatchConfig:
     timeout_s: int = 300
     json_schema: dict[str, Any] | None = None
     capture_json_envelope: bool = True
-    # System prompt passed via the `--system` flag rather than
+    # System prompt passed via the `--system-prompt` flag rather than
     # prepended to the user prompt. Pre-fix `ClaudeCodeLLMProvider`
     # concatenated `f"{system_prompt}\n\n{prompt}"` and sent the
     # combined text as the user message — the model then saw it as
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by grokjc** [comment]
Hi - thanks for this (well spotted!) I think we need to version-gate this. Do you know when the CLI option changed/which version numbers are at the cut-off? 

Alternatively we could do something like this
```
  _RESOLVED_SYSTEM_FLAG: str | None = None  # process-cached after first probe
  
  def _system_prompt_flag(claude_bin: str) -> str:
      """Return whichever flag this CC build accepts. Probes once per process."""
      global _RESOLVED_SYSTEM_FLAG
      if _RESOLVED_SYSTEM_FLAG is not None:
          return _RESOLVED_SYSTEM_FLAG
      # Cheaper probe than a real dispatch: claude --help, grep.
      try:
          out = subprocess.run(
              [claude_bin, "--help"], capture_output=True, text=True, timeout=5,
              env=RaptorConfig.get_safe_env(),
          )
          text = (out.stdout or "") + (out.stderr or "")
          _RESOLVED_SYSTEM_FLAG = (
              "--system-prompt" if "--system-prompt" in text else "--system"
          )
      except (subprocess.TimeoutExpired, FileNotFoundError, OSError):
          _RESOLVED_SYSTEM_FLAG = "--system-prompt"  # newer is the safer default
      return _RESOLVED_SYSTEM_FLAG
  
  Then cmd.extend([_system_prompt_flag(claude_bin), config.system_prompt]) at the call site.
```

I'd like this to work for all users, even if they are using very new or very old versions of CC.

grokjc:
Or, it occurs to me that perhaps --system was never a real registered flag, and it was accepted as a unique prefix abbreviation of --system-prompt until --system-prompt-file was added and made the abbreviation ambiguous?

igama:
@grokjc I think it was that. Looking at what google provided, --system was only available in initial versions of claude code, so might have been a "non official" helper.

grokjc:
In which case, thanks! LGTM - appreciate the contribution.

PR #815, opened 2026-06-19
feat: /frida dynamic instrumentation, CLI QoL, and code-review hardening

ZephrFish:
Dynamic instrumentation
- New /frida command: spawn-only Frida function/argument tracing and Stalker basic-block coverage (drcov), sandboxed and degrading gracefully when the frida binding is absent. Documented in docs/DYNAMIC_INSTRUMENTATION_QUICKSTART.md + DEPENDENCIES + README.

CLI quality-of-life
- daily-loop helpers: 'last'/'latest' run tokens, agentic next-steps hints, --version, doctor --json, describe visibility, short flags, cleaner errors.
- Shared run-resolver: 'last'/'latest'/unique-substring now work uniformly across project show / diff / annotations-diff / remove.
- --json output for project status / findings / coverage (joining doctor and describe) for CI parsing.
- Next-steps hints after /scan and /codeql.
- Fail-closed flag validation: malformed/out-of-range --max-cost-usd, project clean --keep, project coverage --fail-under, and fuzz/frida --timeout now exit 2 instead of running with a bad value.

Refactor
- Factor the duplicated libexec trust-guard into core.security._trust_guard; add bash/zsh completion.

Security & correctness fixes
- Mask the remote Ollama endpoint in logs and errors.
- Close the $EDITOR argument-injection vector (allow only -w/--wait).
- Bound bin/ symlink resolution with a 32-hop guard.
- Spawn JDK/rustup probes with get_safe_env.
- Opt-in CodeQL credential isolation for untrusted builds.
- Sandbox binutils ELF parsers; planted-binary source-coverage floor at any project size.
- reachability: require unanimity for line-less same-name suppression; an affirmative binary call-edge beats a name-lookup 'absent'.
- dispatch: stop spending on a model declared dead; back-fill aborted work by (item, model) pair; tighten retry/quota error classification.
- orchestrator: --aggregate requires >=2 analysis models.
- agentic: fail the run on analysis failure; single finalize owner; reanalyze uses the prior run's recorded target.
- reach-cache: derive header magic from _CACHE_VERSION.
- sandbox: RAPTOR_REQUIRE_SANDBOX fail-closed kill-switch.

Credit to @Splinters-io for original Frida wiring idea

### DIFF
--- a/.claude/commands/frida.md
+++ b/.claude/commands/frida.md
@@ -1,0 +1,76 @@
+---
+description: Dynamic instrumentation with Frida (function trace, Stalker coverage)
+dispatch: skill
+---
+
+# /frida - Dynamic Instrumentation (Frida)
+
+Spawn an operator-chosen binary under Frida and instrument it: trace function
+entry/arguments, or collect Stalker basic-block coverage that flows into
+RAPTOR's coverage store (and thus `/project coverage` and the binary-oracle
+reachability view). This is the dynamic complement to the static analysis,
+binary-oracle, and fuzzing surfaces.
+
+**Authorization & safety:** `/frida` is **spawn-only by default** - RAPTOR
+instruments a binary it launches, not arbitrary already-running PIDs. The
+target runs under the sandbox (network blocked, credentials unreadable) using
+the `debug` profile (ptrace-permitted). Only instrument targets you are
+authorized to test, same as `/fuzz` and `/crash-analysis`.
+
+**`--help`:** run `CLAUDECODE=1 libexec/raptor-frida probe` to check whether
+Frida is available on this host (no spawn, no side effects).
+
+## Your task
+
+1. **Identify the target binary** (full path) and what the operator wants:
+   - function/argument **trace** of specific symbols, or
+   - **coverage** (which basic blocks / source lines executed).
+   Ask if unclear. Note any target arguments (passed after `--`).
+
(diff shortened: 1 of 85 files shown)


grokjc:
Hi! There's some great stuff in here but I am concerned this has the same issue that the original (aborted) Frida PR has - there are at least 4 different unrelated things in one commit here and that makes reviewing and integration very difficult. Can we split them out?

Also, I have to push back on the refactor.  I made a conscious design decision not to centralise the trust marker stuff (in fact, I did exactly what you suggested and then reverted it) - the check is a six-line inlined code block that has minimal dependencies, don't rely on the path being correct, and has a near-zero chance of breakage.  Putting it into that module means that a single syntax error, path issue, or other problem breaks ~50 scripts and also breaks our CI that effectively greps for that marker in a simple reproducible way.  And honestly I don't want two competing ways of enforcing what should be consistent.

Security fixes are great, QoL stuff too, and I need a proper look at the Frida stuff but yes please in general.

ZephrFish:
Yeah I'm happy to break out the commits separately, is it easiest for me to put in different pulls with the different changes or what would you prefer? 

No problem re refactoring, happy to change whatever is needed

ZephrFish:
I'll close this and open separate PRs with the split https://github.com/gadievron/raptor/pull/817 has QoL

grokjc:
Thanks, appreciated

PR #817, opened 2026-06-20
fix(qol): portability and test-guard improvements

ZephrFish:
Follow up on https://github.com/gadievron/raptor/pull/815 submitting split out pull request with QoL improvements separately for review, will split out others too and open PRs

Portability:
- shellcode_emulator: resolve AArch64 via getattr fallback (capstone >=5 renamed CS_ARCH_ARM64 -> CS_ARCH_AARCH64); eager dict broke x86/x64 too
- sandbox/fingerprint: portable _host_cpu_count() (sched_getaffinity is Linux-only) so build_persona(HOST_CPU_COUNT) does not crash off Linux
- pwntools_scaffold: drop stray f-prefixes (F541)

Test guards:
- gate Linux-only tests on the real precondition: ELF magic (not /bin/ls), /proc, and the codeql/python-queries pack being resolvable
- make core/llm tests hermetic (pin OLLAMA_HOST, clear provider keys)
- tuning: mock RAM in the cpu-backed worker-limit test (non-hermetic on low-RAM CI runners)
- codeql e2e: skip when pack is not resolvable

### DIFF
--- a/core/binary/tests/test_elf.py
+++ b/core/binary/tests/test_elf.py
@@ -92,10 +92,17 @@ class TestNegativeCases:
 class TestRealBinaryParse:
     @pytest.fixture(autouse=True)
     def _require_elf_host(self):
-        """All tests in this class need a Linux-shaped host with
-        /bin/ls. macOS test runners (Mach-O coreutils) skip."""
-        if not Path("/bin/ls").exists():
-            pytest.skip("/bin/ls not present on host")
+        """All tests in this class need a Linux-shaped host with an
+        ELF /bin/ls. macOS test runners (Mach-O coreutils) skip — the
+        old guard only checked existence, so it failed to skip on a
+        host where /bin/ls exists but is Mach-O."""
+        p = Path("/bin/ls")
+        try:
+            is_elf = p.is_file() and p.read_bytes()[:4] == b"\x7fELF"
+        except OSError:
+            is_elf = False
+        if not is_elf:
+            pytest.skip("/bin/ls is not an ELF binary on this host")
 
     def test_parse_bin_ls(self):
         """Parser doesn't crash on a real binary; the metadata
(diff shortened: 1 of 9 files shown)


grokjc:
The new portable helper replaces `len(os.sched_getaffinity(0))` in `build_persona`, but `set_cpu_affinity()` (line ~487) still calls `os.sched_getaffinity(0)` and `os.sched_setaffinity()` directly — both are Linux-only. So `build_persona(cpu_count=HOST_CPU_COUNT)` would succeed on macOS, then `set_cpu_affinity()` would AttributeError. The rest of the fingerprint system is inherently Linux-only too (fakes  /proc/cpuinfo, /sys/devices/system/cpu/, calls setdomainname via ctypes).
  
Was the intention to make `build_persona` independently usable off-Linux (e.g. for test fixture generation on macOS), or is this heading toward broader portability? If the latter, `set_cpu_affinity` and the ctypes paths need the same treatment.

grokjc:
Also, the shellcode emulator etc are completely new capabilities rather than 'portability and test-guard improvements' 😄 Ideally we'd split them out too - and maybe hold fire -  and get the other fixes merged faster. There's a lot of stuff that needs reviewing in those features as they conflict with some forthcoming work that has been in dev for a while and neither have consumers on main right now. They also (silently) introduce deps that we currently don't have anywhere else, so I'd recommend carving those bits off and landing separately at some point as a whole PR as needed.

ZephrFish:
My bad, I've been working against a local fork that has some additional functionality pulled in which isn't part of this PR. The shellcode-related pieces came from a broader MCP/RAG pipeline that also includes Raptor, so I can split those changes out separately.

The `_host_cpu_count()` change was only intended to stop `build_persona()` blowing up on macOS when generating test fixtures offline. No wider portability work was intended; the rest of the fingerprinting implementation remains Linux-only by design. `set_cpu_affinity()` was missed largely because this was developed and tested on Linux. 

In practice it's only called from `_spawn.py` in the post-fork path, which is already Linux-gated, and the associated tests are wrapped with `skipif(sys.platform != "linux")`, so nothing is actually broken today in testing. That said, I agree that raising an `AttributeError` is a bit of a shooting yourself in the foot if someone calls it directly on a non-Linux platform. Adding an explicit guard such as something along the lines of this, will make it cleaner:

```
 if not hasattr(os, "sched_getaffinity"):
      raise NotImplementedError("set_cpu_affinity requires Linux (sched_setaffinity syscall)")
```

grokjc:
Thanks - going to merge this now.  The other stuff is interesting but need to understand the Pwntools templating vs the /exploit work that I was talking about the other day (some of which I have started to merge)

PR #818, opened 2026-06-20
fix(config/llm): harden path guard, Ollama URL normalisation, host_is_local

ZephrFish:
Security fixes split per https://github.com/gadievron/raptor/pull/815

- get_out_dir: check both os.path.normpath(input) and resolved path so a literal RAPTOR_OUT_DIR=/etc doesn't slip the system-path guard on macOS where /etc is a symlink into /private/etc.
- _validate_ollama_url: accept schemeless host:port (Ollama's canonical 127.0.0.1:11434) by defaulting the scheme to http://; reject empty.
- _host_is_local: parse the hostname with urllib.parse instead of substring-matching so localhost.attacker.example and 127.0.0.1.evil are not mis-classified as local hosts.
- client.py: replace the same substring check with _host_is_local.


grokjc:
Thanks!

PR #836, opened 2026-06-21
feat(understand): mechanical sink discovery and framework API detection (MAP-5f)

grokjc:
Add a new /understand --map enrichment step that mechanically discovers dangerous sinks, computes transitive reverse reachability through the call graph, and autonomously detects framework APIs from call-target frequency.

- core/inventory/sink_discovery.py: language-agnostic sink discovery engine — matches call chains against dangerous targets (merging function_taxonomy.EXEC_FUNCS for C with source-level sinks for Python/JS/Lua/Go/Java), BFS reverse reachability, framework API discovery from cross-file call frequency with builtin filtering
- core/orchestration/context_map_sinks.py: bridges sink discovery into context-map.json — merges mechanical sinks into sink_details, adds reachable_sinks to entry points, populates meta.frameworks_discovered
- libexec/raptor-enrich-context-map-sinks: MAP-5f libexec shim
- Adversarially reviewed: fixed BFS diamond-graph sink-set propagation, empty-chain crash, self.eval() FP avoidance, idempotency bugs
- 53 tests including E2E with synthetic and real targets

### DIFF
--- a/.claude/skills/code-understanding/map.md
+++ b/.claude/skills/code-understanding/map.md
@@ -313,6 +313,31 @@ context-map's entry points and sinks. Entry points and sinks that frida
 confirmed at runtime get a `runtime_confirmed: true` annotation. Idempotent.
 Skip-silent when no frida evidence is discoverable.
 
+**[MAP-5f] Enrich with mechanically-discovered sinks and framework APIs**
+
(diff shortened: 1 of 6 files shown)


grokjc:
No lua support in CI, need to rewrite tests...

PR #843, opened 2026-06-22
Feat/Path homology as a CFG structural signal — reported-only telemetry + two negative results

unprovable:
# Path homology as a CFG structural signal — reported-only telemetry + two negative results

## Summary

Implements **path homology** — a directed-graph generalization of cyclomatic complexity (Huntsman 2020, [arXiv:2003.00944](https://arxiv.org/abs/2003.00944)) — for RAPTOR's binary reverse-engineering path, and tests whether it improves triage. It does not. The PR's value is an honest, instrumented evidence cycle: we built the higher-dimensional (β₂) signal, gated it on a real corpus (**β₂ is vacuous on compiled code**), pivoted to β₁ (the dimension Huntsman's grep result actually used), and ran the decisive head-to-head (**β₁ does not beat plain cyclomatic complexity**). So nothing is wired into ranking; path homology lands as opt-in, reported-only telemetry plus reusable validation tooling. The merged contribution is two clean negative results, a correct dependency-free homology implementation validated against an independent closed form, and the harness that produced the verdicts — not a triage improvement. Everything is off by default; standard runs are byte-for-byte unchanged.

Even though this is a negative result, might be of interest to @grokjc @thomasdullien - perhaps you both have some ideas for what might be nice to cherry pick and develop further? 

## Background

Cyclomatic complexity is the first Betti number of the *undirected* CFG (`|E|−|V|+c`), zero in all higher dimensions. Path homology is a homology theory for *directed* graphs whose Betti numbers `β₀,β₁,β₂,…` can be nonzero in any dimension: `β₀` = weakly-connected components, `β₁` ≈ loop count (direction-aware — it *fills* the commutative square between if/else branches, so `β₁ ≤ cyclomatic`), `β₂>0` ⇔ irreducible control flow. The hypothesis under test was that the higher-dimensional structure flags the unstructured control flow where bugs and obfuscation hide and where decompilers fail to recover structure.

## Why TDA, Betti numbers, and (eventually) persistence

**Why topological data analysis (TDA).** TDA studies the *shape* of data with algebraic topology — features invariant to continuous deformation, robust to noise, and independent of any coordinate system or arbitrary threshold. A control-flow graph is literally a graph, and its topology (how execution paths loop and interconnect) is a coordinate-free fingerprint of structural complexity that scalar metrics (LOC, nesting depth, even cyclomatic complexity) summarize lossily. The premise (Huntsman's) is that "structurally relevant" control flow -the kind that hides bugs or resists analysis — surfaces as topological features.

**What Betti numbers are.** The Betti number `β_p` is the rank of the p-th homology group — informally, the count of *p-dimensional holes* in a space, and crucially an *integer you can compute*, not a fuzzy score. `β₀` counts connected components (0-dimensional "pieces"); `β₁` counts independent loops/cycles (1-dimensional holes — the hole through a donut); `β₂` counts enclosed voids/cavities (2-dimensional holes- the air inside a hollow sphere); higher `β_p` generalize upward. They are the most basic topological invariants: two spaces with different Betti numbers are provably not the same shape.

**Why they matter for a CFG.** Mapped onto control flow: `β₀` = how many disconnected pieces a function's CFG has (normally 1); `β₁` = its loop/cycle structure (the direction-aware refinement of cyclomatic complexity); `β₂` = "tangled"/irreducible control flow that cannot be decomposed into cleanly nested loops and branches — the hypothesized fingerprint of obfuscation, hand-written assembly, and decompiler-defeating structure. The whole arc reduces to one empirical question — which Betti dimension is both *present* on real code and *predictive*? The data answers it: β₂ is real-but-vacuous (it essentially never occurs in compiled code), and β₁, while dense, does not predict better than the cyclomatic baseline it would replace.

**Why persistent homology (the principled next step — Phase 7, not in this PR).** Plain homology gives a single snapshot: one Betti vector per function. Persistent homology computes homology across a *filtration* (a growing sequence of subgraphs, e.g. by nesting depth or dominator level) and records, as a persistence diagram/barcode, when each topological feature is born and dies — separating robust large-scale structure from incidental noise and giving a multi-scale fingerprint far richer than a single Betti count. It is the natural extension and the reason Phase 7 exists in the design, but it is deliberately not built: persistence only earns its substantial complexity once the snapshot signal proves valuable, which neither β₂ nor β₁ did here.

## What's in this PR (by phase)

- **Phase 1 — homology core.** `core/inventory/path_homology.py`: `betti(graph, max_dim)` over GF(2) via bitset Gaussian elimination, dependency-free (no NumPy/SciPy/NetworkX), consuming the existing `core.inventory.dominators.Graph` protocol; `BettiVector` and a `cyclomatic_number` companion. Validated against the closed-form layered-digraph oracle `K_{n…}` (β₂=1, β₃=1, β₂=2) from the paper's eq. 8
— an external theorem, not a self-check.
- **Phase 2 — binary CFG extraction.** `packages/binary_analysis/function_cfg.py`: `BasicBlockCFG` (a `Graph`-protocol adapter), `parse_afbj` (r2 basic-block JSON → intra-function CFG), and a per-build-id on-disk cache. Wired into `radare2_understand.py` under opt-in `analyse(extract_cfgs=True)`.
- **Phase 3 — surface (reported-only).** `compute_path_homology`/`homology_report` populate `path_betti`/`cyclomatic`/`decompiler_low_confidence` on each `FunctionInfo`, surfaced in `binary-context-map.json` only when computed. Flag plumbed end to end: `--path-homology` → `raptor_fuzzing` → `orchestrator.execute` → `analyse_binary_context` → `analyse`.
- **Phase 4 — validation gate (β₂).** `core/inventory/path_homology_precision.py` + `libexec/raptor-path-homology-precision`: cross-tab, risk ratio, rule-of-three bound, non-vacuousness guard.
- **Phase 4.5 — β₁ pivot + head-to-head.** `beta1_report` and `compare_to_cyclomatic` (threshold-free AUC) added to the harness.
- **Phase 5 — reverted.** A bounded β₁ term was briefly added to the fuzz-priority score, then removed once the head-to-head showed β₁ ≤ cyclomatic. Nothing feeds ranking.

## Empirical results (the gate working — twice)

Ran on **8,459 functions** across six macOS system binaries (`grep`, `awk`, `sed`, `bash`, `perl`, `vim`), radare2 6.1.6.

- **β₂ → NO-GO.** β₂>0 occurred in **1 / 8,459** functions — compiler output is overwhelmingly reducible, so the irreducibility / `decompiler_low_confidence` signal essentially never fires. By the Universal Coefficient Theorem GF(2) Betti ≥ ℝ Betti, so this is conservative: β₂≈0 over GF(2) implies β₂=0 over ℝ. The decompiler-confidence proxy fails independently too — **68%** of *reducible* functions already contain `goto` in r2's `pdc`, so β₂>0 adds nothing over the decompiler's baseline.
- **β₁ → no improvement over cyclomatic.** β₁ is dense (54% of functions have β₁>0) and does track attack surface, but the decisive comparison is against the baseline it would replace, not against chance. Threshold-free AUC for separating dangerous-sink-reaching functions: **cyclomatic 0.707, β₁ 0.693, gap (cyclomatic−β₁) 0.700.** β₁ is 0.014 *worse* than plain cyclomatic; all three are collinear (jointly tracking size/complexity), and the gap — β₁'s distinctive direction-aware part — carries no independent signal. β₁ also costs far more to compute (r2 `afbj` + GF(2) elimination vs `|E|−|V|+c`). So the briefly-added Phase 5 β₁ ranking term was reverted.

## Honest caveats

The β₁ label is a size-confounded attack-surface proxy (the pipeline's own dangerous-sink reachability), not CVE ground truth; an AUC of ~0.70 for *all three* metrics is essentially "complexity correlates with attack surface," which the existing dangerous-sink scoring already captures. The one remaining possibility is that β₁ separates *within* size strata on labelled data where cyclomatic cannot — that is the open follow-up; on the evidence here, β₁ does not beat the baseline.

## On the GF(2) field choice (vs the paper's ℝ)

Betti numbers are computed over GF(2); Huntsman computes over ℝ. By the Universal Coefficient Theorem the two agree except under 2-torsion, which path homology of digraphs can exhibit in principle (Huntsman, fn. 5) but which is exotic and unobserved on real CFGs. The choice is conservative, not lossy, for the decisions made here: GF(2) Betti ≥ ℝ Betti always, so the β₂ NO-GO is strengthened (β₂≈0 over GF(2) ⇒ β₂=0 over ℝ); and β₁ can differ from ℝ only via 2-torsion in H₁ (H₀ is always torsion-free), which is not load-bearing for a metric that isn't wired into anything. GF(2) keeps the core exact and dependency-free; switching rank computation to a large random prime recovers the exact-ℝ values with the same algorithm, reserved for the labelled NIST SARD comparison if it is ever run.

## Usage

`raptor-fuzz --orchestrator --path-homology …` computes the per-function signal during binary analysis and writes `path_betti` / `cyclomatic` into `binary-context-map.json` for inspection. It is reported-only: it does **not** affect prioritisation. Off by default; adds one r2 `afbj` call per function (cached per build-id). The validation harness is `libexec/raptor-path-homology-precision <manifest.json>` where the manifest is `{"binaries":[{"path":..,"vulnerable_functions":[..]}]}`.

## Testing

~70 unit tests, all r2-free (so they run without radare2 installed): homology core (27, incl. the layered-digraph oracle and the if/else `β₁<cyclomatic` divergence), CFG parser/adapter/cache (16), precision harness incl. the β₁-vs-cyclomatic AUC head-to-head (19), and binary surfacing (8). The β₂ NO-GO and the AUC numbers are from real radare2 runs (radare2 6.1.6). Existing radare2 suite unchanged (33 pass). `ruff` (F401/F811/F821/F841) clean.

## Soundness boundary

Path homology is a *metric only*. It is deliberately kept out of the sanitizer-cut chokepoint and never gates a suppression — that path requires sound reachability, which a topological complexity metric is not.

## Follow-up (open)

Size-controlled β₁ validation on NIST SARD ground truth: stratify β₁ by cyclomatic size on real vuln/fixed pairs and test whether β₁ separates *within* strata where cyclomatic cannot. If it does → reconsider wiring β₁ and revisit Phases 6–7 (source-level homology, persistent path homology, both on hold). If it does not → the arc closes with cyclomatic as the metric of record. Corpus build is operator-run (compilation, not CI).

## Design doc

`docs/design-path-homology-cfg.md` carries the full rationale, the per-phase status table, both result sections, and the GF(2) note.

This version is fully consistent with the reverted state: reframed as reported-only + two negatives, Phase 5 marked reverted, the AUC head-to-head table added (cyclomatic 0.707 / β₁ 0.693 / gap 0.700), caveats sharpened, GF(2) note retained, TDA/Betti/persistence section kept, test counts updated (~70 new). Lines are unwrapped within paragraphs per your "newline-free" preference.

One judgment call to flag: I kept the title as a negative-result framing ("reported-only telemetry + two negative results") — if you'd rather the PR read as a capability addition, I can retitle, but the honest framing is the stronger submission and matches what's actuallymerged.


grokjc:
@unprovable If you take a look at #864 you'll see a potential consumer for some of this. 

Right now _FunctionInfo_ carries name/address/size/reachability but zero per-function complexity metrics, and there's no
intra-function CFG at all - `radare2_understand.py` only uses the inter-function call graph (_aflcj_). 

Your `function_cfg.py` fills that gap: a _BasicBlockCFG_ implementing the existing Graph protocol from `core/inventory/dominators.py`, with the r2 _afbj_ parser and per-build-id cache. And _cyclomatic_number_ from `path_homology.py` gives us the obvious first metric to surface on _FunctionInfo_.

@danielcuthbert what do you think? 

If this works we'd want to carve out a separate PR I think?

grokjc:
Right, I keep coming back to this one. I think I have a plan. What we don't need is to merge this as-is: the code is great but gives 'the wrong result' by your own admission. But there is a great use case.

**What we do want — and you called this in the PR body — is a carve-out.** You offered cherry-pick suggestions; here's the concrete scope, which matches what was flagged back in July re #864:

1. **`packages/binary_analysis/function_cfg.py` + `tests/test_function_cfg.py`, essentially as-is.** `BasicBlockCFG` on the existing `Graph` protocol, the defensive `afbj` parser, and the build-id cache with the collision guard fill a real gap — we have no intra-function CFG anywhere today — and the `/binary` and dominator machinery are live consumers.
2. **`cyclomatic_number`, extracted into its own small module** (e.g. `core/inventory/cfg_metrics.py`) with its tests. It's the metric that won your head-to-head, so it's the one that earns a place — surfaced as a `cyclomatic` field on `FunctionInfo` behind the `extract_cfgs=True` wiring.
3. Optionally the tie-aware `_auc` helper — it's a clean dependency-free primitive other precision harnesses could reuse.

Two fixes to fold into the carve-out:

- **Self-loop undercount.** `parse_afbj` drops self-edges at extraction, so a single-block spin loop reports cyclomatic 0 where McCabe says 1. Keep self-edges in the extracted CFG and let the homology normalisation drop them (it already does) — loopless-ness is path homology's constraint, not the CFG's.
- **No design doc in `docs/`** — that directory is end-user documentation only, and the doc also references `design-aggregation-dominators-wp.md`, which doesn't exist in this repo.

Rebase warnings, since the drift bites in two specific places: `radare2_understand.py` now imports only `Any` from typing, so `Optional`/`Dict`/`List` annotations will NameError — use `X | None` and builtin generics throughout. And `raptor_fuzzing.py` replaced `Path(args.dict) if args.dict else None` with `_resolve_dict_path(args, out_dir)`; taking your side of that hunk would silently regress dict resolution.

**For the record, so the results survive the branch** — 8,459 functions across six macOS system binaries (grep, awk, sed, bash, perl, vim), radare2 6.1.6:

| finding | result |
|---|---|
| β₂ > 0 occurrence | 1 / 8,459 — vacuous, hard NO-GO |
| goto in `pdc` among *reducible* functions | 68% — invalidates the decompiler-confidence proxy independently |
| AUC vs dangerous-sink proxy | cyclomatic **0.707** · β₁ 0.693 · gap 0.700 |

Once the carve-out lands, we'll close this PR unmerged with a pointer back here. The branch stays retrievable, and if the size-stratified NIST SARD follow-up ever shows β₁ separating within strata where cyclomatic can't, everything here is self-contained, and reviving it is straightforward.

unprovable:
How did this carving go? If it's done, we can close this PR :)

grokjc:
> How did this carving go? If it's done, we can close this PR :)

I was hoping you'd do it 😄

grokjc:
Carving done! Thanks for all your efforts on this. Closing this PR now as suggested.

PR #864, opened 2026-07-06
Add black-box binary understanding

danielcuthbert:
A monster I tell ya! the result of nearly a month worth of work to make our 'what in gods name is this $blob$ and is it good or bad? 

Lots of fun here, stuff like an evidence-backed binary workflow so RAPTOR can map compiled targets, rank ingress points, retain graph memory and hand off to runtime tracing, harness planning and fuzzing without pretending heuristics are proof.

We also have a new /binary surface keeps the existing /understand flow working, but gives unknown binaries a proper mechanical path instead of a thin radare2 pre-pass.

Screenshot 2026-07-06 at 09 02 26

and it works pretty damn well too!


grokjc:
This is awesome. Going to land some patches on this branch before proper review - no point describing gaps that I can fix trivially and faster 😄

grokjc:
We will definitely need to rebase this before merge or we run the risk of silently dropping a few important things

grokjc:
OK. I've pushed two rounds of fix-up commits onto your branch (17 commits total) covering connection lifecycle, path resolution, env scoping, payload deduplication, truncation logging, graph-query validation, and test coverage for 6 previously untested modules. I also rebased onto main so the SCA calibration data won't silently revert on squash-merge.

I did the sandboxing on my end but couldn't run the full integration path (no r2/frida wired up on that box) — could you pull the branch and give it an end-to-end run on a real binary to make sure nothing broke?

packages/binary_analysis/pipeline.py
```diff
@@ -0,0 +1,2426 @@
...
+        "unchecked_flows": [],
+        "runtime_signals": [signal.to_dict() for signal in manifest.runtime_signals],
+        "runtime_observations": _runtime_observation_summary(
+            runtime_events,
+            [record.id for record in runtime_records],
+        ),
+        "runtime_input_flows": runtime_input_flows,
+        # Patched by the caller after external_ingress recovery (depends on
+        # this context_map, so can't be computed here).
+        "runtime_parser_flows": [],
```
grokjc, line 1045: _context_map returns placeholder empty lists for runtime_parser_flows, call_graph_edges, and parser_boundary_candidates that the caller mutates after construction. Between construction and patching, the context_map is already passed to recover_external_ingress and other functions that receive it with those fields empty. Works today because those functions don't read the empty fields, but it's a latent ordering dependency — if any of those functions starts reading one of the placeholder fields, it silently gets stale data. Consider computing the sub-results before calling _context_map, or having _context_map
  accept them as parameters.

packages/binary_analysis/cli.py
```diff
@@ -0,0 +1,732 @@
...
+        print("raptor-binary: --fuzz-dir cannot be combined with --fuzz/--active", file=sys.stderr)
+        return 2
+
+    start_run(out_dir, "understand", target=str(target))
+    active_phases: list[dict[str, Any]] = []
+    try:
+        # Always map first. Agentic follow-on work needs a mechanical view of
+        # the artefact before it decides whether runtime or fuzzing is sensible.
+        result = _analyse_for_args(args, target, out_dir)
+        if args.runtime or args.active:
```
grokjc, line 369: _run_investigate mutates the parsed args namespace as a side-effect channel (args.runtime_dir = str(runtime_dir)) then re-calls _analyse_for_args. If analysis ever caches on the namespace, or future code reads args.runtime_dir expecting the operator's original value, this silently gives the wrong answer. Consider passing runtime_dir/fuzz_dir as explicit keyword overrides instead.

packages/binary_analysis/harness.py
```diff
@@ -0,0 +1,636 @@
...
+    "document_type",
+}
+
+
+def _slug(value: str) -> str:
+    clean = re.sub(r"[^A-Za-z0-9_.-]+", "_", str(value or "")).strip("._")
+    return clean[:80] or "harness"
+
+
+def _c_string(value: str) -> str:
```
grokjc, line 54: _c_string uses json.dumps as a C string literal escaper. JSON encodes non-ASCII as \uXXXX which is only valid in C11+. If a binary path contains non-ASCII characters, the generated harness won't compile on pre-C11 compilers. Worth deciding whether pre-C11 support matters here — if so, use \xHH sequences instead.

grokjc (commented):
~~Several test files use SimpleNamespace stand-ins instead of the real production class shapes (test_cli.py lines 91-105, test_ingress.py lines 19-26). If the production classes gain or rename fields, these tests pass while production breaks. Also, test_radare2_understand.py and test_target_detector.py use NamedTemporaryFile(delete=False) with manual unlink instead of pytest's tmp_path — fragile on test failure. And test_blackbox_pipeline.py has no tests for corrupt/truncated binary input. Not blockers, but worth hardening over time.~~

Fixed

grokjc:
If your E2E test works and the sandboxing hasn't broken anything, this is definitely ready to merge now 😄

danielcuthbert:
Graph reuse was crashing as it crashed when there was an existing graph (my bad) so fixing that now

grokjc:
That bug was lurking elsewhere too, have extracted your code into a shared helper.

PR #890, opened 2026-07-23
fix(sage): switch MCP transport from SSE to stdio (#888)

### LINKED ISSUE #888: Issue w/SAGE 11.11.6
`core/sage/mcp-entry.json` registers SAGE using an invalid SSE config (`type` and `url` are not valid MCP fields), pointing at `/mcp/sse` which returns 404 in SAGE 11.11.6.

  ### Fix

  Replace `.mcp.json` (and `core/sage/mcp-entry.json`) with stdio transport via `docker exec`:

  ```json
  {
    "mcpServers": {
      "sage": {
        "command": "docker",
        "args": ["exec", "-i", "sage-sage-1", "/usr/local/bin/sage-gui", "mcp"],
        "env": {
          "SAGE_PROVIDER": "claude-code"
        }
      }
    }
  }
  ```

  Restart Claude Code after updating.

Paging @l33tdawg: what do we need to do to fix this properly in RAPTOR's scripts/setup please? 😄

grokjc:
The SSE config at /mcp/sse has never worked — written in PR #192 for SAGE 6.6.5, never updated across three version bumps. The correct endpoint /v1/mcp/sse requires OAuth (auth code + PKCE), which is impractical for a localhost Docker sidecar on headless dev boxes.

Switch to stdio transport per SAGE maintainer recommendation:

- Add libexec/raptor-sage-mcp wrapper that exec's into the container via docker compose exec with the required SAGE env vars. Re-execs under sg docker when the user lacks direct docker socket access.
- Rewrite raptor-sage-setup: generate the MCP entry inline (replacing .mcpServers.sage entirely to clear stale SSE fields), add volume migration from /root/.sage-gui to /root/.sage, add MCP smoke test.
- Fix compose volume mount: sage_data:/root/.sage with explicit SAGE_HOME env var.
- Delete core/sage/mcp-entry.json (no longer used).
- Update SAGE_INTEGRATION.md for stdio transport and migration.

The agent key at /root/.sage/agents/raptor-claude-code/agent.key is created automatically by sage-gui mcp on first launch and registered on-chain by the first sage_inception call.

Tested E2E: migration, compose up, health, seed, register, MCP initialize + tools/list all pass against SAGE 11.11.6.

Closes #888


grokjc:
docker-compose.yml:
  - ollama-init: service_completed_successfully — sage waits for model pull to finish
  - Replace OLLAMA_URL with the canonical SAGE_EMBEDDING_PROVIDER, SAGE_EMBEDDING_BASE_URL, SAGE_EMBEDDING_MODEL
  
  raptor-sage-setup:
  - check_docker() — validates docker compose plugin + daemon access, auto re-execs via sg docker if needed
  - check_compose_file() — pre-flight compose file existence check
  - check_port() — port conflict detection, idempotent skip when own containers hold the port
  - verify_embeddings() — confirms ollama embeddings active via printenv + API (no log parsing — SAGE's zerolog ANSI codes break grep)
  - smoke_test_mcp — timeout → gtimeout fallback for macOS
  - install_sage — script existence checks for seed/register, proper ordering (verify → seed → register → mcp → env → smoke)
  - --status subcommand — reports health, embedding mode, ollama model status, memory count, MCP entry presence
  - _ORIG_ARGS saved for sg docker re-exec

l33tdawg (changes requested):
The stdio wrapper/config-generation direction matches the #888 guidance, but the migration still has a data-loss path.

`volume_name` is the existing `sage_data` volume that was mounted at `/root/.sage-gui`. If it contains any legacy data while the real SAGE 11.x state lives in the old container's writable `/root/.sage`, lines 301–323 see the non-empty volume and skip the copy. The script then deletes `$backup` at line 326 and `docker compose up` recreates the container with that same volume mounted on `/root/.sage`, so the writable-layer ledger/key state is lost.

A non-empty destination must not be treated as a successful migration; we should fail closed before recreation and retain the backup/old container for recovery, or implement a verified conflict-aware merge. Add a regression case where both the old volume and `/root/.sage` contain state, and verify the chain identity, data, and agent key before discarding the backup.

grokjc:
Have made changes

l33tdawg:
Thanks — the stdio/Ollama work and preserving the backup look good.

One small thing I would tighten before shipping: if the old volume and the writable `/root/.sage` both contain the same SAGE_HOME files, the current `tar -k` path silently chooses the existing one. Could we detect that collision (or stop safely) and add a small dual-state regression check for the chain, key, and data? That would make me much more comfortable with the migration path.

grokjc:
7a311da — collision detection + post-migration regression check

  When the destination volume already has data during migration:
  - Compares each backup file against its destination counterpart via cksum and reports collisions (differing content). Destination version kept; backup preserved for manual comparison.
  - Post-migration regression check verifies chain database, agent keys, and data directory exist. Warns if chain is missing (SAGE will init a fresh one).
  - Uses find -exec sh -c '...' _ {} + to handle filenames with spaces safely.
  - Regression check only fires when migration actually ran — idempotent re-runs unaffected.

  e5e3a99 — macOS portability
  
  - mktemp -p DIR TEMPLATE → mktemp "DIR/TEMPLATE" (4 sites) — BSD mktemp has no -p flag.
  - grep -oP 'Memories:\s*\K\d+' → sed -n 's/.../\1/p' — BSD grep has no -P (Perl regex).

  Both forms are POSIX-portable across GNU coreutils, BusyBox, and macOS.

l33tdawg:
Hey bro - the macOS portability fixes look good, and the updated script parses cleanly.

One migration blocker remains. The post-check searches for chain.db, but SAGE 6.6.5 and 11.11.6 store state under data/cometbft/, data/badger/, and data/sage.db; there is no chain.db. It also checks only agents/*.key, not the node operator root agent.key.

The migration still converts copy, comparison, extraction, and validation failures into success, then writes .migration-complete and recreates the container. Could we fail closed before that point, and abort on collisions involving critical state (agent.key, vault.key, config.yaml, data/sage.db, data/badger/**, or data/cometbft/**) while preserving both versions for the operator?

For non-critical comparisons, cmp -s or the complete cksum output would be preferable to comparing only the CRC field.

grokjc:
Feels like this is getting closer to working 😄

l33tdawg:
Yep, definitely closer 😄 The critical-collision handling looks right now. One last gap: copy, comparison, merge, and validation failures are still swallowed (`|| true` / warning-only checks), so `.migration-complete` can still be written after an incomplete migration. Could those paths return non-zero before the marker? Then I think we are there.

l33tdawg:
I pushed the remaining fail-closed changes and regression coverage directly here: https://github.com/grokjc/raptor/pull/2. It preserves recovery state and stops before recreation on every migration failure path. Feel free to merge that PR or cherry-pick `d0304db5` 😄

grokjc:
Thanks for PR. I noted a subtlety with old-vs-new SAGE versions (and which parts are essentially optional vs mandatory) - hopefully resolved in 09d24c5

l33tdawg:
Thanks — I found one lifecycle issue in 09d24c5: the new docker exec probe runs after compose stop, so real Docker always rejects it and writable 11.x state is treated as absent. I fixed it in https://github.com/grokjc/raptor/pull/3. The probe now runs before stop with explicit present/absent results and fails closed on every other outcome; the regression tests cover present, absent, failed-probe, post-stop exec rejection, and writable-copy failure. All 126 SAGE tests pass locally.

grokjc:
OK. I'm happy if you're happy!

grokjc (commented):
I think we're good - you're still showing as a pending reviewer

l33tdawg (approved):
All SAGE integration concerns are resolved. The fail-closed migration changes and pre-stop writable-state probe are merged, regression coverage is in place, and the current CI is green. Approved — thanks for working through this with us!

grokjc:
Thanks for your help as always @l33tdawg - I have a bunch of new SAGE integrations on main and this will unlock them all for everyone. Cheers 🍻

l33tdawg:
Anytime brother - always happy to help :) (and awesome to see that SAGE integration is being used 👍 )

PR #892, opened 2026-07-27
fix(audit): make G2 structural, unbreak Mode-2 sweep, hydrate gap source

dinosn:
# Strengthen three load-bearing paths in `/audit`

## Context

The `/audit` landing is a genuinely new capability for RAPTOR, and the design idea behind it is the right one. Everything else in the stack — `/scan`, `/agentic`, `/codeql` — starts from what a scanner already flagged and triages downward. `/audit` inverts that: it starts from the functions **no tool reported on**, forms a hypothesis per function, and tries to ground that hypothesis mechanically. That inversion is the part you cannot get by tuning a scanner, and the surrounding machinery — the coverage-gap denominator, the per-function annotation trail, the strategy exemplars, Mode-2 checker synthesis, the Reflexion feedback loop back from `/validate` — is a lot of well-organised infrastructure to support it.

While reading through it in depth (and running three independent adversarial reviews over it), three specific paths came up where the implementation doesn't yet deliver what the design intends. All three are the kind of thing that only shows up when you trace an end-to-end run: each is a few lines, each sits under a large amount of correct code, and each quietly costs the subsystem something it was built to provide.

This PR fixes those three. It doesn't change the architecture or the operator surface — it makes three existing mechanisms actually do what their own docstrings and `SKILL.md` already promise.

---

## 1. Make the G2 tool-grounding gate structural

**Why this improves things:** G2 is the load-bearing claim of the whole subsystem. `SKILL.md` states it plainly — *"The LLM never directly classifies code as vulnerable... Ungrounded findings are annotation-only (`suspicious`), never `finding`."* That promise is what lets a downstream consumer treat an `/audit` finding as different in kind from an LLM opinion. This change makes the code able to keep it.

The gap was that a receipt and a claim arrived as the same string. `REVIEW_SCHEMA` asks the model to name its supporting tool (`llm_review.py:103`), and that answer landed in the same `evidence_tool` field the orchestrator stamps after a real tool run. Two consequences followed:

- A model answering `evidence_tool: "prefilter"` satisfied `_is_tool_confirmed`, so `_sweep_validate` short-circuited at `orchestrator.py:4064` and **the tool chain never ran** — precisely the validation G2 exists to force.
- A model answering `"joern"` passed G2 (the check was a blacklist of sentinels, so any unlisted string counted as grounded) and `evidence_grade.py:278` then exported it as HIGH confidence, *"confirmed by Joern CPG analysis"*. On a host with no Joern installed.

The fix makes the distinction structural rather than lexical:

- `core/audit/_util.py` now owns the contract. A **receipt** is `<namespace>:<detail>` from a registered namespace, optionally `+`-joined, or one of the bare tokens only a verified run sets (`dynamic`, `frida`, `lifecycle`, `dark_verify`). Everything the model supplies is namespaced `llm-claimed:` at the `llm_review` boundary — including inside `review_result`, since G2, `_sweep_validate` and `_proactive_validate` all read the nested dict in preference to the outcome field. The model's original wording is kept as `evidence_tool_claim` so the annotation trail doesn't lose it.
- `_is_ungrounded_evidence` is now stated positively — G2 passes only for a value `_stamp_evidence` could have issued. The old blacklist let every unenumerated string through.
- `_stamp_evidence` warns if it is ever handed a label that isn't a well-formed receipt, so adding a new stamp without registering its namespace is loud instead of silently downgrading real evidence.

**Two things this also fixes in the positive direction**, which matter as much as the negative:

- `evidence_grade.py` now grades by namespace across receipt components. Previously it matched *bare* tool names only — so a genuine `joern:live` or `smt:check-oob` receipt graded as **nothing**, while the model's bare `"joern"` graded HIGH. That was exactly backwards. Real evidence now grades, and a composite receipt grades each component that is itself a valid receipt — so `smt:check-oob+joern` grades SMT only, never the unstamped tail.
- A G2 demotion is now **recoverable**. `_promote_suspicious` skipped anything demoted by a gate, which would have made stricter grounding a one-way door. G2 means "no receipt *yet*" — running a tool later and getting one is the resolution — so a demotion citing only G2 stays eligible for post-loop promotion (`_is_evidence_only_gate_demotion`). Every other gate (no hypothesis, empty body, language/CWE mismatch) is a reasoning defect a tool hit shouldn't override, and those stay authoritative as before.

**Deliberate behaviour change:** findings whose only support is a model-named tool now land as `suspicious` rather than `finding`. That is what G2 was written to do. Three existing tests encoded the old behaviour and are updated here — `test_g2_joern_passes` is now split into `test_g2_joern_receipt_passes` / `test_g2_bare_joern_fails` / `test_g2_arbitrary_string_fails`, and the five `test_evidence_tool_*` grading tests now pass real receipts.

## 2. Stop a successful Mode-2 sweep from aborting the run

**Why this improves things:** checker synthesis is one of the best ideas in the subsystem — confirm a bug once, generalise it to a rule, sweep the whole codebase. Right now that payoff is unreachable, because the failure fires on *success*.

Synthesis hits are sites, carrying `function` (`checker_synthesis.py:204`). `TaskGraph.from_workqueue` dereferences `gap['name']` unguarded (`task_graph.py:197`), and the second-pass graph construction sits outside the synthesis `try/except` — `run_orchestrator` has only a `finally` for Joern cleanup. So a confirmed finding whose synthesized rule matched in another file raised `KeyError: 'name'` **after the entire review loop had run**, before state sync-back and before `collector.flush()` — losing the run's findings and its buffered audit log.

`_synthesis_hits_to_gaps` resolves each site to its enclosing checklist function via a new `gaps.gap_for_site`, which mirrors the resolution already used by `understand_bridge._find_containing_function`: strict smallest-containing-span (so a nested function or closure is attributed to the innermost match, not the enclosing one), then a closest-preceding-`line_start` fallback, bounded by the next definition, for inventories that record only the definition line. A trailing definition with no next start has no derivable bound, so those sites stay unresolved rather than being given a fabricated span that would truncate review.

Two details worth flagging for review:

- Sites that resolve to no reviewable function (a top-level initializer, a macro, generated code) can't become review tasks, but they're still tool signal — they now go to `unresolved-synthesis-hits.json` rather than being dropped.
- Functions reviewed in an earlier round or run are deliberately **not** skipped. A synthesized checker encodes a pattern nobody knew about when those functions were first reviewed, which is exactly when re-review pays.

## 3. Populate `gap["source"]` so the mechanical detectors run

**Why this improves things:** negative-space analysis and sibling-asymmetry detection are two of the most distinctive things `/audit` brings — "what check does this function omit that its siblings all perform" is a bug class scanners structurally cannot reach. Both are wired into the default path and both are currently no-ops.

`discover_conventions` and `check_negative_space` read `gap["source"]` and skip any gap without it (`negative_space.py:288`, `:359`); sibling peer groups are built from `g.get("source", "")` (`orchestrator.py:1435`). `compute_gaps` builds gaps from the checklist, which carries line spans but no text — so no gap ever had a `source` key, `conventions` came back empty, and `check_sibling_negative_space` was never even called (it's guarded on `if conventions`).

`hydrate_gap_source` fills it in, at the point where the gap list is final — after the stale merge, scoring and budget truncation — so stale gaps are covered too and no source is read for gaps the budget dropped. It is bounded in a way that holds on a hostile or very large tree: gaps are grouped by file so each file's line list is released after use rather than cached for the whole run; files are size-capped and NUL-probed before reading (a minified bundle can satisfy a line-count bound and still be hundreds of MB); per-function and running-total byte caps are checked *before* storing, counting UTF-8 bytes. Path containment goes through `safe_join`, since checklist paths come from the scanned tree.

`write_gaps` strips `source` before serialising — otherwise `gaps.json` becomes a copy of the target tree in the run directory. No consumer reads it back (checked `raptor-audit gaps`, `raptor-review gaps`, `priority.score_functions`, `report.py`, the validate bridge).

---

## Testing

- **99 new tests**, covering each patch and the specific failure each prevents. `test_synthesis_hits.py` asserts `pytest.raises(KeyError)` on the raw hits before showing the resolved gaps work, so it fails loudly if the resolution is removed. `test_gaps.py::TestGapSourceHydration::test_negative_space_sees_hydrated_gaps` asserts `discover_conventions` returns `[]` on unhydrated gaps and non-`[]` on hydrated ones — the end the change serves. `test_llm_review.py::TestEvidenceProvenanceThroughFactory` drives the real `make_review_fn` with 15 parametrized model claims rather than mirroring its logic.
- **Blast radius** (`core/audit`, `core/analysis`, `core/concepts`, `core/orchestration`, `packages/checker_synthesis`, `packages/llm_analysis`): **18 failures before and after, the same 18**; passed `6043 → 6142`, the difference being exactly the new tests. Baseline measured in a clean worktree at `main`. The 18 are environment-dependent (z3/SMT, a compiler-driven binary-oracle fixture) and unrelated.
- **Full suite** (`core/ packages/`): 111 failures before and after (spatch, frida, npm/pip/go, compilers).
- `ruff check core/audit/` clean.

## Reviewed against

Three adversarial review rounds by an independent model, each of which found something real:

- **Round 1** caught that the first cut of patch 1 was only half-effective — the sanitizer set `outcome.evidence_tool` but left the raw model string in `review_result`, which G2 and `_sweep_validate` read in preference. It also found two receipt labels (`lifecycle`, `dark_verify`) I hadn't registered, which would have caused false-negative demotions of genuine evidence, and that a G2 demotion was unrecoverable because `_promote_suspicious` skips gate-demoted outcomes.
- **Round 2** caught that my `reviewed_set` fix didn't reach the real filter (it lives upstream in `checker_synthesis.py`, keyed on `<matched file>:<seed function>`, which also discarded every same-file sibling); that the hydration ceiling wasn't a ceiling; and — importantly — that splitting receipts on every `+` would reject a real Semgrep receipt whose `rule_id` contains a `+`. That last one is a false negative, the more dangerous direction here.
- **Round 3** caught that the validator and the grader had drifted apart: the validator anchors on the leading component, so `smt:check-oob+joern` validated, but the grader split every `+` and graded Joern too. It also caught that seed-site exclusion by line was wrong, since `outcome.line` is the function's `line_start` rather than the vulnerable line (and is `0` when the inventory recorded none) — exclusion is now by resolved function identity.

Round 3's verdict on the current tree: net-positive versus `main`, and the deferred finalizer below is explicitly **not** a regression risk, since those bypass paths already exist on `main` and sanitizing at the LLM boundary makes their grading better rather than worse.

## Known follow-ups, deliberately not in this PR

Kept out to keep this reviewable; happy to pick any of them up next:

- **Centralising LLM ingress.** The live-sink requeue, flow-trace, error-retry and trivial-batch paths reach `_tally_outcome` without passing the full gate chain. One `finalize_outcome()` chokepoint would make the guarantee uniform instead of per-path.
- **CLI receipts.** `raptor-audit record --status finding` accepts an `--evidence-tool` name without a matching sweep. Consuming a sweep-issued receipt ID would close it.
- **Remaining any-non-`llm`-string consumers**: `refinement.py:60`/`:126`, mid-loop synthesis (`orchestrator.py:943`), session observations (`orchestrator.py:3406`).
- **`dynamic_sweep.py`** treats any non-zero harness exit as `confirmed`, which conflates a bare crash with sanitizer-class evidence.
- **Docs**: `4f273a81` added the `/audit` and `/review` sections to `docs/commands.md` — thanks, that closes most of this. Two small leftovers: CLAUDE.md's COMMANDS list still doesn't mention either command, and `docs/commands.md:179` labels the sandbox syscall-`--audit` flag as "Enable systematic code review layer", which is a different flag entirely (it's the seatbelt/strace audit log) and predates all of this.

Happy to split this into three PRs if that's easier to review.


grokjc:
I haven't finished landing all the fixes from my machine yet so there are conflicts here 😄 but I appreciate the support!

dinosn:
No problem, disregard ! :) I will check back when things are complete

grokjc:
This definitely needs a rebase as this branched before 03db3922 and #893 landed. Most of the 94-file diff is base divergence; rebase will collapse it significantly I hope.

  Three things to check post-rebase:

  1. llm_review.py ~L602 — empty response fallback

  The else branch that returned {"status": "error", "body": "empty LLM response"} was dropped. An empty response list now hits response[0] → IndexError. The caller catches Exception so it won't crash the run, but the structured error context is lost. Might be worth keeping the guard.

  2. packages/joern/runner.py — inner try/except in fallback path (build_cpg ~L228, run_query ~L389)
  
  The inner try/except for TimeoutExpired/OSError inside the except TypeError: block was removed. Sibling except clauses don't catch exceptions from each other, so those exceptions now propagate to the caller instead of returning a graceful JoernCPG/JoernResult. Probably unintentional? The outer handlers look like they'd cover it but they don't!

  3. packages/joern/runner.py ~L725 — _parse_output flow parsing

  if not in_flows: continue → if not in_flows: in_flows = True. This changes behaviour: JOERN_FLOW lines outside START/END sentinels are now parsed instead of skipped. If that's intentional (e.g. handling Joern versions that omit sentinels), worth a comment? What was the intention here?

  The core provenance work looks really good: receipt/claim distinction, namespace validation, gap hydration bounds, synthesis hit resolution all look solid and internally consistent. This is great work, thank you 🙏

grokjc:
Three edits to fix the CI failure on test_evidence_tool_dynamic_crash:

  1. Add "dynamic" to TOOL_EVIDENCE_NAMESPACES — "dynamic" was in BARE_TOOL_EVIDENCE but not TOOL_EVIDENCE_NAMESPACES, so is_stamped_tool_evidence("dynamic:crash") failed at the entry guard. The namespaced form never reached _grade_tool_receipt.
  2. Add "dynamic:crash" to _RECEIPT_GRADES — without this, the namespace lookup resolves to "dynamic" → DYNAMIC_SANITIZER (HIGH), which is the bug this whole fix addresses. The new entry maps to DYNAMIC_CRASH
  (MEDIUM).
  3. Try full stamp before namespace in _grade_tool_receipt — _RECEIPT_GRADES.get(part) or _RECEIPT_GRADES.get(namespace). Without this, _receipt_namespace("dynamic:crash") returns "dynamic" and the dict entry from (2) is never hit.

  All three are needed — each alone still fails. The root cause is that "dynamic:crash" is a qualifier within the dynamic namespace, not a separate tool, so it needs special handling at every layer: validation, lookup, and grading.

grokjc:
I've landed fixes for 3/4 of the items from your follow-up list (co-authored), and tweaked the PR branch so CI passes — we needed to add "dynamic" to TOOL_EVIDENCE_NAMESPACES, add a "dynamic:crash" entry to
  _RECEIPT_GRADES, and try the full stamp before namespace fallback in _grade_tool_receipt.

  The first fix(centralising LLM ingress) is deferred as I think it will be easier post-rebase - and the consumer-side fixes close the exploitable gaps for now anyway.

  Apart from that this PR needs a rebase and then if you could take a look at the empty response guard in llm_review.py ~L602 and the inner try/except removal in packages/joern/runner.py we can close the last few gaps out. Thanks!

dinosn:
Rebased — `d930200b` merges current `main` (`ba4a85ef`) in cleanly, no conflicts, and the branch is now zero commits behind. The diff has collapsed to the 13 files it was always meant to be (+1670 / −102). Thanks for landing the `dynamic:crash` chain and the follow-ups; the CI diagnosis in your earlier comment was exactly right, and all three layers were indeed needed.

## Your three items — all three were base divergence, nothing was dropped

I checked each against the tree rather than the diff, and all three are artifacts of the pre-`03db3922` base rather than changes on this branch:

1. **`llm_review.py` empty-response guard.** Still present — `else: result = {"status": "error", "body": "empty LLM response"}` at `core/audit/llm_review.py:607`, identical to `main`. The `response[0]` → `IndexError` path you were worried about can't be reached.

2. **`packages/joern/runner.py` inner `try/except`.** Still present in both fallback paths — `build_cpg` at `:229-241` and `run_query` at `:390-403`. Your reasoning about sibling `except` clauses not catching each other was correct, which is what made this worth checking; the guards just never left. The decisive check is that `git diff HEAD origin/main -- packages/joern/runner.py` is **empty** — the file is byte-identical to `main`, so this branch contributes no delta to it at all.

3. **`_parse_output` flow parsing.** Unchanged: still `if not in_flows: continue`, so `JOERN_FLOW` lines outside the START/END sentinels are skipped exactly as before. No intent to change behaviour here — the diff was showing you `main`'s own content.

So no code changes for these, and the same `git diff … -- packages/joern/runner.py` check covers items 2 and 3 together.

## CI

Green, and I confirmed the new tests actually execute rather than being filtered out of the batch: 569 + 593 passed, 61 skipped, 0 failed, including `TestEvidenceProvenanceThroughFactory` and `TestGapSourceHydrationBounds`.

## Follow-up status — one is only half-closed

`dynamic_sweep`, the remaining non-`llm`-string consumers, and both doc leftovers are genuinely closed. **CLI receipts is not**, and I think it got counted as landed when it's only partly there.

`choices=sorted(VALID_EVIDENCE_TOOLS)` at `libexec/raptor-audit:1110` does stop free text, which is a real improvement. But two gaps remain:

- `--status finding --evidence-tool joern` still records a finding with no sweep having run. The gate checks presence, not that a tool executed.
- More awkwardly, the CLI's permitted vocabulary and the receipt contract have diverged. Six of the ten allowed choices — `coccinelle`, `codeql`, `compilation`, `joern`, `semgrep`, `smt` — are bare names, and bare names are exactly what `is_stamped_tool_evidence` rejects. Only `dynamic`, `dynamic:crash`, `frida` and `dark_verify` overlap with `BARE_TOOL_EVIDENCE`. So a CLI-recorded finding now grades as *nothing* where it previously graded HIGH:

  ```
  grade_review_result({"status": "finding"}, "joern")       -> []          # CLI can emit this
  grade_review_result({"status": "finding"}, "joern:live")  -> joern/HIGH  # only the sweep can
  ```

  That's a consequence of this PR meeting your CLI change, not of either one alone. The fix is to make the CLI consume a sweep-issued receipt rather than a tool name, as originally scoped.

## Three things that came out of reviewing the merged tree

**1. There are now two evidence validators, and they disagree.** `is_stamped_tool_evidence` (`core/audit/_util.py:77`) drives G2, `_is_tool_confirmed`, `_stamp_evidence` and the grader. `is_tool_evidence` (`core/audit/evidence_grade.py:92`) drives `refinement.py:60` and `:127`. On 16 sampled values they disagree on 10, always in the same direction — lax says grounded where strict says not: bare `joern`, `semgrep`, `codeql`, `smt`, `prefilter`, `triage`, `sweep`, `critique`, `sarif_cache`, and `critique:madeup:x`.

I traced it and it's latent on the stock path: everything reaching refinement has already passed the `llm-claimed:` sanitiser, and the two remaining bare-stamp sites (`triage`, `prefilter`) are terminal `clean` outcomes that neither gate acts on. It is reachable through the programmatic `review_fn` boundary, where an alternate reviewer can return a bare `joern`, though there the effect is to skip a useful refinement round rather than to bypass G2.

Worth fixing in this PR regardless — it's the predicate the whole guarantee rests on, and it's the same drift class round 3 already caught between the validator and the grader. Suggest deleting the body of `is_tool_evidence` and repointing both refinement call sites at the canonical parser. If refinement actually wants "was a tool attempted", that's `tools_dispatched` — a different question that deserves its own name.

Worth naming the cause, because it'll recur: this happened because we both repaired the same semantic contract in different modules while the PR was open. Git merged the text without a conflict and the invariant forked silently. Not really avoidable by review — only by sequencing (rebase first, then follow-ups on the branch).

**2. The `+` grammar is ambiguous, and the validator and grader read it differently.** The validator anchors on the leading component and treats the tail as opaque; the grader splits on every `+` (`evidence_grade.py:394`). So a genuine leading receipt can carry a tail that gets graded on its own:

```
"semgrep:corp.rule+dynamic"  -> validator: True
                             -> grader:    mechanical:semgrep/HIGH + dynamic:sanitizer/HIGH
```

That second one is fabricated sanitiser evidence for a run that never happened. Same for `+frida` and `+dark_verify`.

**It is not reachable today** — I checked all seven `confirmed.append` sites and every component is a constant, a catalogue stem, or a tempfile path, none of which can contain a `+` (Python's tempfile charset has no `+`, and no catalogue keyword or rule stem does either). So this is a latent grammar defect, not a live bug.

The reason I'd still act on it is that the PR's own justification for prefix-anchoring is "a real Semgrep `rule_id` can legitimately contain `+`" — which is precisely the condition that would make the grader fabricate evidence. The design is defended by the one scenario it can't survive. Cheapest fix is to grade only the leading component so the two agree; the better fix is to stop packing receipts into one string and carry a list.

**3. G2 has an explicit bypass, and it predates this PR.** At `orchestrator.py:3128-3140`, when evidence is ungrounded but the hypothesis lexically overlaps a domain-model invariant, G2 is skipped and the finding stands with no receipt — the log line says `"G2 bypassed"` outright. The match (`:3104`) is three shared words of four-plus characters covering 15% of the invariant's vocabulary, which any hypothesis about the same subsystem will hit routinely.

The docstring justifies the leniency with "the invariant's mechanical provenance (from the study pipeline)". That premise doesn't hold: `core/concepts/study.py` opens with *"Phase 2 + 3 of `/understand --study`: LLM concept extraction"*. So it's LLM-authored hypothesis text matched against LLM-authored invariant text, and the result grounds a finding — which is the thing SKILL.md says cannot happen.

This is on `main` already and is not something this PR introduced, so I'd keep it out of here. But it's a wider opening than the string-forgery route this PR closes, and it should probably be an issue of its own before we describe G2 as structural anywhere externally.

## Suggested scope

In this PR: collapse the two validators (1), and align the grader with the validator on `+` (2) — both small and both directly on the guarantee this PR is about.

Separate issues: the CLI receipt consumption, and the domain-model G2 bypass (3).

On centralising LLM ingress — the rebase that was blocking it is done, so it's unblocked whenever you want it. Worth flagging that the live-sink requeue, flow-trace and error-retry paths do still reach tally without the full gate chain, so I'd treat it as the next piece of work rather than a nice-to-have. Happy to take it, or to split any of the above out if you'd rather review them separately.

grokjc:
I'm making some changes on main to resolve some of this. Stand by!

grokjc:
I've landed the namespace unification on main (4114316) — all evidence_tool stamps are now namespace:detail format (e.g. "dynamic:sanitizer", "frida:runtime", "triage:classifier"). After rebase, the remaining work on your side:

  - Rename "dynamic" → "dynamic:sanitizer" in _RECEIPT_GRADES
  - Add "dynamic", "frida", "dark_verify", "triage" to TOOL_EVIDENCE_NAMESPACES in _util.py (so namespaced composites like "dynamic:sanitizer" pass is_stamped_tool_evidence)
  - Drop is_tool_evidence, VALID_EVIDENCE_TOOLS, _TOOL_NAMESPACES, _RECEIPT_MAP from evidence_grade.py — your is_stamped_tool_evidence + _RECEIPT_GRADES are the single source of truth now

grokjc:
Landed a temporary  +-join fix on main - is_tool_evidence and grade_review_result now split on + and handle each part. Your _grade_tool_receipt already does this, so once the PR collapses the validators per Comment 10, the main-side code gets deleted and yours becomes canonical.

  For Issue 2: the remaining ambiguity is a rule ID containing + - "semgrep:corp.rule+dynamic" splits into a valid semgrep receipt and a valid bare "dynamic" receipt, fabricating evidence. Simplest fix: strip or reject + in the detail segment at _stamp_evidence time. Then + is unambiguously a separator and the grader is safe.

  The way forward after rebase:

  1. Rename "dynamic" to "dynamic:sanitizer" in _RECEIPT_GRADES
  2. Add "frida", "dark_verify", "triage" to TOOL_EVIDENCE_NAMESPACES in _util.py
  3. Drop is_tool_evidence, VALID_EVIDENCE_TOOLS, _TOOL_NAMESPACES, _RECEIPT_MAP from evidence_grade.py - is_stamped_tool_evidence + _RECEIPT_GRADES become the single source of truth
  4. Guard + in the detail segment at _stamp_evidence - reject or replace so +-splitting in _grade_tool_receipt is unambiguous

grokjc:
Latest batch landed on main (3fc7946a):

  - CLI receipt vocabulary: bare --evidence-tool joern now becomes joern:cli, distinguishing operator-asserted stamps from sweep-issued receipts. This closes the "half-closed" CLI diagnosis you flagged — the strict validator now recognises all CLI-issued stamps.
  - LLM evidence sanitiser: sanitize_llm_evidence_tool() namespaces LLM-supplied evidence_tool values under llm-claimed: so they can't satisfy is_tool_evidence(). Wired at all 4 LLM ingress points in the orchestrator.
  - dark_verify + compilation receipts: DARK_VERIFY and COMPILATION added to EvidenceSource with receipt map entries, so the grader recognises stamps the orchestrator already emits.
  - Malformed + guard: grade_review_result now checks is_tool_evidence() before entering the grading loop, so "semgrep+", "+joern", "semgrep++joern" are rejected.

  Combined with the earlier commits (namespace unification 4114316, +-join handling 7384c41), items 1 and 2 from your analysis in comment 4 are fully resolved on main. Rebase should collapse those conflicts.

  Remaining:
  - finalize_outcome centralisation (the 12 _tally_outcome call sites that bypass the gate chain) — happy for you to take this if you're still interested.
  - G2 domain-model bypass (L3128-3140) — tracking separately.

dinosn:
Thanks for picking these up — and you were right on the `+` parser, so let me concede that first.

## The `+` validator: your version should stand

Validating only the leading component and treating the tail as opaque isn't validation — it accepts a valid first component followed by an invalid one. I argued a false negative on a genuine receipt was the more dangerous direction, and for a detection tool that's right, but this is an evidence *gate*: rejecting a real receipt loses a finding, while accepting a fabricated one breaks the guarantee the subsystem exists to provide. False confirmation is categorically worse here, and I had it backwards.

I was also being inconsistent. I used an unreachable case (`semgrep:my+rule` — a genuine Semgrep rule ID containing `+`) to argue for overturning your parser, while treating an equally unreachable fabricated tail as not worth acting on. Same reachability, opposite treatment. And `+` is our composite delimiter — no producer emits one inside a detail, so `semgrep:my+rule` isn't a valid RAPTOR receipt today; I was conflating Semgrep's identifier space with our serialisation grammar.

The one part worth keeping in the record: raw `+` stays ambiguous if details can ever contain it. The answer when that day comes is structured or escaped receipt components, not prefix acceptance. Your all-parts parser is the right implementation for the grammar we actually have.

## One thing that survives: the canonical parser isn't the G2 decision point

`_check_finding_gates` doesn't use `is_tool_evidence`. On `main` today:

```python
evidence = outcome.evidence_tool or ""
if outcome.review_result:
    evidence = outcome.review_result.get("evidence_tool", evidence)   # :3263 — prefers the raw nested dict
...
if not evidence or evidence.lower().strip() in _LLM_ONLY_EVIDENCE:    # :3268 — the old 9-element blacklist
```

Two things follow. The nested `review_result["evidence_tool"]` is written only by `_stamp_evidence` (`:6075`) and `_promote_suspicious` (`:5619`) — both tool paths — and `_sanitize_llm_et` never touches it, so the copy G2 reads in preference is the one the sanitiser doesn't reach. And the gate still tests membership of `_LLM_ONLY_EVIDENCE` (`:3205`), a nine-element blacklist, rather than the canonical parser.

So a model returning `evidence_tool: "joern"` lands it in `review_result`; G2 reads that copy; `"joern"` isn't in the blacklist; the finding passes. That's the original bug, and `test_g2_joern_passes` (`test_orchestrator.py:1278`) currently asserts it as intended behaviour.

This is the same defect my round 1 caught in my own first cut — I set `outcome.evidence_tool` and left the raw string in `review_result`, which G2 and `_sweep_validate` read in preference. Which is really the argument for doing it at the LLM boundary rather than per-path: four ingress points is four things to keep in sync, and the nested copy is a fifth. `joern:cli` is a symptom of the same thing — it grades identically to bare `joern`, so the suffix records provenance without conferring or withholding authority.

I'm not reopening #892 over this; it wants its own change. Flagging it because it's the specific gap between what the canonical parser can do and where the decision actually gets made.

## Patch 2 split out — #897

The Mode-2 synthesis crash fix is independent of all of the above, so I've split it out against current `main`: **#897**, CI green.

It also carries a durability fix beyond the original. Fixing the `KeyError` removes the known crash but not the class — the second pass runs after the main review loop and before the state sync-back and the single one-shot `collector.flush()`, so *any* exception there still discards the run's findings and buffered audit log. It now degrades to "second pass skipped" with main-pass results kept, and `gap_for_site` is total against a malformed checklist. Tests cover 12 malformed hit shapes and 11 malformed checklist shapes.

One thing I noticed while rebasing: `92cd8a85` changed the synthesis dedup key to `{m.file}:{m.line}`, and the bulk-land in `aea9d17d` reverted it — `main` currently has the pre-`92cd8a85` form. #897 supersedes it by removing both filters, but you may want to check whether anything else from that batch was rolled back the same way.

## Patch 3 held back

Source hydration is also independent, but I'm not resubmitting it yet. Three reasons:

- **Scope.** I described it as switching on two detectors. It's seven modules that read `gap["source"]` — `negative_space` (14 refs), `orchestrator` (4), `spec_inference` (4), `taint_specs` (2), `iris_specs`, `project_context`, and `triage`. `triage.py:249` changes which functions get reviewed at all via generated-file detection, and `spec_inference` can add LLM calls. That's a bigger activation than the PR described, and it should be narrowed to the detectors it's for.
- **Your G7 work.** `negative_space.py` has no awareness of `lexical_dead` / `module_aborts_on_load` / `build_excluded`, so convention discovery would learn its baseline partly from code G7 just declared dead, and sibling asymmetry is a majority-vote over that baseline. Skipping hydration for dead gaps isn't enough either — `check_sibling_negative_space` reads an unhydrated sibling as non-adherent, so they need excluding from the detector input rather than just left unhydrated.
- **Two implementation defects of my own**: `_read_spans` is effectively O(file lines × spans per file), and the byte ceiling isn't a peak bound because bodies are collected for a whole file before the running total is enforced.

I'll resubmit it narrowed, with the dead-code exclusion and a calibration run.

Oof on the CI is fighting me today :-)

dinosn:
Closing this — after the split there's nothing left in it to land.

- **Patch 1 (evidence contract)** — conceded. Your all-parts parser is the right implementation for the grammar we have; my prefix-anchored version validated only the leading component and was the weaker design. Nothing to merge.
- **Patch 2 (Mode-2 synthesis crash)** — split out to **#897**, branched off current `main`, CI green. Also carries the durability fix: the second pass no longer takes the run's findings and buffered audit log with it when anything in resolution, graph construction or the executor raises.
- **Patch 3 (gap source hydration)** — held. It activates seven `gap["source"]` consumers rather than the two detectors I described, it needs to exclude the gaps your G7 work flags dead before conventions are learned from them, and it has two performance defects of mine to fix first. Coming back as a fresh, narrower PR.

The three files GitHub flags as conflicting here are all patch-1 or patch-3 content, so resolving them would mean hand-merging two competing evidence contracts and then not shipping the result.

The one open thread worth carrying forward is the G2 ownership gap in [the previous comment](https://github.com/gadievron/raptor/pull/892#issuecomment-5107685591) — `_check_finding_gates` reads `review_result["evidence_tool"]` in preference (`orchestrator.py:3263`) and tests the `_LLM_ONLY_EVIDENCE` blacklist (`:3268`) rather than the canonical parser, so bare `joern` from a model still passes G2 and `test_orchestrator.py:1278` asserts it. That wants its own change rather than this one. Happy to take it if you'd like it picked up.

Thanks for the review throughout — the `+` catch was right and I'd have shipped the weaker parser without it.

grokjc:
G2 gate fix landed on main (b9b691ba):
  
  - _check_finding_gates now prefers outcome.evidence_tool (authoritative — set by _stamp_evidence or sanitised at LLM ingress). Falls back to review_result["evidence_tool"] only when empty, with _sanitize_llm_et as defence in depth.
  - Gate test switched from _LLM_ONLY_EVIDENCE blacklist to is_tool_evidence(). The constant is removed.
  - 3 new tests: LLM-claimed "joern" in review_result without a stamp is rejected; stamped outcome.evidence_tool still passes; llm-claimed: prefix is rejected.

grokjc:
On Patch 3 - The dead-code signals (lexical_dead, module_aborts_on_load, build_excluded) flow through the checklist already. I'll try to tag dead gaps in compute_gaps so negative_space and sibling_asymmetry can exclude them from
  their input rather than having to understand G7's internals. That way convention discovery doesn't learn its baseline from code we've already declared dead, and unhydrated dead siblings don't read as non-adherent.

  Your narrowing plan sounds right - hydrate for negative_space and sibling_asymmetry only, leave the other five modules (triage, spec_inference, orchestrator, taint_specs, iris_specs, project_context) degrading gracefully when source is absent. Worth verifying none of them silently change behaviour when it's populated - triage.py:249's generated-file detection changing which functions get reviewed is exactly the kind of unintended activation you'd want to catch.

  I'll get the dead-gap tagging done on main so you have a clean interface to build against when you resubmit.

grokjc:
Dead-gap tagging landed on main (6f0d7ea6) for patch 3 prep:
  
  - compute_gaps sets a unified dead flag when any dead-code signal (lexical_dead, module_aborts_on_load, build_excluded) is present.
  - The orchestrator filters to live_gaps before passing to discover_conventions and check_sibling_negative_space, so new consumers get live-only gaps by default.
  - negative_space also filters per-function as defence in depth (detect_framework, discover_conventions, check_all_negative_space, check_sibling_negative_space).

  This gives you a clean interface for the narrowed resubmission — hydrate for negative_space and sibling_asymmetry, skip dead gaps via gap.get("dead"), and the baseline/majority-vote won't learn from code G7 declared dead.

grokjc:
All work from this PR is now resolved on main:

  Evidence contract (Patch 1):
  - Namespace unification (41143166), +-join handling (7384c41), CLI receipt vocabulary + LLM sanitiser (3fc7946), G2 gate fix (b9b691ba)

  Rollback fix:
  - Synthesis dedup key restored to {file}:{line} — was rolled back by the bulk-land in aea9d17d, now fixed on main

  Patch 3 prep:
  - Dead-gap tagging in compute_gaps + live_gaps filter in orchestrator (6f0d7ea) — clean interface for the narrowed source hydration resubmission

  Remaining (separate tracking):
  - PR #897 (Patch 2, Mode-2 synthesis crash) — ready for review
  - Patch 3 (source hydration) — awaiting narrowed resubmission
  - finalize_outcome centralisation — Over to you?
  - G2 domain-model bypass — pre-existing, tracked separately

  Thanks for the thorough analysis throughout — the G2 nested-copy finding in particular was a real gap we'd missed.

PR #897, opened 2026-07-28
fix(audit): resolve Mode-2 synthesis hits to functions before graph build

dinosn:
Split out of #892 so it can land independently of the evidence-contract work there, which is now contested and which this does not depend on. Branched from current `main`.

## The bug

Checker synthesis is one of the better ideas in `/audit` — confirm a bug once, generalise it to a rule, sweep the codebase. The payoff is currently unreachable, because the failure fires on **success**.

Synthesis hits are *sites*, carrying `function` (`checker_synthesis.py:204`). `TaskGraph.from_workqueue` dereferences `gap['name']` unguarded (`task_graph.py:197`), and the second-pass graph construction sits outside the synthesis `try/except` — `run_orchestrator` has only a `finally` for Joern cleanup. So a confirmed finding whose synthesized rule matched in another file raises `KeyError: 'name'` **after the entire review loop has run**, before state sync-back and before `collector.flush()` — losing the run's findings and its buffered audit log.

Reproduced against current `main`:

```
1) raw hits -> TaskGraph.from_workqueue   KeyError('name')
2) after this patch                        resolved -> [('src/a.c','parse_hdr')], graph builds
3) unresolvable site                       0 gaps, dropped cleanly
```

## The fix

`_synthesis_hits_to_gaps` resolves each site to its enclosing checklist function via a new `gaps.gap_for_site`, mirroring the resolution already used by `understand_bridge._find_containing_function`: strict smallest-containing-span first, so a nested function or closure is attributed to the innermost match rather than the enclosing one; then a closest-preceding-`line_start` fallback bounded by the next definition, for inventories that record only the definition line. A trailing definition with no next start has no derivable bound, so those sites stay unresolved rather than being given a fabricated span that would truncate review.

Two deliberate choices:

- Sites that resolve to no reviewable function (a top-level initializer, a macro, generated code) can't become review tasks, but they're still tool signal — they go to `unresolved-synthesis-hits.json` rather than being dropped silently.
- Functions reviewed in an earlier round or run are deliberately **not** skipped. A synthesized checker encodes a pattern nobody knew about when those functions were first reviewed, which is exactly when re-review pays.

This also removes the two filters in `checker_synthesis.py` that dropped real variants: the dedup key was `<matched file>:<seed function>` — a seed function name paired with a different file, which identifies nothing and matched `seen_keys` only by accident — and `m.file != file_path` discarded every same-file sibling, so a second vulnerable function in the same source file was lost. Both belong downstream, once a site has been resolved to a function. Excluding by line in `checker_synthesis` would be wrong anyway, since `outcome.line` is the function's `line_start` rather than the vulnerable line, and is `0` when the inventory recorded none.

Incidentally this restores the intent of 92cd8a85, which changed that dedup key and was then reverted by the bulk-land in aea9d17d — `main` currently has the pre-92cd8a85 form. Worth a look independently of this PR.

## Durability

Fixing the `KeyError` removes the known crash but not the class. The second pass is optional enrichment that runs after the main review loop and before the state sync-back and the single one-shot `collector.flush()`, so **any** exception in resolution, graph construction or the executor still discards the run's output.

It is now wrapped: the pass degrades to "second pass skipped" with a warning, main-pass results are kept and buffered state is flushed. `gap_for_site` is also made total — a malformed checklist yields fewer resolved gaps rather than an exception, which is the correct degradation for something holding the run's results hostage.

## Testing

`test_synthesis_hits.py` asserts `pytest.raises(KeyError)` on the raw hits *before* showing the resolved gaps work, so it fails loudly if the resolution is ever removed. Added on top: 12 malformed hit shapes and 11 malformed checklist shapes assert resolution never raises, that one bad hit doesn't discard the batch, and that whatever resolution returns is always safe to hand to `TaskGraph.from_workqueue`.

### DIFF
--- a/core/audit/checker_synthesis.py
+++ b/core/audit/checker_synthesis.py
@@ -116,7 +116,10 @@ def synthesize_and_sweep(
     Args:
         outcome: ReviewOutcome with file, function, hypothesis, review_result.
         config: OrchestratorConfig with target_path, out_dir, codeql_db_path.
-        seen_keys: Set of file:function keys already in the workqueue.
+        seen_keys: Retained for logging only. Site-level dedup happens
+            downstream in ``orchestrator._synthesis_hits_to_gaps``, after
+            each site is resolved to its enclosing function — a set of
+            ``file:function`` keys cannot filter function-less sites.
         synthesis_count: How many syntheses have been done this run.
         max_per_run: Maximum synthesis attempts per run.
 
@@ -201,16 +204,31 @@ def synthesize_and_sweep(
     file_path = getattr(outcome, "file", "")
     function = getattr(outcome, "function", "")
 
+    # Emit every match, each carrying its origin. Two filters used to
+    # live here and both dropped real variants:
+    #
+    #   * the dedup key was `<matched file>:<seed function>` — the seed's
+    #     function name paired with a different file, which identifies
+    #     nothing, and matched `seen_keys` only by accident;
+    #   * `m.file != file_path` discarded every same-file match, so a
+    #     second vulnerable sibling in the same source file was lost.
+    #
+    # Sites are function-less by nature (a codebase-wide rule matches
+    # text), so both dedup and seed exclusion belong downstream in
+    # `orchestrator._synthesis_hits_to_gaps`, once each site has been
+    # resolved to its enclosing function. Excluding by line here would be
+    # wrong anyway: `outcome.line` is the function's line_start, not the
+    # vulnerable line, and is 0 when the inventory recorded none.
     new_hits = []
     for m in cs_result.matches:
-        key = f"{m.file}:{function}"
-        if key not in seen_keys and m.file != file_path:
-            new_hits.append({
-                "file": m.file,
-                "line": m.line,
-                "function": "",
-                "snippet": m.snippet or "",
-            })
+        new_hits.append({
+            "file": m.file,
+            "line": m.line,
+            "function": "",
+            "snippet": m.snippet or "",
+            "origin_file": file_path,
+            "origin_function": function,
+        })
 
     logger.info(
         "synthesis %s: %d sweep matches, %d new (deduped against %d seen)",
--- a/core/audit/gaps.py
+++ b/core/audit/gaps.py
@@ -196,6 +196,126 @@ def load_checklist(out_dir: Path) -> Dict[str, Any]:
         return {}
 
 
+def gap_for_site(
+    checklist: Dict[str, Any],
+    file_path: str,
+    line: int,
+    *,
+    priority: int = 1,
+) -> Optional[Dict[str, Any]]:
+    """Build a reviewable gap for the function enclosing ``file_path:line``.
+
+    Mechanical sweeps (Mode-2 checker synthesis, rule replay) report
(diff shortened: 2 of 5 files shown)


dinosn:
Merged `main` in. One conflict worth explaining rather than just resolving, because it's about `b31e27bf` — thanks for picking that up off the revert note.

I resolved `checker_synthesis.py` in favour of this branch, because the restored dedup can't fire.

`b31e27bf` sets the key to `<file>:<line>`, but the callers build `seen_keys` as `<file>:<function>`:

```python
orchestrator.py:1074   seen_keys = (reviewed_set or set()) | {f"{g['file']}:{g['name']}" for g in (workqueue or [])}
orchestrator.py:4829   seen_keys = {f"{o.file}:{o.function}" for o in result.outcomes}
```

A `file:line` string is never a member of a set of `file:function` strings, so `key not in seen_keys` is always true. Running main's loop against the key format the callers actually produce:

```
seen_keys (file:function) : ['src/a.c:parse_hdr', 'src/b.c:parse_body']
computed keys (file:line)  : ['src/a.c:42', 'src/b.c:77', 'src/seed.c:10']
membership hit?            : False
-> emitted [('src/a.c', 42), ('src/b.c', 77)]
```

`src/a.c:parse_hdr` is already reviewed, and site `src/a.c:42` is emitted anyway. So the dedup is inert, and the only filter still doing work is `m.file != file_path` — the one that drops same-file siblings. The docstring at `checker_synthesis.py:119` still says "Set of file:function keys", which is the accurate description of what callers pass.

That's not really a flaw in your change so much as confirmation that the key belongs elsewhere: a match is a *site*, and `seen_keys` is keyed by *function*, so the two can only be compared after resolution. This branch removes both filters and dedups in `_synthesis_hits_to_gaps`, once each site has an enclosing function — where the formats line up. Same-file siblings survive, and a site whose function was already reviewed is still excluded, which is what the original key was reaching for.

Your test edit was a one-line change to the hardcoded `seen` fixture; this branch makes `seen` a parameter, so nothing is lost — `test_sites_survive_a_prior_run_key` and `test_same_file_sibling_is_kept` cover it from both directions.

If you'd rather keep a cheap site-level dedup upstream as well, that works fine alongside the downstream one — it just needs its own `seen_sites` set in `file:line` form rather than sharing `seen_keys`. Happy to add that if you want the earlier cut.

Full `core/audit` suite on the merge: 2993 passed, 199 skipped, 0 failed. Ruff clean.

grokjc:
Verified all four claims against current origin/main:

  1. KeyError: 'name' — confirmed. Synthesis hits carry function, not name. TaskGraph.from_workqueue (task_graph.py:198) crashes.
  2. Crash loses main-pass output — confirmed. No try/except around the second pass; KeyError propagates past state sync and collector.flush().
  3. m.file != file_path drops same-file siblings — confirmed (checker_synthesis.py:207).
  4. Dedup key mismatch — confirmed. Caller sends {file}:{name} (orchestrator.py:1074), checker tests {file}:{line} (L206). Inert by design flaw, not just by accident — as you noted in your merge comment, the
  formats can never intersect.

  Resolved the import conflict on our side (took your gap_for_site, dropped mark_checked — unused on main, only referenced in a comment). Merging. Thanks!

PR #898, opened 2026-07-28
fix(audit): hydrate source for the negative-space detectors

dinosn:
Resubmission of patch 3 from #892, narrowed as discussed there and rebuilt against your dead-gap work in 6f0d7ea6. Branched from current `main`.

## The gap

`discover_conventions`, `check_negative_space` and `check_sibling_negative_space` all read `gap["source"]` and skip any gap without it. Gaps are built from the checklist, which carries line spans but no text, so the key was never populated — convention discovery returned empty, and `check_sibling_negative_space` was never called at all, being guarded on `if conventions`. Three detectors wired into the default path have been silent no-ops since they were written.

## Narrowed, as agreed

`gap["source"]` has **seven** readers across `core/audit` — `negative_space`, `orchestrator`, `spec_inference`, `taint_specs`, `iris_specs`, `project_context`, `triage`. Hydrating the shared gap dicts switches on all of them; `triage.py:249` would begin generated-file detection, which changes *which functions get reviewed*, and `spec_inference` would issue extra LLM calls.

So hydration returns **copies** and hands them only to the detectors that need them. The originals are untouched, the other five modules see exactly what they see today, and `write_gaps` needs no knowledge of any of it — which also drops the `source`-stripping the #892 version needed.

## Dead gaps

Excluded before any I/O, via the unified `dead` flag from 6f0d7ea6 — thanks, that interface was exactly the right shape. Worth noting the exclusion has to happen at the input, not just at hydration: `check_sibling_negative_space` reads an unhydrated sibling as *non-adherent*, so leaving dead gaps in but unhydrated would actively manufacture asymmetry findings rather than merely skew the baseline. There's a test asserting dead gaps cost no `_read_spans` call at all.

## The two defects from the #892 review

**`_read_spans` was O(lines × spans).** It tested every requested span against every line — pathological on a large file carrying many functions. Now a sliding active set: spans sorted by start, opened when the stream reaches them, dropped once past their end. Measured on 200k lines × 2000 spans: **0.664s → 0.004s, ~150x, byte-identical output**. There's a scaling test that fails if it regresses to the quadratic form.

**The ceiling wasn't a peak bound.** Bodies for a whole file were built before the caller enforced the running total. Now `budget_bytes` carries the caller's remaining allowance and is charged during accumulation. Three related corrections while I was in there: abandoning a span refunds what it buffered, so oversized spans early in a file can't starve valid later ones; the active set is iterated sorted, so which span survives a tight budget doesn't depend on hash iteration order; and duplicate spans are charged once, matching the single string actually retained.

A span still open at EOF is now abandoned rather than returned partially. For absence detection a truncated body is worse than none — the check we'd report as missing may simply lie in the part we didn't read.

## Two more, only reachable once the detectors run

Both surfaced in review of this patch, and both are pre-existing, but they're in the blast radius of turning these detectors on so they belong here:

- **`check_negative_space` was still getting the raw unhydrated gap** in the review loop (`orchestrator.py`), so discovered conventions never reached the per-function check. Hydrating for `discover_conventions` alone would have delivered half the feature. It now gets a throwaway copy carrying the body already present in `ctx`.
- **Sibling findings were attached back by substring-matching** the function name against the finding's prose `evidence`. That misroutes on duplicate names across files and on prefixes like `handle_user` / `handle_user_admin`. `NegativeSpaceFinding` already carries `file` and `function`, so matching is now on exact identity.

## Testing

73 new tests in `test_gap_hydration.py`, covering the end the change serves (conventions `[]` unhydrated, `['auth']` hydrated — so the assertions aren't vacuous), blast-radius containment, dead-gap exclusion, the bounds, budget accounting, and span correctness **differentially against a naive file-slicing reference** across randomised configurations with overlapping, nested, duplicate and adjacent spans, spans on line 1, spans past EOF, and files with no trailing newline.

Full `core/audit` suite: **3013 passed, 199 skipped, 0 failed**. `ruff --select F401,F811,F821,F841` clean.

## Known limits, pinned by tests rather than claimed away

- The NUL binary probe reads the first 8 KiB only. It's a cheap filter, not a guarantee; there's a test documenting that a NUL past the window isn't caught.
- The convention baseline is drawn from the **budgeted** gap list, so it samples the current work batch rather than the whole project — it shifts with `--budget`, prior coverage and ordering. Defensible for a relative signal, but it isn't "the project's conventions" in the strong sense the module's docstring implies. Worth a follow-up building the corpus from the checklist independently of review budget, if you think it's worth the extra pass.

One thing I did **not** touch, which you may want to look at before these detectors carry weight: peer groups in `sibling_analysis.py` are global by verb, so every `handle_*` in the repo can land in one group. With the detector previously dead that was harmless; now that it runs, unrelated subsystems can be treated as siblings. Happy to take that as a separate change.

### DIFF
--- a/core/audit/gaps.py
+++ b/core/audit/gaps.py
@@ -22,7 +22,7 @@ from pathlib import Path
 from typing import Any, Dict, List, Optional
 
 from .strategy import strategies_from_item
-from ._util import extract_context_map_set
+from ._util import extract_context_map_set, safe_join
 
 logger = logging.getLogger(__name__)
 
@@ -37,6 +37,14 @@ _TRIVIAL_SLOC = 5
 _SMALL_ENTRY_SLOC = 20
 _LARGE_SLOC = 200
 _REVIEWABLE_KINDS = frozenset({"function", "method", ""})
+
+# Bounds for detector source hydration. These exist so a large or
+# hostile tree degrades by hydrating fewer functions rather than by
+# exhausting memory.
+_MAX_HYDRATED_SLOC = 2000            # per function, lines
+_MAX_HYDRATED_FUNCTION_BYTES = 256 * 1024
+_MAX_HYDRATED_FILE_BYTES = 8 * 1024 * 1024
+_MAX_HYDRATED_TOTAL_BYTES = 64 * 1024 * 1024
 
 
 def compute_gaps(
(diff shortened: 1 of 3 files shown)


grokjc:
Verified claims against origin/main. All three detectors (discover_conventions, check_negative_space, check_sibling_negative_space) are confirmed silent no-ops — gap["source"] is never populated from the
  checklist.

  The copy-not-in-place design is correct: source has seven readers across core/audit, and populating the shared dicts would switch on triage's generated-file detection and spec_inference LLM calls as side effects.

  Sibling matching fix confirmed: current code (orchestrator.py:717) does gap.get("name", "") in f.evidence — substring match against prose. handle_user matches findings for handle_user_admin. The identity match (f.file, f.function) == (gap_file, gap_name) is the right fix.

  Resolved conflicts (kept both hydrate_live_gaps_for_detectors/_read_spans and gap_for_site in gaps.py; import line carries all three new names, dropped unused mark_checked).

  Re: the peer grouping flag — confirmed. identify_peer_groups (sibling_analysis.py:255) groups globally by verb prefix with no file or subsystem scoping. Every handle_* across the repo lands in one group. Was harmless when the detector was inert; worth scoping before these findings carry real weight. Separate concern, agree it doesn't belong in this PR.

grokjc:
Landed a layered peer group resolver in 86efe60 that replaces the flat identify_peer_groups with a seven-layer hierarchy:

  - L0–L1: Joern CPG and r2 binary co-callee groups (definitive — functions called from the same caller body)
  - L2: Dispatch table handlers (tree-sitter extracted)
  - L3: Domain model concepts (from study runs)
  - L4: Type cohort (functions operating on the same struct/type)
  - L5: Verb-prefix, now scoped per-directory with signature-shape filtering (the fix for the issue you raised)
  - L6: Paired operations (encode/decode, lock/unlock, etc.)

  L0–L3 claim functions exclusively (higher-confidence layers remove functions from lower layers' input). L4–L6 run independently. All inputs are optional — missing data skips that layer gracefully.

  The directory scoping on L5 means src/api/handle_get and src/db/handle_get land in separate groups, and signature filtering further splits groups where the verb prefix is coincidental (e.g. handle_shutdown() with no params won't group with handle_get(req: Request) -> Response).

grokjc:
Worth noting on convention baseline per-batch: the cleaner fix might be to fold convention discovery into the per-peer-group check rather than widening the sample.

  check_sibling_negative_space already iterates peer groups and checks each sibling's adherence — the only change is discovering conventions within each group rather than receiving them from a separate project-wide pass. Three handle_* functions in src/api/ all call sanitize_input(), the fourth doesn't — that's both the convention and the violation, discovered from the most relevant sample possible.

  The current split (discover globally, check per-function) means the baseline shifts with --budget and ordering, and project-wide patterns from unrelated subsystems dilute the signal. Per-group discovery eliminates the sample-size problem entirely — the sample is always exactly the peer group — and removes the separate hydration pass.

grokjc:
514f276 also wires resolve_peer_groups into check_sibling_negative_space in negative_space.py, so once your hydration lands the sibling detector uses the directory-scoped groups too, not the old global grouper.
  
Good to merge. Thanks!

grokjc:
## Follow-up: full gap["source"] consumer consolidation

  PR #898 fixed the negative-space *detectors* (pre-loop). The follow-up work (landed across 4 commits on main after this PR) closed **every** remaining `gap["source"]` consumer:

  ### What landed

  1. **`read_gap_source()` shared helper** (`gaps.py`) — single implementation wrapping `_read_spans` with type validation, `safe_join` path containment, binary detection, and 256 KiB per-function cap. All consumers now go through this.

  2. **Consumer migration:**
     - `taint_specs.py` — replaced 20-line duplicated reader with 4-line delegation
     - `triage.py` (`_read_file_header`) — migrated from hand-rolled path join to `safe_join`
     - `spec_inference.py` — wired to `read_gap_source`
     - `iris_specs.py` — wired to `read_gap_source`

  3. **Post-loop negative-space hydration** — all 9 pattern-scan functions (`check_resource_exhaustion`, `check_protocol_ambiguity`, `check_missing_app_features`, `check_ub_patterns`, `check_signal_safety`,
  `check_side_channels`, `check_multi_process`, `check_deployment_assumptions`, `check_lock_ordering`) now accept `target_path` and read source from disk when `gap["source"]` is empty.

  4. **`check_missing_app_features` refactor** — from unbounded `"\n".join(all_source)` concatenation to per-gap search with `found_features: Set[int]` and early exit when all features found.

  5. **Tests** — 10 new `TestReadGapSource` tests + 7 new `TestPostLoopHydration` tests. Full audit suite passes (3304 tests).

  ### Design: no caps on post-loop reads

  The pre-loop `hydrate_live_gaps_for_detectors` uses SLOC + total byte caps because sibling-check holds all gaps in `gap_by_key` simultaneously. The post-loop functions iterate per-gap and don't accumulate, so they use only the per-function 256 KiB cap via `read_gap_source` — no SLOC or total byte ceiling.

  All three copies of path-traversal guard are now consolidated into `read_gap_source` → `_read_spans` → `safe_join`.

PR #903, opened 2026-07-31
fix(libexec): mark the study helpers executable

dinosn:
Two files in `libexec/` are committed `100644` while the other 65 are `100755`:

- `libexec/raptor-study-prep`
- `libexec/raptor-study-loop`

Both begin with `#!/usr/bin/env python3` and are documented to be invoked directly, so the prep half of the study pipeline could not run as written:

```
$ "$RAPTOR_DIR/libexec/raptor-study-prep" <subsystem> <out-dir> --root <root> --concept <lens>
permission denied: /path/to/raptor/libexec/raptor-study-prep
```

`raptor-study-run` was already `100755`, so the breakage only showed up on the first stage.

**Mode-only change.** Both blobs are byte-identical — `git diff-tree` reports the same SHA on each side, `100644 -> 100755`, zero content lines:

```
:100644 100755 3415934f 3415934f M  libexec/raptor-study-loop
:100644 100755 378c49f9 378c49f9 M  libexec/raptor-study-prep
```


grokjc:
Well spotted - just landing a bunch more changes to study right now, too! Thanks.

PR #913, opened 2026-08-11
feat(dispatcher): add OrcaRouter as a named provider

Marc-oss-hub:
## Summary

Adds **[OrcaRouter](https://www.orcarouter.ai)** as a named LLM provider in the dispatcher, mirroring the existing OpenRouter wiring exactly. OrcaRouter is an OpenAI-compatible gateway (base URL `https://api.orcarouter.ai`, API rooted at `/v1`) that fronts a broad model catalog under namespaced IDs (e.g. `openai/gpt-5.5`, `anthropic/claude-sonnet-4.6`).

## What this changes

- `core/llm/dispatcher/auth.py` — new `ProviderRule` for `orcarouter` (`upstream_base_url="https://api.orcarouter.ai"`, Bearer auth via `ORCAROUTER_API_KEY`), plus the credential read in `CredentialStore` (read-and-erased like every other key).
- `core/llm/dispatcher/server.py` — route prefix `/orcarouter/` → `orcarouter`, so an OpenAI SDK pointed at `<dispatcher>/orcarouter/v1` forwards `/v1/chat/completions` etc. to the OrcaRouter upstream.
- `core/config/__init__.py` — `ORCAROUTER_API_KEY` added to the LLM env allowlist.
- `core/security/llm_family.py` — `orcarouter/` treated as an aggregator prefix for model-family detection (cross-family retry safety).
- Tests — provider-injection, unconfigured-503, env read-and-erase, seed-from-config, and family-detection coverage for the new provider (all mirroring the openrouter cases).

## Why a separate provider

The dispatcher already treats OpenAI-compatible aggregators as first-class named providers (OpenRouter, Groq, Together, Fireworks, ...), each with its own key, upstream URL, and path prefix. OrcaRouter is the same shape, so routing it through the generic OpenAI passthrough would silently skip the credential management, env erasure, and audit-path guarantees every other aggregator gets. Keeping it as a named rule preserves those guarantees.

## How to use it

```bash
export ORCAROUTER_API_KEY=sk-orca-...
# point an OpenAI-compatible client at the dispatcher:
# base_url = http://<dispatcher>/orcarouter/v1, model = openai/gpt-5.5
```

## Verification

- `pytest core/llm/dispatcher/tests/test_providers.py core/llm/dispatcher/tests/test_seed_from_config.py core/security/tests/test_llm_family.py` — 63 passed (the dispatcher server-socket cases require `AF_UNIX` and are skipped on this Windows host; they fail identically on `main` without these changes).
- L3 live: `POST https://api.orcarouter.ai/v1/chat/completions` with a real key and `openai/gpt-5.5` → HTTP 200, `content: "ORCA-OK"` — confirming the exact upstream URL the dispatcher constructs (`bare host` + `/v1/chat/completions`).

## Disclosure

I'm an engineer on the OrcaRouter team. OrcaRouter is an API gateway that never inspects or stores prompt/response content in plaintext; credentials are forwarded only to the upstream model provider chosen for the request.

### DIFF
--- a/core/config/__init__.py
+++ b/core/config/__init__.py
@@ -849,6 +849,7 @@ class RaptorConfig:
         "GROQ_API_KEY",         # aggregator + family stem (batch 067)
         "TOGETHER_API_KEY",     # aggregator
         "OPENROUTER_API_KEY",   # aggregator
+        "ORCAROUTER_API_KEY",   # aggregator
         "FIREWORKS_API_KEY",    # aggregator
         "DEEPINFRA_API_KEY",    # aggregator
         "PERPLEXITY_API_KEY",   # aggregator
--- a/core/llm/dispatcher/auth.py
+++ b/core/llm/dispatcher/auth.py
@@ -245,6 +245,7 @@ class CredentialStore:
             "groq":       _read_env("GROQ_API_KEY"),
             "together":   _read_env("TOGETHER_API_KEY"),
             "openrouter": _read_env("OPENROUTER_API_KEY"),
+            "orcarouter": _read_env("ORCAROUTER_API_KEY"),
             "fireworks":  _read_env("FIREWORKS_API_KEY"),
             "deepinfra":  _read_env("DEEPINFRA_API_KEY"),
             "perplexity": _read_env("PERPLEXITY_API_KEY"),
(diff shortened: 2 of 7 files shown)


grokjc:
Thanks, clean PR! LGTM

PR #917, opened 2026-08-18
New Sandbox Triage

grokjc:
I built on your excellent idea @danielcuthbert and added some more goodness. I had recently changed the sandbox on main and your stuff was well timed 😄

### DIFF
--- a/.github/scripts/vocab_baseline.json
+++ b/.github/scripts/vocab_baseline.json
@@ -1206,6 +1206,11 @@
   "count": 19,
   "kind": "literal"
  },
+ "core/sandbox/escalation_signatures.py::ESCAPE_PRIMITIVE_SYSCALLS": {
+  "count": 16,
+  "kind": "literal",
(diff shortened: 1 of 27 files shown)


grokjc:
Just need to sort the OSX CI failures now ...

danielcuthbert:
Had a proper rummage through this and pushed a small fix-up commit because there were a few trust-boundary bits that felt a bit too forgiving.

The main one was unstamped triage input being treated as legacy during a live run. That meant a target could plant its own `sandbox-summary.json` and still influence the final verdict. The lifecycle path now fails closed on unstamped telemetry, while manual re-triage of genuinely old runs still works.

Also tightened the deep triage pass so it will not reason over unstamped reports by default, scoped resolved-IP alert caps to the active sandbox bucket rather than the proxy singleton, and stopped telemetry MAC from accepting mangled key files.

Added regression coverage for planted summaries, planted triage files, stale resolved-IP suppression, and short/oversized MAC keys. Focused sandbox suite is green: `147 passed, 34 skipped`.

Nothing dramatic, just closing off a few places where the target could be a bit too helpful with its own evidence. Also this is sexy af. seriously

grokjc:
Done some cleanup and squashed some more bugs. Good to go?

PR #918, opened 2026-08-19
Feat/appsec controls docs

grokjc:
I updated and refreshed your docs - I think this is good to merge?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@
 
 ```
 
+<a href="https://smithery.ai/skills?ns=gadievron&utm_source=github&utm_medium=badge"><img src="https://smithery.ai/badge/skills/gadievron"></a>
 <a href="https://github.com/gadievron/raptor/actions/workflows/github-code-scanning/codeql"><img src="https://github.com/gadievron/raptor/actions/workflows/github-code-scanning/codeql/badge.svg"></a>
 
 **Authors:** Gadi Evron, Daniel Cuthbert, Thomas Dullien (Halvar Flake), Michael Bargury, John Cartwright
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[project]
+name = "raptor"
+version = "3.0.0"
```
danielcuthbert, line 3: I guess this should be 3.1.0 right?

README.md
```diff
@@ -112,46 +93,13 @@ Once inside, just say "hi" to get started, or jump straight to a command.
...
-
----
-
 ## What RAPTOR can do
 
 | Command | What it does | Status |
 |---------|-------------|--------|
 | `/agentic` | Full autonomous workflow: scan, validate, exploit, patch | Stable |
 | `/scan` | Static analysis with Semgrep and CodeQL | Stable |
 | `/understand` | Map attack surface, trace data flows, hunt vulnerability variants | Stable |
```
danielcuthbert, line 151: why did we remove /binary? it works well no?

README.md
```diff
@@ -161,9 +109,7 @@ Environment variables that could inject code into the launcher chain are strippe
 | `/crash-analysis` | Autonomous root-cause analysis for C/C++ crashes | Stable |
 | `/oss-forensics` | Evidence-backed forensic investigation for GitHub repositories | Stable |
 | `/project` | Named workspaces to organise runs and track findings over time | Stable |
-| `/threat-model` | Create, inspect, and maintain per-project threat models | Stable |
-| `/sage` | Persistent memory layer (store, recall, link, corroborate) | Stable |
-| `/frida` | Dynamic instrumentation via Frida | Alpha |
+| `project threat-model` | Project-owned threat model for focus areas, trust boundaries, and proof expectations | Beta |
```
danielcuthbert: also /threat-model

danielcuthbert (commented):
a few comments about why we've removed some bits

pyproject.toml
grokjc: It is auto-stamped by the release process, so 3.0.0 is correct right now 😄

README.md
grokjc: Because the rebase screwed this up 😄

grokjc:
Rebase was bad, now this looks to be in better shape. Sorry about that! sca-diff bug fixed on main, so you can ignore that too!

PR #927, opened 2026-08-28
Rework container-based workflow

terryf82:
- update Dockerfile's `PATH` and `CMD` to ensure the user is dropped into a proper shell (not a python REPL) and can run `raptor`
- update README to show how container can be run natively & via devcontainers, with and without a target-folder for local analysis

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -153,7 +153,7 @@ RUN npm install -g yarn@1.22.22 pnpm@11.8.0
 # ENVIRONMENT CONFIGURATION
 # ============================================================================
 
-ENV PATH="/opt/codeql:${PATH}"
+ENV PATH="/opt/codeql:/workspaces/raptor/bin:${PATH}"
 ENV GIT_TERMINAL_PROMPT=0
 ENV PYTHONUNBUFFERED=1
 # Ensure python can import repo packages when container runs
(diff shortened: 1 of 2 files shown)


grokjc:
Thanks - two things to fix before merge please:

### Broken shell comment in the docker run example ###

  The inline # optional target-folder mount inside the line-continued docker run command isn't valid. After a \ continuation, bash doesn't treat # as a comment because it becomes a literal argument, and docker will choke on it. Suggest either removing the comment and showing two separate examples (with and without target mount), or moving the comment above the command block as prose. Two separate examples is probably clearest:

```
  Without a target folder:
  docker run --privileged -it -v "$(pwd):/workspaces/raptor" raptor:latest

  With a local target folder:
  docker run --privileged -it -v "$(pwd):/workspaces/raptor" -v "/path/to/target:/workspaces/target" raptor:latest
```

### Wrong syntax highlighting on the devcontainer JSON block ###

  The mounts snippet is JSONC but the fenced block is tagged as bash. Should be json or jsonc.

### Minor suggestions (non-blocking): ###

  - The --privileged flag is still in the primary copy-pasteable command. Since most users won't need rr, consider showing the non-privileged form as the default and noting --privileged as an addition for rr users. Fewer people running privileged containers unnecessarily is always better.
  - The devcontainer instructions end with cd /path/to/raptor then 'code .' - it might be worth noting this opens VS Code and triggers the "Reopen in Container" prompt, since that's the step that actually launches the devcontainer. Someone unfamiliar with devcontainers might not know what to expect after 'code .'
  - The removed blank line before WORKDIR is cosmetic but makes the section headers less visually distinct in the Dockerfile. Trivial either way.

Overall: great work and we appreciate it!

grokjc:
OK, I have time 😄 I'm going to merge as-is and then follow up. Thanks again!

Issue #762, opened 2026-06-02
RFC: Migrate project dependency management from pip/requirements.txt to Poetry


rphipps75:
## Problem Statement

Current development workflow relies on a traditional `requirements.txt` pattern via `pip`. This setup is introducing manual friction and stability risks to the pipeline:

- Environment Drift
  - Teammates frequently run into "works on my machine" bugs due to unpinned transitive (sub-level) dependencies.
- Manual Setup
  - Developers must manually create, manage, and activate virtual environments (`venv`), leading to accidental global packages or missed activations.
- Cluttered Roots
  - managing split configuration setups (like `requirements.txt` vs `requirements-dev.txt`) alongside disjointed linter configurations.
- Production Bloat
  - lack a native, elegant way to strip out development-only utilities (like `pytest` or black) during production image builds.

## Proposed Solution

Migrate the package, environment, and build architecture entirely to Python Poetry.
By adopting Poetry, it can consolidate the ecosystem under modern PEP [518](https://peps.python.org/pep-0518/)/[621](https://peps.python.org/pep-0621/) standards using a single `pyproject.toml` file, replacing the fragmented tooling.

### Phase 1: Initialize Poetry & Ingest Requirements

Automate the structural transition by initialising Poetry and feeding the existing dependency list directly into the new schema:

```bash
# Initialize a new poetry framework in the current repo
poetry init --no-interaction
# Automatically parse and migrate the existing production dependencies
cat requirements.txt | xargs -I {} poetry add {}
# Isolate the testing and formatting tools into a designated development layer
cat requirements-dev.txt | xargs -I {} poetry add {} --group dev
```

## Phase 2: Lock and Verify Dependencies

Generate a strict cryptographic map of the entire environment tree to eliminate build variance:

```bash
# Resolve the deep dependency tree and generate poetry.lock
poetry lock
# Verify the environment installs smoothly inside a fresh, isolated state
poetry install
```

```toml
[tool.poetry.dependencies]
python = "3.11.8"
fastapi = "0.110.0"
uvicorn = "0.28.0"

[tool.poetry.group.dev.dependencies]
pytest = "8.1.1"
black = "24.3.0"
flake8 = "7.0.0"
```

## Phase 3: Linter & Formatter Configurations Consolidation

example...

```toml
# Linter & Formatter Configurations Consolidated Here:

[tool.black]
line-length = 88
target-version = ['py311']
include = '\.pyi?$'

[tool.isort]
profile = "black"
line_length = 88
[...]

PR #939, opened 2026-09-03
build: migrate Python dependencies to uv

acepace:
2026, it is time to get with the times and use uv.


grokjc:
There's a lot in here. I think we should get CI migrated/working first and then consider the user-facing components rather than landing this all in one hit. The PEP 735 parser addition is useful too but also stands alone as a great addition to our SCA functionality.

We also have prior work on the feat/poetry-migration branch (see #762) so we need to make a decision on that before anything happens - @danielcuthbert ?

To keep momentum, could you please carve out the PEP735 stuff @acepace (and add a test for include-group duplication too perhaps? i.e. pytest should only appear once) so we can land that whilst we sort out the rest?

(And beyond that, perhaps 'uv tool install semgrep==<ver>' works better, rather than treating it as a dependency - we'd avoid a lot of headaches with transient deps too that way)

grokjc:
See branch _feat/uv-migration_ for foundational work.

grokjc:
See f521c040d6d175eb56f107876a865b35ed2fdc7d - the branch now contains all the non-user-facing changes for review. If you can check @acepace and @danielcuthbert we could PR this branch into main and dogfood this in our own CI environment without any breaking changes for our users.

That leaves the user migration and the SCA stuff to take care of (the SCA work can land on main whenever)

grokjc:
Update: PR #959 brings the CI side of this PR to main. The PEP 735 parser was landed in https://github.com/gadievron/raptor/commit/363ff62657721dc2b72b872eec407e7dd0ab7e25.

Once that is working in CI, we can follow up with the user facing components, and we're done.

Full Co-Author credits to @acepace where applicable - thanks 😄

PR #949, opened 2026-09-11
fix(sandbox): retain namespace isolation when bind mounts fail

cinerieus:
## Summary

Keep RAPTOR's namespace sandbox active when a Linux host supports namespaces but rejects the bind-mount tree.

WSL is one affected environment. Bind-mounting `/usr` can return `EINVAL` even though PID, network, Landlock, and seccomp isolation are available.

The reduced backend is a lesser filesystem-isolation tier. Untrusted runs must opt in through the existing `RAPTOR_ALLOW_DEGRADED_UNTRUSTED=1` control. No new consent mechanism is added.

## Behavior

When bind-tree setup or in-tree execution fails, RAPTOR retries through the existing spawn backend with mount setup disabled. The retry retains:

- user, PID, IPC, and network namespaces;
- fresh procfs;
- Landlock read and write enforcement;
- seccomp and resource limits;
- audit tracing;
- proxy-controlled network access.

The caller's read policy is preserved. RAPTOR does not force read restriction onto calls that use the historical read-everywhere default, so venv, pipx, and user-installed tools keep working without per-consumer grants.

The result records `backend=landlock-pidns`, `mount_ns_active=false`, and the reduced posture in `sandbox-summary.json`. Repeated demotions warn once per process; later cache hits and per-call scratch diagnostics log at debug level.

## Untrusted-run gates

An untrusted run refuses the reduced backend unless `RAPTOR_ALLOW_DEGRADED_UNTRUSTED=1` was set by the operator. The consent value is retained across RAPTOR-owned worker environments, then stripped at target execution seams by both the Linux PID 1 shim and the macOS seatbelt shim.

Mountless untrusted execution also refuses Landlock ABI below 3. Older ABIs cannot block `truncate()` outside the write allowlist.

Modes that require bind mounts, including rootfs and fingerprint-persona execution, continue to fail closed. The retry also fails closed if its remaining isolation layers cannot engage.

## Temporary storage

Read-restricted host-visible execution must not inherit writable host-shared `/tmp` or `/dev/shm` grants. Every such mountless path now uses the same private-scratch policy:

- direct mountless selection;
- explicit per-call mount skipping;
- the first bind-tree retry;
- speculative-cache retries;
- the legacy subprocess demotion.

RAPTOR creates a private mode-`0700` scratch directory, steers `TMPDIR`, `TEMP`, and `TMP` to it, and removes host-shared `/tmp` and `/dev/shm` from that call's write grants. Explicit non-shared writable paths remain available. Unrestricted calls retain the documented host temporary-directory baseline.

## Residual exposure

The bind-tree backend remains the stronger default. In mountless mode host paths remain visible by name. Landlock limits requested reads and writes, but it cannot mediate metadata-only operations such as `chmod`. `ENOENT` versus `EACCES` also exposes a filesystem-existence oracle. These limits are documented in `docs/sandbox.md` and the threat model.

## SCA diagnostics

Agentic SCA failures now redact captured stderr once, persist a bounded mode-`0600` diagnostic, and log the final line from the same redacted value. SCA execution and model-triage behavior are unchanged. The earlier global `--no-llm` change is not part of this PR.

## Compatibility

- Hosts with working bind mounts continue to use the preferred bind-tree backend.
- Caller read policy and unrestricted `/tmp` behavior remain unchanged.
- No calibration, SCA, or CodeQL compensating path grants are added.
- `GIT_ASKPASS` uses the absolute inert path `/usr/bin/true`.
- Context-built child environments are already wrapper-scrubbed before the lower backend receives them.
- Upstream's fresh-procfs demotion fixes are inherited from current `main`; this PR does not duplicate them.
- The seatbelt shim's environment-strip tuple is recorded in the vocabulary baseline as policy data, not learned name vocabulary.

## Commits

- `f58bd1049` retains namespaces when bind mounts fail, adds the security gates, private-scratch policy, reporting, documentation, and tests.
- `dd6ffaee9` adds bounded redacted SCA failure diagnostics.
- `4737ffcf0` propagates degraded-run consent through RAPTOR workers, strips it at Linux and macOS target seams, and records the seatbelt policy tuple in the vocabulary baseline.

## Validation

Local validation on the affected WSL host:

- reviewer-focused sandbox and vocabulary suite: 48 passed;
- wider sandbox/config/SCA set from the previous revision: 136 passed, 4 skipped;
- vocabulary checker, Ruff, `py_compile`, and `git diff --check`: passed;
- live reduced-backend probe: host `/tmp` write denied, private-scratch tempfile creation succeeded.

One low-level audit test fails on this host because it invokes the mount backend directly and `/usr` is not bind-mountable. It bypasses the context retry changed by this PR. Linux and macOS PR jobs provide the mount-capable and seatbelt coverage.

### DIFF
--- a/.github/scripts/vocab_baseline.json
+++ b/.github/scripts/vocab_baseline.json
@@ -1641,6 +1641,11 @@
     "count": 10,
     "kind": "literal"
   },
+  "libexec/raptor-seatbelt-shim::_strip": {
+    "count": 10,
+    "kind": "literal",
+    "note": "target environment-strip tuple mirrored from core.config.TARGET_ENV_STRIP_SET for the isolated macOS shim, not a learned name vocabulary"
+  },
   "libexec/raptor-self-test::_LLM_ENV_VARS": {
     "count": 10,
     "kind": "literal",
--- a/core/config/__init__.py
+++ b/core/config/__init__.py
@@ -862,7 +862,9 @@ class RaptorConfig:
     # on the platform (e.g. some Windows builds).
     GIT_ENV_VARS: ClassVar[dict] = {
         "GIT_TERMINAL_PROMPT": "0",
-        "GIT_ASKPASS": "true",
+        # Absolute so Git cannot resolve an operator-controlled executable
+        # through PATH inside a sandbox.
+        "GIT_ASKPASS": "/usr/bin/true",
         "GIT_CONFIG_GLOBAL": "/dev/null",
         "GIT_CONFIG_SYSTEM": "/dev/null",
         "GIT_CONFIG_NOSYSTEM": "1",
--- a/core/config/__init__.py
+++ b/core/config/__init__.py
@@ -561,6 +561,11 @@ class RaptorConfig:
         # TARGET_ENV_STRIP_SET (sandboxed code never sees the
         # credential; see core/sandbox/context.py).
         "RAPTOR_SESSION_PID", "RAPTOR_SESSION_TOKEN",
+        # Host-level consent for RAPTOR's reduced untrusted sandbox tier.
+        # RAPTOR-owned workers must inherit it so the decision made at the
+        # launcher boundary is not silently lost.  Target-bound envs strip it
+        # below, preventing scanned code from changing sandbox policy.
+        "RAPTOR_ALLOW_DEGRADED_UNTRUSTED",
     })
 
     # Variables that must NEVER reach code executed on behalf of a
@@ -577,6 +582,7 @@ class RaptorConfig:
     TARGET_ENV_STRIP_SET = frozenset({
         "CLAUDECODE", "_RAPTOR_TRUSTED",
         "RAPTOR_SESSION_PID", "RAPTOR_SESSION_TOKEN",
+        "RAPTOR_ALLOW_DEGRADED_UNTRUSTED",
         # Framework-identity values: each one names RAPTOR (or its
         # checkout / output layout) to any target that runs `env`,
         # defeating the anti-fingerprint posture in one getenv. No
--- a/core/config/tests/test_config.py
+++ b/core/config/tests/test_config.py
@@ -92,6 +92,13 @@ class TestGetSafeEnv:
         env = RaptorConfig.get_safe_env()
         assert env.get("PYTHONUNBUFFERED") == "1"
 
+    def test_degraded_untrusted_consent_reaches_raptor_workers(self):
+        with patch.dict(os.environ, {
+            "RAPTOR_ALLOW_DEGRADED_UNTRUSTED": "1",
+        }):
+            env = RaptorConfig.get_safe_env()
+        assert env.get("RAPTOR_ALLOW_DEGRADED_UNTRUSTED") == "1"
+
     def test_ef_benign_knobs_survive_scrub(self):
         """Numeric/boolean feasibility knobs must reach scrub-spawned
         children (the loader runs inside them); the exec-path class
--- a/core/config/tests/test_config.py
+++ b/core/config/tests/test_config.py
@@ -156,7 +156,7 @@ class TestGetGitEnv:
 
     def test_sets_askpass(self):
         env = RaptorConfig.get_git_env()
-        assert env.get("GIT_ASKPASS") == "true"
+        assert env.get("GIT_ASKPASS") == "/usr/bin/true"
 
     def test_also_strips_dangerous_vars(self):
         injected = {var: "bad" for var in RaptorConfig.DANGEROUS_ENV_VARS}
--- a/core/git/clone.py
+++ b/core/git/clone.py
@@ -20,7 +20,7 @@ Both wrap their ``git`` subprocess in ``core.sandbox.run_untrusted_networked``
     the target / repo directory;
   - sanitised env (``RaptorConfig.get_git_env()`` — clears
     HTTP_PROXY / NO_PROXY etc., sets GIT_TERMINAL_PROMPT=0 and
-    GIT_ASKPASS=true so a malformed-credential prompt can't hang
+    GIT_ASKPASS=/usr/bin/true so a malformed-credential prompt can't hang
     the run);
   - bounded timeout (``RaptorConfig.GIT_CLONE_TIMEOUT``).
 
--- a/core/sandbox/_macos_spawn.py
+++ b/core/sandbox/_macos_spawn.py
@@ -587,7 +587,11 @@ def run_sandboxed(cmd: list[str], *,
         child_env = dict(env)
         if strict_env:
             from core.config import RaptorConfig
-            child_env = {k: v for k, v in child_env.items() if k not in RaptorConfig.DANGEROUS_ENV_VARS}
+            child_env = {
+                k: v for k, v in child_env.items()
+                if (k not in RaptorConfig.DANGEROUS_ENV_VARS
+                    or RaptorConfig.GIT_ENV_VARS.get(k) == v)
+            }
     else:
         from core.config import RaptorConfig
         child_env = RaptorConfig.get_safe_env()
--- a/core/sandbox/_spawn.py
+++ b/core/sandbox/_spawn.py
@@ -2533,9 +2533,11 @@ def run_sandboxed(
                     if strict_env:
                         from core.config import RaptorConfig
                         _dangerous = set(RaptorConfig.DANGEROUS_ENV_VARS)
+                        _safe_git = RaptorConfig.GIT_ENV_VARS
                         exec_env = {
                             k: v for k, v in exec_env.items()
-                            if k not in _dangerous
+                            if (k not in _dangerous
+                                or _safe_git.get(k) == v)
                         }
                 else:
                     # env=None → scrubbed allowlist env, NOT the full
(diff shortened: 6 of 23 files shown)


grokjc:
This is great work, but there are a few gaps/blockers to address.

**Test Failures**

You've turned the old "cmd[0] resolves outside the bind tree" case (Landlock-only demotion, historical read default) into the new mountless backend with forced read restriction. That path fires on any host where the tool is a venv Python, pipx, or something in `~/.local/bin` - a venv Python now dies on `pyvenv.cfg` because `$HOME` isn't in the allowlist, and the scratch swap breaks the documented `/tmp` write baseline (`test_env_layout_scrub`, `test_default_rlimits` nofile, `test_tmp_baseline_e2e`). 

The PR's own CI run shows 13 failures in the same three files: the mountless retry dies at exec (`X: exec: permission denied` → `SandboxSetupError`), because exec needs the file to be readable and the tool binary isn't in the forced allowlist. So on a runner-shaped host the fallback doesn't degrade, it fails outright - the opposite of the PR's goal. And the latest commits point at the underlying problem: calibrate, SCA, and now the CodeQL wrapper have each needed a hand-rolled compensating grant. When every consumer you notice needs a patch, the forced restriction is scoped wrong. 

**The untrusted downgrade can't be automatic** 

Today a mount failure on an untrusted run is a hard error unless the operator sets `RAPTOR_ALLOW_DEGRADED_UNTRUSTED=1`. Your retry trades mount isolation away behind a single `logger.warning`, and then the speculative cache demotes later calls of that binary at DEBUG. Nothing outside core/sandbox reads `backend`/`mount_ns_degraded`, so reports never see it. Same posture as `require_proxy_netns`: if the tier isn't available, fail closed unless the operator opted in. We don't want a new env var for this: either treat the reduced backend as a lesser tier under the existing `ALLOW_DEGRADED_UNTRUSTED` opt-in, or make it a project setting. 

**Mountless changes what Landlock can't do into real exposure** 

With the bind tree, host files simply aren't there. Without it: Landlock has no metadata controls, so untrusted code can `chmod 0666 ~/.ssh/id_rsa` (you warn about this, but don't stop it); on ABI < 3 (exactly the WSL2 5.15/6.1 kernels you're targeting) `truncate()` outside the allowlist succeeds; and stat/EACCES-vs-ENOENT is a host filesystem oracle. Untrusted runs on ABI < 3 should refuse, and the new THREAT_MODEL paragraph needs to say "reads are limited, metadata isn't" in order to be accurate.

**Smaller stuff**

- `raptor_agentic.py` logs the raw SCA stderr line on the failure path; `redact_secrets` only covers the file sink. We need to redact both.
- The `--no-llm` flip for agentic SCA changes behavior on every host, not just demoted ones, and drops SCA's specialised triage on an untested "parent covers it" assumption. Separate issue, separate PR please.
- Applying `strict_env=False` to the backends fixes a real bug you'd introduced (`TMPDIR` is in `DANGEROUS_ENV_VARS`, so the re-strip was defeating your own scratch override - the fix is right). But it also turns the two-layer env scrub into one layer. Can you add a test or at least a comment re the fact that context-built child envs always derive from the wrapper-scrubbed env?
- `GIT_ASKPASS="true"` is PATH-relative and git will exec it in the sandbox. Keep the preserved-value set to absolute inert paths.
- Others: the `locals().get("_spawn_without_mount", …)` stuff; `_run_spawn_backend` return annotation; literal `\n\n` in one commit message.

Thanks for taking a look at this. Sandbox changes are hard to get right first time. Ask me how I know 😄

cinerieus:
Thanks for the detailed review. I have reworked the branch to address each point.

The mount failure path now retries through the existing namespace backend with only the bind tree disabled. It retains user, PID, IPC, and network namespaces, fresh `/proc`, Landlock, seccomp, proxy enforcement, resource limits, and audit tracing.

The caller's read policy and documented `/tmp` baseline are preserved. This avoids forcing read restriction onto venv, pipx, and user-installed tools. The calibration, SCA, and CodeQL compensating grants are no longer present.

For untrusted execution, the reduced filesystem tier requires the existing `RAPTOR_ALLOW_DEGRADED_UNTRUSTED=1` opt-in. Untrusted-workload state is tracked separately from that opt-in, so every mountless untrusted run refuses Landlock ABI below 3, including calls with `restrict_reads=False`.

The reduced tier is reported as `backend=landlock-pidns`, with `mount_ns_active=false` and `posture.mountless_backend=true` in the persisted sandbox summary.

The threat model now states the residual exposure directly: host paths remain visible, Landlock cannot mediate metadata-only operations such as `chmod`, and `ENOENT` versus `EACCES` provides a filesystem-existence oracle. The bind-tree backend remains the stronger default.

The smaller points are also addressed:

- SCA stderr is redacted once before both terminal logging and bounded mode-`0600` persistence.
- The agentic SCA `--no-llm` change has been removed from this PR and kept on a separate branch.
- The lower backend receives the already-scrubbed context environment without stripping wrapper-owned values again. A regression test documents this boundary.
- `GIT_ASKPASS` now uses the absolute inert path `/usr/bin/true`.
- The `locals().get(...)` workaround is gone.
- `_run_spawn_backend` has an explicit return annotation.
- The malformed commit-message escaping and stale "Landlock-only" diagnostics are fixed.

The branch is now split into two focused commits:

- `8868ef0bc`: namespace-preserving fallback, security policy, reporting, documentation, and tests
- `812832d1d`: SCA failure-diagnostic redaction

Validation:

- [Full Test Suite](https://github.com/cinerieus/raptor/actions/runs/34616597790): passed, including Linux and macOS sandbox jobs
- [Lint](https://github.com/cinerieus/raptor/actions/runs/34616600534): passed
- Reviewer regression set: 131 passed
- Focused fresh-procfs and mountless contract: 6 passed

The fork-only workflow adjustment used for validation is not part of this PR.

grokjc:
The follow-up consent-propagation commit is a good one: but it misses the pid1-shim's macOS twin - the seatbelt shim keeps its own hand-copied strip set, and its tripwire test now fails: `test_seatbelt_shim_env_strip.py::test_default_strip_covers_target_env_strip_set` highlights that RAPTOR_ALLOW_DEGRADED_UNTRUSTED leaks into the sandbox.

One new problem though, found in a second adversarial pass, and it's a blocker:

**The mountless fallback grants read-restricted runs write access to host-shared `/tmp` and `/dev/shm`.** On a mount-capable host, `writable_paths` is computed at construction time as literal `/tmp` + `/dev/shm` (the private-scratch swap only fires when `use_mount` is false at construction), and the retry passes that list to the mountless spawn unchanged. Base's demotion lane deliberately recomputes a private 0700 scratch for exactly this case (`_demoted_call_writable`) and the rationale comment explains why: cross-process plant/poison and scanned-repos-under-/tmp. The `# private mountless TMPDIR` comment near the retry is wrong on these hosts as `_private_scratch_dir` is None there. Fix is to apply the same private-scratch recompute (and TMPDIR steering) to the mountless retry lane, plus a test pinning it. The new docs should also mention the write side, not just read visibility.

Three smaller things:

- **The PR description still describes the previous design** and the forced read restriction, scratch TMPDIR swap, the SCA `--no-llm` change, the compensating grants don't exist in this tree, and the description omits the two most important properties you added (untrusted refusal without the opt-in; ABI < 3 refusal). 
- Commit message accuracy: the sandbox commit frames the `/tmp` baseline as "preserving historical behavior", but for *restricted* runs the historical demoted behavior was the private scratch - that's the regression above. And the redaction commit's "previously persisted a redacted copy but could log the raw line" describes an earlier revision of your own branch, not the merge base (where SCA stderr was captured and discarded) - the fix is right, the history isn't.
- The per-call mountless warning will spam on cached demotions (a Semgrep-heavy workload hits it hundreds of times per run) suggest we log the cache-hit arm once, like base did.

While testing I also confirmed a **pre-existing** gap on both trees (not yours to fix here, flagging for transparency since I found it reviewing your change): the mid-setup spawn-*exception* ladder can land an untrusted run on the legacy host-procfs subprocess lane without the opt-in. Not as bad as it sounds as needs environmental trigger. Working on that now, thanks for giving me a kick to look at this code again!

cinerieus:
Second-review fixes are now pushed.

The branch is rebased onto current `main` at `2aa50e1d0`, including the upstream spawn-exception fail-closed fix and sandbox-test consent stripping.

This revision:

- strips degraded-run consent in both Linux PID1 and macOS seatbelt target shims;
- uses private mode-0700 scratch storage for every read-restricted mountless path, including direct, retry, cached, explicit-skip, and legacy demotion paths;
- removes host-shared `/tmp` and `/dev/shm` write grants from those calls;
- deduplicates the operator warning and keeps per-call diagnostics at debug level;
- updates the threat model, sandbox documentation, and commit descriptions;
- replaces the stale PR description.

grokjc:
Thanks!

One thing left, and it's mechanical: `check_vocab_lists` fails on this head because the seatbelt shim's `_strip` tuple grew to 10 names, which trips the checker's literal-list threshold. Add a `libexec/raptor-seatbelt-shim::_strip` entry to `.github/scripts/vocab_baseline.json` with a short review note (it's an env-strip tuple, not a name vocabulary - exactly what the baseline is for). I'm updating CI to include this test on PRs moving forwards to avoid this trap elsewhere.

Also a heads-up: main just gained another change (https://github.com/gadievron/raptor/commit/ddecf1b273fbe0da80b7969639f9181fa29126bd) in the same neighbourhood so you'll want one more rebase and then we're done!

grokjc:
LGTM - thanks for all your work on this.

Issue #810, opened 2026-06-16
feat: Persistent RAPTOR Graph Store


danielcuthbert:
# Persistent RAPTOR Graph Store

Right now RAPTOR, the little dinosaur that could, is super good at taking /understand and turning code into attacker-relevant maps. Then she maps, traces, hunts, then feeds /validate and /threat-model. 

What I wanted was the ability to ask “what can the attacker control?”, “where are the trust boundaries?”, “what sinks matter?”, “which flows are unchecked?”, “what path should /validate prove or kill?”

I wanted a map that could be proved/disproved and one thing led to another where maybe it's now time to pray at the altar of SQLite and make this even more usable. 

## So Wut Daniel?
Imagine, if you may, a project-scoped SQLite graph store behind `/understand`, while keeping the existing JSON artefacts as the public contract. `/understand` becomes the main graph producer; `/validate`, `/threat-model`, `/agentic`, `/diagram`, SCA, fuzzing, web, and CodeQL consume it through one internal API.

Default storage location:

- Active project: `out/projects/<project>/graph/raptor.graph.sqlite`
- No active project: `<run-output-dir>/graph/raptor.graph.sqlite`
- SQLite uses WAL mode, schema migrations via `PRAGMA user_version`, and degrades to the existing JSON flow if the DB is absent.

## Waht changes innit?

We now have a `core/understand_graph/` with:

- `store.py`: open DB, apply migrations, transactions, WAL setup.
- `schema.py`: stable node/edge IDs and graph types.
- `ingest.py`: ingest `checklist.json`, `context-map.json`, `flow-trace-*.json`, `variants.json`, verified outcomes, SCA reachability, web/fuzz evidence.
- `queries.py`: reachable sinks, entry-to-sink paths, stale data, high-risk boundaries, verified outcomes near a finding.
- `adapters.py`: small adapters for `/validate`, `/threat-model`, `/diagram`, `/agentic`.

Core tables:

- `snapshots`: target path, target hash, git SHA, run dir, created time.
- `nodes`: functions, files, entry points, trust boundaries, sinks, findings, threats, dependencies, verified outcomes.
- `edges`: calls, contains, reaches, taints, crosses boundary, confirmed by, blocked by, imports dependency.
- `artifacts`: provenance for JSON files and generated diagrams.
- `metadata`: schema/version info.

All flexible fields live in `props_json` / `evidence_json` so the schema does not become brittle.

## Integration Flow
[...]

PR #950, opened 2026-09-12
Feat/understand graph store

grokjc:
As per #810, LGTM!


danielcuthbert:
Oh you've made me a happy man!! Amazing, thanks for fixing glaring issues. This is a solid addition

PR #951, opened 2026-09-12
Feat/openant integration

grokjc:
I think this is pretty much ready to merge now.


danielcuthbert:
100% agree here, time to get this into main and people loving it

